PR #3, opened 2015-06-12
Insertion-ordered BSON documents

kyeah:
Hiya! This is a drop-in replacement for the BTreeMap BSON document that maintains an ordering of the keys based on insertion. This is necessary for certain types of operations, such as [sorting](http://docs.mongodb.org/manual/reference/method/cursor.sort/), where the sort precedence relies on the key ordering.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -21,12 +21,11 @@
 
 //! BSON definition
 
-use std::collections::BTreeMap;
-
 use chrono::{DateTime, UTC};
 use rustc_serialize::json;
 use rustc_serialize::hex::ToHex;
 
+use ordered::OrderedDocument;
 use spec::{ElementType, BinarySubtype};
 
 /// Possible BSON value types.
@@ -51,8 +50,8 @@ pub enum Bson {
 
 /// Alias for `Vec<Bson>`.
 pub type Array = Vec<Bson>;
-/// Alias for `BTreeMap<String, Bson>`.
-pub type Document = BTreeMap<String, Bson>;
+/// Alias for `OrderedDocument`.
+pub type Document = OrderedDocument;
 
 impl Bson {
     /// Get the `ElementType` of this value.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,4 +54,4 @@ pub mod spec;
 mod bson;
 mod encoder;
 mod decoder;
-
+mod ordered;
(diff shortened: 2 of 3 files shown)


zonyitoo:
For `OrderedDocument`

1. `OrderedDocument::keys` should return an iterator instead of a `Vec`.

2. `OrderedDocument::values` should return an iterator instead of a `Vec`.

zonyitoo:
Also for `OrderedDocumentIntoIterator` and `OrderedDocumentIterator`,

1. `OrderedDocumentIterator` may be renamed to `OrderedDocumentIntoIterator`. I think that is what you want, right?

2. Those iterators actually could make use of `Vec`'s iterators.

3. For `IntoIterator`, when getting an element from the underling `BTreeMap`, IMHO, could use `BTreeMap::remove` to get an element from the map.

Also, IMHO, to maintain insertion order, I think we could customize a `BTreeMap`, which will also store tree nodes in a LinkedList in insertion order.

zonyitoo:
For the last suggestion, a data structure like [LinkedHashMap](http://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashMap.html) could be used.

And I think someone has already implemented that: https://crates.io/crates/linked-hash-map

kyeah:
1. `keys()` and `values()` now returns iterators. Since `Vec`'s iterators are non-clonable, these are also non-clonable.

2. `OrderedDocumentIterator` and `OrderedDocumentIntoIterator` names swapped. This now accurately reflects the borrowed and owned states of the contained document.

3. The iterators now use `Vec`'s iterators internally.

4. `OrderedDocumentIntoIterator` now uses `BTreeMap::remove`.

I agree that a custom/generalized BTreeMap implementation would be a better way to do it. If LinkedHashMap didn't use unstable features, I would suggest using it directly. We are currently building on top of this library for a new Mongo driver prototype, so I needed a quick fix, but I would definitely prefer building a more general data structure if I had time. If anyone wants to jump on it, be my guest :)

zonyitoo:
@kyeah I still recommend to use [linked-hash-map](https://crates.io/crates/linked-hash-map).

Are you using stable version? Why not use nightly?

kyeah:
hey @zonyitoo,

We're trying to build a stable Rust driver for the Mongo community, so we'd rather avoid using any unstable features within nightly.

PR #4, opened 2015-06-18
Fix boolean encoding

kyeah:
The boolean encodings are mixed. http://bsonspec.org/spec.html


zonyitoo:
Oh crap! You are right!

PR #5, opened 2015-07-03
ObjectId container and generator

kyeah:
This replaces the generic [u8; 12] representation of OIDs with an ObjectId struct, providing some more type safety around OIDs that are passed around and allowing valid OIDs to be generated client-side. This is a breaking change -- existing code must use `ObjectId::with_bytes()` and `oid.bytes()` to pass and get raw [u8; 12] oid representations.

Opening a PR since this is a moderately-sized update, @zonyitoo :)

[Relevant PR in mongo-rust-driver-prototype.](https://github.com/mongodbinc-interns/mongo-rust-driver-prototype/pull/75)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,10 @@ homepage = "https://github.com/zonyitoo/bson-rs"
 name = "bson"
 
 [dependencies]
-chrono = "*"
-byteorder = "*"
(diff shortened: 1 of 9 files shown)


zonyitoo:
Looks good! I think this could be merged.

PR #6, opened 2015-07-03
BSON Macro + Moved tests to separate module

kyeah:
Also added some From traits for u32, u64, and f32. 

Tests should be run single-threaded since the OID unit tests use a static OID_COUNTER variable.
`export RUST_TEST_THREADS=1`


zonyitoo:
This macro is very useful!

kyeah:
Yep! It still needs a little work to support arrays of documents and remove the parentheses at the root level (`"key" => "value"` instead of `"key" => ("value")`), but it's super awesome. Credit to @saghm :)

PR #7, opened 2015-07-09
add to_object method for Bson

elexing:
Before , user cannot directly convert bson to bussiness object ,but this is easy to do using the method `to_object` .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bson"
-version = "0.1.2"
+version = "0.1.3"
 authors = ["Y. T. Chung <zonyitoo@gmail.com>"]
 description = "Encoding and decoding support for BSON in Rust"
 license = "MIT"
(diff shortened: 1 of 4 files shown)


zonyitoo:
I am not sure this is the best way to do this. Maybe we could have an `Encoder` and `Decoder` that impl `rustc_serialize::Encoder` and `rustc_serialize::Decoder` instead of converting it to `Json` and then decode it to the object.

cc @kyeah

kyeah:
Yes, I think doing two conversions is a bit roundabout and takes away from one of BSON's advantages, which is its encoding and decoding efficiency. A Bson `Encoder` and `Decoder` would be ideal to convert it directly, although this is a bit more involved.

elexing:
@zonyitoo Yes, it's better.

thijsc:
I think #16 solves this same issue.

thijsc:
I think we can close this one since #16 has been merged.

kyeah:
I believe this and #16 solve different issues: #16 provides easy access to inner bson values, while this is converting arbitrary user-defined structs into Bson documents and vice versa.

thijsc:
Ah right. In that something on top of https://github.com/serde-rs/serde seems like a good idea.

PR #9, opened 2015-07-23
Format display

saghm:
For the logging functionality in the MongoDB driver, we need to be able to print out documents in the format that they would appear in the shell. The ability to print the documents in a readable format would be useful for almost anyone using BSON, though.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -21,6 +21,9 @@
 
 //! BSON definition
 
+use std::fmt::{Display, Error, Formatter};
+use std::str;
+
 use chrono::{DateTime, UTC};
 use rustc_serialize::json;
 use rustc_serialize::hex::ToHex;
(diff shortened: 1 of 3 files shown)


zonyitoo:
One comment: the `Binary` type may be printed with `rustc_serialize::hex`.

saghm:
It doesn't seem to get the correct result when I use `ToHex` to convert the binary data. I think it might just convert each hex number to its string equivalent (e.g. `74` converts to `"74"` rather than to `"t"`).

zonyitoo:
Indeed. Then I think this line

```rust
format!("BinData({}, \"{}\")", u8::from(t), string)
```

could be modified to

```rust
format!("BinData({}, \"{:?}\")", u8::from(t), string)
```

saghm:
I just took another look at how the MongoDB shell prints out the binary values, and it's actually much stranger than I thought (for instance, both "hello---" and "hello___" get printed out as "helloAAA", and I really have no clue why). Trying to emulate this properly would be much more work than I originally thought would be, so I think it's probably better to just go with your `to_hex()` suggestion and just print out the raw number.

zonyitoo:
Thanks for your contribution!

PR #10, opened 2015-07-26
Use option for mutable static

kyeah:
Replaces the [0; 3] check for MACHINE_BYTES with an Option check.

Also implements Clone for Errors by manually cloning io::Error. Not sure why io::Error doesn't implement Clone, but it inhibits the functionality and usability of structs that have owned Error fields like Cursor in the [mongo driver](https://github.com/kyeah/mongo-rust-driver-prototype/commit/d534c4156ce1092e4f87167c76fb6fd5f13a8049#diff-90872b5d72d7eb7ca5dac072e26fb38eR228).

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -24,7 +24,7 @@ const COUNTER_OFFSET: usize = PROCESS_ID_OFFSET + PROCESS_ID_SIZE;
 const MAX_U24: usize = 0xFFFFFF;
 
 static OID_COUNTER: AtomicUsize = ATOMIC_USIZE_INIT;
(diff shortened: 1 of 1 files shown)


zonyitoo:
Why `Error` has to be Clone?

kyeah:
On second thought, you're right. Structs with err fields can be implemented / used in ways to work around the cloning restriction; reverting the second commit.

zonyitoo:
One more "Rusty" suggestion:

```rust
if let Some(bytes) = MACHINE_BYTES.as_ref() {
    return bytes.clone();
}
```

kyeah:
Good catch.

PR #11, opened 2015-07-29
Fix UtcDatetime en/decoding

matsimitsu:
Currently UtcDatetime is encoded as a `i64` from the chronos timestamp, however the chronos timestamp is returned in seconds since epoch(https://lifthrasiir.github.io/rust-chrono/chrono/datetime/struct.DateTime.html#method.timestamp).

According to the BSON spec a UtcDateTime should be encoded as miliseconds since epoch: http://bsonspec.org/spec.html

> UTC datetime - The int64 is UTC milliseconds since the Unix epoch.

This pull request attempts to solve this issue. 

(I've only been using rust for a few days now, so I'd love to know if this could be done better :))

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -25,7 +25,8 @@ use std::io::{self, Read};
 use std::{str, error, fmt};
 
 use byteorder::{self, LittleEndian, ReadBytesExt};
-use chrono::{DateTime, NaiveDateTime, UTC};
+use chrono::{UTC};
+use chrono::offset::TimeZone;
 
 use spec::{self, BinarySubtype};
 use bson::{Bson, Array, Document};
@@ -215,7 +216,7 @@ fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8) -> DecoderResult<Bson>
         Some(TimeStamp) => read_i64(reader).map(Bson::TimeStamp),
         Some(UtcDatetime) => {
             let time = try!(read_i64(reader));
-            Ok(Bson::UtcDatetime(DateTime::from_utc(NaiveDateTime::from_timestamp(time / 1000, 0), UTC)))
+            Ok(Bson::UtcDatetime(UTC.timestamp(time / 1000, (time % 1000) as u32 * 1000000)))
         },
 	Some(Deprecated) |
         Some(Undefined) |
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -215,7 +215,7 @@ fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8) -> DecoderResult<Bson>
         Some(TimeStamp) => read_i64(reader).map(Bson::TimeStamp),
         Some(UtcDatetime) => {
             let time = try!(read_i64(reader));
-            Ok(Bson::UtcDatetime(DateTime::from_utc(NaiveDateTime::from_timestamp(time, 0), UTC)))
+            Ok(Bson::UtcDatetime(DateTime::from_utc(NaiveDateTime::from_timestamp(time / 1000, 0), UTC)))
         },
 	Some(Deprecated) |
         Some(Undefined) |
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -24,6 +24,7 @@
 use std::io::{self, Write};
 use std::iter::IntoIterator;
 use std::{mem, error, fmt};
+use chrono::Timelike;
 
 use byteorder::{self, LittleEndian, WriteBytesExt};
 
@@ -162,7 +163,9 @@ fn encode_bson<W: Write + ?Sized>(writer: &mut W, key: &str, val: &Bson) -> Enco
             try!(writer.write_u8(From::from(subtype)));
             writer.write_all(data).map_err(From::from)
         },
-        &Bson::UtcDatetime(ref v) => write_i64(writer, v.timestamp() * 1000),
+        &Bson::UtcDatetime(ref v) => {
+            write_i64(writer, (v.timestamp() * 1000) + (v.nanosecond()  / 1000000)  as i64)
+        },
         &Bson::Null => Ok(())
     }
 }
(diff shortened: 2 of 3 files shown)


zonyitoo:
Oh, this should be a bug. @kyeah Have you encountered this in your MongoDB driver? Can you test this PR?

kyeah:
Ah, I think we have encountered odd dates, but we never looked into it. Thanks for bringing this up.

src/decoder.rs
```diff
@@ -215,7 +215,7 @@ fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8) -> DecoderResult<Bson>
         Some(TimeStamp) => read_i64(reader).map(Bson::TimeStamp),
         Some(UtcDatetime) => {
             let time = try!(read_i64(reader));
-            Ok(Bson::UtcDatetime(DateTime::from_utc(NaiveDateTime::from_timestamp(time, 0), UTC)))
+            Ok(Bson::UtcDatetime(DateTime::from_utc(NaiveDateTime::from_timestamp(time / 1000, 0), UTC)))
```
kyeah: We are losing accuracy here; you can specify nanoseconds with 

```rust
NaiveDateTime::from_timestamp(time / 1000, (time % 1000) * 1000000)
```

src/encoder.rs
```diff
@@ -162,7 +162,7 @@ fn encode_bson<W: Write + ?Sized>(writer: &mut W, key: &str, val: &Bson) -> Enco
             try!(writer.write_u8(From::from(subtype)));
             writer.write_all(data).map_err(From::from)
         },
-        &Bson::UtcDatetime(ref v) => write_i64(writer, v.timestamp()),
+        &Bson::UtcDatetime(ref v) => write_i64(writer, v.timestamp() * 1000),
```
kyeah: Same here; you can use the nanoseconds as well.

```rust
v.timestamp() * 1000 + v.nanoseconds() / 1000000
```

src/decoder.rs
matsimitsu: How about this? 

```
Ok(Bson::UtcDatetime(UTC.timestamp(time / 1000, (time % 1000) as u32 * 1000000)))
```

src/encoder.rs
matsimitsu: `v.nanoseconds()` doesn't seem to exist, There is a `nanosecond()` trait in the `TimeLike` trait (https://lifthrasiir.github.io/rust-chrono/chrono/trait.Timelike.html#method.nanosecond) 

If I understand the docs correctly it should become: 

```rust
        &Bson::UtcDatetime(ref v) => {
            write_i64(writer, ((v.timestamp() * 1000) + (v.nanosecond()  / 1000000)  as i64))
        },
```

src/decoder.rs
kyeah: :+1:

src/encoder.rs
kyeah: Yup, for methods implemented through traits, the trait needs to be imported. The outer parens are unnecessary, but that's a small nit :)

kyeah:
Heya, didn't realize this was updated. I decided to test the values being inserted and returned by encoding/decoding, and realized that the arithmetic isn't being performed with the right types. It seems like the type system always defaults to 32-bit types, even if the original value is i64.

```rust
let gives_overflow = v.timestamp() * 1000;
let gives_i64: i64 = v.timestamp() * 1000;
```

Here's the full test I used:

```rust
#[test]                                                                                                              
fn test_encode_decode_utc_date_time() {                                                                                     
    let src = UTC.timestamp(1286705410, 2000000);
                                                                                                     
    let mut doc = Document::new();                                                                                   
    doc.insert("key".to_owned(), Bson::UtcDatetime(src));                                                            
                                                                                                                     
    let mut buf = Vec::new();                                                                                        
    encode_document(&mut buf, &doc).unwrap();                                                                        
                                                                                                            
    let mut cursor = Cursor::new(buf);                                                                                 
                                                                                                                     
    let new_doc = decode_document(&mut cursor).unwrap();                                                               
    match new_doc.get("key") {                                                                                       
        Some(&Bson::UtcDatetime(ref datetime)) => {                                                                  
            assert_eq!(1286705410, datetime.second());  // panics on value '10'                                                               
            assert_eq!(2000000, datetime.nanosecond());                                                              
        },                                                                                                           
        _ => panic!("!"),                                                                                            
    }                                                                                                                          
} ```

**Comment by matsimitsu** [comment]
I think it's correct the `.second` call on a `UtcDatetime` in chrono returns the seconds of the timestamp. In this case the timestamp we use to create the bson is: 
`2010-10-10T10:10:10.002Z` 

Calling `.second` on this returns `10`, because that's the number of seconds in that timestamp for the hour.  See: https://lifthrasiir.github.io/rust-chrono/chrono/trait.Timelike.html#method.second

If you call `.timestamp` on that value, it returns the correct value.

The test would become:


```rust
#[test]
fn test_encode_decode_utc_date_time() {
    let src = UTC.timestamp(1286705410, 2000000);
    println!("{:?}", src);

    let mut doc = Document::new();
    doc.insert("key".to_owned(), Bson::UtcDatetime(src));

    let mut buf = Vec::new();
    encode_document(&mut buf, &doc).unwrap();

    let mut cursor = Cursor::new(buf);

    let new_doc = decode_document(&mut cursor).unwrap();
    match new_doc.get("key") {
        Some(&Bson::UtcDatetime(ref datetime)) => {
            println!("{:?}", datetime);
            assert_eq!(1286705410, datetime.timestamp());  // panics on value '10'
            assert_eq!(2000000, datetime.nanosecond());
        },
        _ => panic!("!"),
    }
}

```

And passes :)

```
running 5 tests
test modules::encoder::test_encode_array ... ok
test modules::encoder::test_encode_decode_utc_date_time ... ok
test modules::encoder::test_encode_floating_point ... ok
test modules::encoder::test_encode_utf8_string ... ok
test modules::encoder::test_encode_utc_date_time ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured
```

matsimitsu:
Maybe it's a nice idea to create some sort of integration test that just encodes en decodes a document with all the various types supported just to make sure input equals output.

kyeah:
Oops, this is why you don't review PRs with a fever :)

LGTM. I agree that there needs to be some more strenuous testing -- I'll open up an issue for it.

PR #16, opened 2015-08-25
Helpers to directly get inner values

thijsc:
There's one annoyance left when working with the bson crate for our use case, getting to the inner typed values is quite verbose:

```rust
let truth = match document.get("truth") {
    Some(&Bson::Boolean(b)) => b,
    _ => panic!("truth is not a boolean")
};

let text = match document.get("text") {
    Some(&Bson::String(ref s)) => s,
    _ => panic!("text is not a string")
};
```

I've been thinking about a way to make this pattern more convenient. This pull is one way to go about it:

```rust
let truth = try!(document.get_bool("truth"));
let text = try!(document.get_str("text"));
```

Do you like this direction? If so I'll implement the remaining getters.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -32,7 +32,7 @@ use spec::{ElementType, BinarySubtype};
 use oid;
 
 /// Possible BSON value types.
-#[derive(Debug, Clone)]
+#[derive(Debug,Clone,PartialEq)]
 pub enum Bson {
     FloatingPoint(f64),
     String(String),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,7 @@ extern crate time;
 pub use self::bson::{Bson, Document, Array};
 pub use self::encoder::{encode_document, EncoderResult, EncoderError};
 pub use self::decoder::{decode_document, DecoderResult, DecoderError};
+pub use self::ordered::{ValueAccessError, ValueAccessResult};
 
 pub mod macros;
 pub mod oid;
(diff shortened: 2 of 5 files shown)


kyeah:
This seems like the most reasonable way to do it. @zonyitoo?

thijsc:
On a related note: I ported all usage of documents in the C wrapper driver to use the new macro: https://github.com/thijsc/mongo-rust-driver/commit/18301441252deb8791f25c6ff4e1d58b2c50555c

Been really happy on that end, kudos!

zonyitoo:
@kyeah Hmm, just like what rustc-serialize did. Well, I think that is reasonable. Merging.

zonyitoo:
Oh, wait, there still need to implement other getters.

thijsc:
Great, I'll finish this pull and post a comment here when that's done.

thijsc:
I've added more helpers, made the error publicly accessible and wrote tests.

Currently these are the only types that don't have a direct accessor:
* RegExp(String, String)
* JavaScriptCode(String)
* JavaScriptCodeWithScope(String, Document)
* Binary(BinarySubtype, Vec<u8>)

I'm not completely sure how to do this in a nice way for types that return a tuple. Shall we skip these, at least for now?

zonyitoo:
Hmm, but there are conflicts in this commit with the master branch, could you solve it?

thijsc:
done

PR #18, opened 2015-08-26
Allow string slices as keys for insertion

thijsc:
Another small ergonomics improvement. Before:

```rust
let mut doc = Document::new();
doc.insert("first".to_string(), Bson::I32(1));
```

After:
```rust
let mut doc = Document::new();
doc.insert("first", Bson::I32(1));
```

Also contains a minor performance improvement, removes one clone of the key.


zonyitoo:
Thank you very much!

Issue #35, opened 2016-03-05
Use linked-hash-map to implement OrderedDocument


zonyitoo:
The `ordered::OrderedDocument` is a temporary solution while the `linked-hash-map` crate can only be compiled with nightly. But now, [linked-hash-map](https://github.com/contain-rs/linked-hash-map) can be compiled with the latest stable rustc v1.7 (https://github.com/contain-rs/linked-hash-map/issues/28).

So I suggest that we could migrate it to `linked-hash-map`.

dariusc93:
How would this impact current use of bson?

zonyitoo:
Currently the `OrderedDocument` is implemented as `Vec` + `BTreeMap` pair to maintain insertion order. Which means that we need to do a `O(n)` search on the `Vec` and store the `String` key twice.

Issue #33, opened 2015-12-30
module `ordered` is private


kimhyunkang:
Is there a specific reason `ordered` module is private?

This makes convenient macros such as `try!` unusable.

For example:

```rust
use bson::Document;

fn set_and_get_age() -> Result<i32, MyCustomError> {
  let mut doc = Document::new();
  doc.insert("age".to_string(), 32 as i32);

  let value = try!(doc.get_i32("age"));
  Ok(value)
}
```

This code does not compile because `MyCustomError` does not implement `From<bson::ordered::ValueAccessError>`. You can't provide a custom impl for From, because the module is private.

zonyitoo:
Because the `ordered` is just a temporary solution for implementing ordered map. I am still waiting for the `linked-hash-map` to be stablized.

zonyitoo:
Merge with #35 .

PR #36, opened 2016-03-06
Use linked-hash-map to implement OrderedDocument

zonyitoo:
Discussion in #35 .

Request for comments!

### DIFF
--- a/src/ordered.rs
+++ b/src/ordered.rs
@@ -1,11 +1,15 @@
-use chrono::{DateTime, UTC};
-use std::collections::BTreeMap;
-use std::{error, fmt, slice};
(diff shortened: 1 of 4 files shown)


src/ordered.rs
```diff
@@ -1,11 +1,16 @@
-use chrono::{DateTime, UTC};
-use std::collections::BTreeMap;
-use std::{error, fmt, slice};
+// use std::collections::BTreeMap;
```
kyeah: Unused comment / import?

kyeah:
LGTM :)

kyeah:
Actually, we should make the ordered module public via #33.

PR #37, opened 2016-03-27
Show object id formatted as string in debug trait and derive Hash

thijsc:
I found it a bit inconvenient sometimes that I could not copy-paste object id's from debug output. This pull fixes that.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -88,7 +88,7 @@ impl error::Error for Error {
 }
 
 /// A wrapper around raw 12-byte ObjectId representations.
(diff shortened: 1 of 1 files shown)


thijsc:
I've also added a derive for `Hash`, it would be nice to use `ObjectId`' as a key in a `HashMap`. It's a small change on the same line so a separate pull would immediately be a merge conflict.

zonyitoo:
Hmm, prints the ObjectId in its Hex form looks better indeed. :)

thijsc:
Thanks, could you bump the crate version too?

zonyitoo:
Of course.

zonyitoo:
Done already. Cheers.

Issue #38, opened 2016-08-03
Test count_is_big_endian sometimes fails


dtolnay:
It seems to work 2 out of 3 times. It fails with:

```
thread 'oid::count_is_big_endian'
panicked at 'assertion failed: `(left == right)` (left: `17`, right: `0`)', src/oid.rs:333
```

dtolnay:
It works consistently with `RUST_TEST_THREADS=1 cargo test` so there is a race condition somewhere when tests are run in parallel.

kyeah:
If I had to guess, `count_generation` and `count_is_big_endian` are conflicting by setting `OID_COUNTER` to different start values. 

Looking at them now, I'm not sure if they are even testing different things -- we could probably combine them into one test and avoid having to restructure the tests/struct.

dtolnay:
Some more failures:

```
---- oid::count_is_big_endian stdout ----
	thread 'oid::count_is_big_endian' panicked at 'assertion failed: `(left == right)` (left: `51`, right: `52`)', src/oid.rs:335
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- oid::count_generation stdout ----
	thread 'oid::count_generation' panicked at 'assertion failed: `(left == right)` (left: `52222`, right: `1122867`)', src/oid.rs:322
```

kyeah:
Yup, would definitely seem like they're conflicting. I'm pretty busy in the next couple of weeks, but if you or anyone else wants to take on combining both tests or finding another option to avoid setting a shared OID_COUNTER, let me know and go for it.

saghm:
Given that `OID_COUNTER` will never actually be initialized with two different starting values in practice, would it be an issue if the tests were changed to use the same initial count? Barring that, would it acceptable to just enforce that our tests should be run single-threaded? (We have to do that in the Mongo driver due to similar issues with accessing the database in parallel).

kyeah:
The issue would remain, since they both increment OID_COUNTER, you'll never be sure if the counter is at the initial count or the other test came in and incremented after you set it.

 Considering this is the only case where we have a parallel issue, I think we should strive for parallel tests; the mongo tests are much more complex since they deal with live databases.

saghm:
[...]

PR #39, opened 2016-08-03
Update to serde 0.8

dtolnay:
The biggest change is the new Serializer API which provides a better way to handle what used to be `Vec<State>` inside `Encoder`.

Before:

```rust
#[inline]
fn serialize_seq<V>(&mut self, mut visitor: V) -> EncoderResult<()>
    where V: SeqVisitor,
{
    let len = visitor.len().unwrap_or(0);
    let values = Vec::with_capacity(len);
    self.state.push(State::Array(values));

    while let Some(()) = try!(visitor.visit(self)) { }

    let values = match try!(self.pop()) {
        State::Array(values) => values,
        state => return Err(EncoderError::InvalidState(state)),
    };

    self.state.push(State::Bson(Bson::Array(values)));
    Ok(())
}

#[inline]
fn serialize_seq_elt<T>(&mut self, value: T) -> EncoderResult<()>
    where T: Serialize,
{
    try!(value.serialize(self));

    let value = match try!(self.pop()) {
        State::Bson(value) => value,
        state => return Err(EncoderError::InvalidState(state)),
    };

    match self.state.last_mut() {
        Some(&mut State::Array(ref mut values)) => values.push(value),
        Some(state) => return Err(EncoderError::InvalidState(state.clone())),
        None => return Err(EncoderError::EmptyState),
    }

    Ok(())
}
```

After

```rust
#[inline]
fn serialize_seq(&mut self, len: Option<usize>) -> EncoderResult<Array> {
    Ok(Array::with_capacity(len.unwrap_or(0)))
}

#[inline]
fn serialize_seq_elt<T>(&mut self,
                        state: &mut Array,
                        value: T) -> EncoderResult<()>
    where T: Serialize,
{
    state.push(try!(to_bson(&value)));
    Ok(())
}

#[inline]
fn serialize_seq_end(&mut self, state: Array) -> EncoderResult<()> {
    self.value = Bson::Array(state);
    Ok(())
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ libc = "0.1"
 rand = "0.3"
 rust-crypto = "0.2.31"
 rustc-serialize = "0.3"
-serde = "0.7"
+serde = "0.8"
 time = "0.1"
 linked-hash-map = "0.0.9"
 
--- a/src/decoder/serde.rs
+++ b/src/decoder/serde.rs
@@ -145,6 +145,43 @@ impl Decoder {
     }
 }
 
+macro_rules! forward_to_deserialize {
+    ($(
+        $name:ident ( $( $arg:ident : $ty:ty ),* );
+    )*) => {
+        $(
+            forward_to_deserialize!{
+                func: $name ( $( $arg: $ty ),* );
+            }
+        )*
+    };
+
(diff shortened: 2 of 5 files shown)


zonyitoo:
Will publish it after #38 is solved properly

PR #41, opened 2016-08-27
Combine oid counter generation tests

kyeah:
Hey! This is a very lazy response to #38 to fix the OID_COUNTER race condition that occurs between `count_generation` and `count_is_big_endian`. There may be a nicer way that keeps the two tests separate, but let's get this out of the way so we can release v0.8.

As a bonus, this will also test that OID_COUNTER is incremented on `gen_count`.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -306,9 +306,11 @@ fn pid_generation() {
 }
 
 #[test]
-fn count_generation() {
-    let start = 52222;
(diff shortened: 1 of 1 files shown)


kyeah:
cc. @zonyitoo . since you're the only crate owner, i'll let you merge this one in and push the new version up to crates 👍

zonyitoo:
Of course! Thanks!

PR #42, opened 2016-09-22
Fix JavaScriptCodeWithScope encoding and decoding

alanhdu:
In the bson spec, a `code_w_s` object is defined as:
```
code_w_s 	::= 	int32 string document 	Code w/ scope
```

On the decoding side, that means we have to read off an `i32` before reading `string` or `document`. On the encoding side, that means we have to add `+4` to the buffer length instead of `+1` (because an i32 is 4 bytes). I tested that this behavior matches up with the golang mongo driver (mgo).


zonyitoo:
Ah, yes. That's true. Thanks for the PR.

zonyitoo:
I will release this today as v0.3.1

PR #56, opened 2017-02-25
Fix json f64 value conversion

kyeah:
Hello! After switching over to serde_json, I noticed a small discrepancy in converting numerical values to Bson. 

Looking at the [Number code](https://github.com/serde-rs/json/blob/master/json/src/number.rs#L55-L63), `as_i64` will cast unsigned values to i64, while floating points will return `None`, so we are missing the floating-point conversion here.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -320,7 +320,7 @@ impl Bson {
         match j {
             &Value::Number(ref x) =>
                 x.as_i64().map(Bson::from)
-                .or(x.as_f64().map(Bson::from))
+                .or_else(|| x.as_f64().map(Bson::from))
                 .expect(&format!("Invalid number value: {}", x)),
             &Value::String(ref x) => x.into(),
             &Value::Bool(x) => x.into(),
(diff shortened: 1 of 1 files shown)


src/bson.rs
```diff
@@ -320,7 +320,7 @@ impl Bson {
         match j {
             &Value::Number(ref x) =>
                 x.as_i64().map(Bson::from)
-                .or(x.as_u64().map(Bson::from))
+                .or(x.as_f64().map(Bson::from))
```
zonyitoo: One historical bug, we should use `.or_else(|_| x.as_f64().map(Bson::from)` to prevent unnecessary method calls.

zonyitoo:
Brilliant.

zonyitoo:
@kyeah please upgrade bson-rs to v0.4.3. I am going to publish these changes.

kyeah:
@zonyitoo upgrading.

kyeah:
ah, you got it. thanks! 🙏

zonyitoo:
Already published. What I meant "upgrade" was to upgrade Mongo driver's dep

PR #57, opened 2017-03-01
Bson::Document also could convert to Document

Matrix-Zhang:


src/bson.rs
```diff
@@ -386,6 +386,7 @@ impl Bson {
                     "$symbol" => (v.to_owned())
                 }
             }
+            Bson::Document(ref v) => v.to_owned(),
```
zonyitoo: I don't think this is a correct implementation. You must at least convert each values to its extended form, instead of cloning the `Document` itself.

Specification: https://docs.mongodb.com/manual/reference/mongodb-extended-json/

zonyitoo:
Close for incorrect implementation.

PR #59, opened 2017-03-01
add getter helpers for Bson, correct implementation for #57

zonyitoo:
Also add more useful documents.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -35,21 +35,41 @@ use spec::{ElementType, BinarySubtype};
 /// Possible BSON value types.
 #[derive(Clone, PartialEq)]
 pub enum Bson {
+    /// 64-bit binary floating point
     FloatingPoint(f64),
+    /// UTF-8 string
     String(String),
+    /// Array
     Array(Array),
+    /// Embedded document
     Document(Document),
+    /// Boolean value
     Boolean(bool),
+    /// Null value
     Null,
+    /// Regular expression - The first cstring is the regex pattern, the second is the regex options string.
+    /// Options are identified by characters, which must be stored in alphabetical order.
+    /// Valid options are 'i' for case insensitive matching, 'm' for multiline matching, 'x' for verbose mode,
+    /// 'l' to make \w, \W, etc. locale dependent, 's' for dotall mode ('.' matches everything), and 'u' to
+    /// make \w, \W, etc. match unicode.
     RegExp(String, String),
+    /// JavaScript code
     JavaScriptCode(String),
+    /// JavaScript code w/ scope
     JavaScriptCodeWithScope(String, Document),
+    /// 32-bit integer
     I32(i32),
+    /// 64-bit integer
     I64(i64),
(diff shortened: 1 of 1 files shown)


zonyitoo:
Hmm, reasonable. `TryFrom` or `TryInto` are more elegant.

`to/from_extended_document` are in the `impl` block above.

zonyitoo:
@kyeah Is it Ok?

src/bson.rs
```diff
@@ -430,3 +466,121 @@ impl Bson {
...
+    }
+}
+
+/// Type alias of `Result<T, GetBsonValueError>`
+pub type GetBsonValueResult<T> = Result<T, GetBsonValueError>;
+
+/// Value helpers
+impl Bson {
+    /// If `Bson` is `FloatingPoint`, return its value. Returns `None` otherwise
+    pub fn as_f64(&self) -> GetBsonValueResult<f64> {
```
kyeah: in the interest of predictability and consistency with From/Into and TryFrom/TryInto, perhaps we should have TryAs?

```rust
pub trait TryAs<T> {
    type Err;
    fn try_as(&self) -> Result<T, Self::Err>;
}

bson = val.from();
into_val: ObjectId = bson.try_into()?;
into_val: &ObjectId = bson.try_as()?;
```

src/bson.rs
```diff
@@ -430,3 +466,121 @@ impl Bson {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("GetBsonValueError")
+    }
+}
+
+impl Error for GetBsonValueError {
+    fn description(&self) -> &str {
+        "failed to get bson value"
+    }
+}
```
zonyitoo: Reasonable. Hmm.

src/bson.rs
zonyitoo: Hmm. Well, I use `as_*` is because JSON libraries are using this name style.

But yeah, `try_as_f64` is indeed better.

src/bson.rs
```diff
@@ -430,3 +466,121 @@ impl Bson {
         Bson::Document(values)
     }
 }
+
+/// Error for representing failure of getting BSON value
+#[derive(Debug)]
```
zonyitoo: Sure.

src/bson.rs
kyeah: ah! didn't realize this was in response to serde-json. I'd be fine with matching their API then, with `as_X(&self) -> Option<X>`, and adding `TryInto` for consuming conversions once that trait becomes stable.

zonyitoo:
@kyeah Reverted. Is it Ok this time? :)

zonyitoo:
Excellent! Will be published to crates.io right now.

zonyitoo:
Already published as `v0.4.4`, please upgrade deps in mongo driver. @kyeah 

cc @Matrix-Zhang

kyeah:
published mongodb 0.2.3.

PR #61, opened 2017-03-11
Lowercase ObjectId string

zonyitoo:


zonyitoo:
Published as `v0.5.0`, :)

PR #63, opened 2017-03-12
Implement From/Into traits for Value

kyeah:
This makes things a little more convenient by allowing us to call `bson_val.into()`, `json_val.into()`, etc, and giving us a more performant, consuming transformation method.

This also allows us to plug bson values into any interface that relies on `From<Value>` or `Into<Value>`.

Keeping `from_json`, `to_json` for backwards-compatibility, and adding an appropriate `into_json` consumer.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -263,6 +263,75 @@ impl From<DateTime<UTC>> for Bson {
     }
 }
 
+impl From<Value> for Bson {
+    fn from(a: Value) -> Bson {
+        match a {
+            Value::Number(x) =>
+                x.as_i64().map(Bson::from)
+                .or(x.as_u64().map(Bson::from))
+                .expect(&format!("Invalid number value: {}", x)),
+            Value::String(x) => x.into(),
+            Value::Bool(x) => x.into(),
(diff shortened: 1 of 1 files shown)


zonyitoo:
I think we should keep that `to_json` method for backward compatibility.

kyeah:
`to_json` is included with the same functionality (a non-consuming conversion to serde_json Value), it just uses `self.clone().into()` instead. It's near the bottom of the github diff; the diff doesn't show it clearly :/

zonyitoo:
All looks good. Maybe you can rebase those two commits into one. :)

zonyitoo:
Oh, btw, is it Ok to add a `#[deprecated]` mark on `to_json` and `from_json`?

kyeah:
yeah, I can add a deprecated warning!

re: rebasing, you can Squash and Merge PRs into one commit now, if that's cool with you :)

zonyitoo:
I didn't realize github supports Squash and Merge PRs! Thanks.
I will publish it later.

zonyitoo:
Published in v0.6

PR #67, opened 2017-04-22
make compatibility changes for serde 1.0

saghm:
Serde 1.0 was released earlier this week! Unfortunately, the API changed a bit, so there are a few changes that needed to be made; the `XxxVisitor` traits were renamed to `XxxAccess`, along with their methods changing to not have the word "visitor" in them, lifetime parameters were added to basically everything in `serde::de`, `Deserializer` added one new method and removed a couple old ones, as well as renaming the `deserialize()` method to `deserialize_any()`, `Serializer` removed one method and changed all existing methods to `u32` for the index rather than `usize` (which seems kind of weird, but I assume they had a good reason for it), `size_hint` was changed to only return `Option<usize>` rather than  rather than `(usize, Option<usize>)`, and the module `serde::de::impls` was made private, which required an implementation of `visit_seq` to be changed to manually iterate over the values rather than  use `VecVisitor`.

I also noticed that one of tests is still using the `to_json()` method that was deprecated recently, but that seemed a bit out of scope for this PR, although it should still probably be fixed.

EDIT: Just realized there's a separate subcrate (sub-crate?) for the serde tests. Because serde_codegen seems to have been deprecated in favor of serde_derive (which uses macros 1.1 from Rust 1.15+), I've updated the tests to use that, which has the benefit of no longer requiring a janky build script to generate a test file at compile time.


zonyitoo:
Every minor version changes of serde are huge breaking changes!

LGTM. What's your opinion? @kyeah

zonyitoo:
One of the tests failed on my laptop

```
test modules::oid::increasing ... FAILED

failures:

---- modules::oid::increasing stdout ----
       	thread 'modules::oid::increasing' panicked at 'assertion failed: oid1 < oid2', tests/modules/oid.rs:87
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    modules::oid::increasing

test result: FAILED. 37 passed; 1 failed; 0 ignored; 0 measured
```

But it seems to occur occasionally. I remember that happened before. Is it Ok? @kyeah

kyeah:
oh, yes! we never _really_ fixed that parallel-testing issue, but this should still be good to go.

zonyitoo:
Published in `v0.7.0`

Issue #70, opened 2017-05-17
BSON likes Strings and hates u32/usize


TrionProg:
Hello. I have struct 
```
pub struct Award{
    pub name:String,
    pub icon:Uuid,
    pub description:String,
    pub id:String,
}
```
And I write it to mongodb
```
                let award=Award{
                //id:id as u32,
                name:String::from(award_name),
                icon:icon,
                description:description,
                id:id .to_string(),
            };

            let serialized_award = bson::to_bson(&award)?;

            let find_filter=doc! {
                "_id" => user_id
            };

            let update_doc=doc! {//All right?
                "$push" => { "awards" => serialized_award }
            };

            mongo_users.find_one_and_update( find_filter ,update_doc, None );//TODO:None
```

And later I read:

```
let doc=mongo_users.find_one(Some(find_filter),Some(find_option))?;

        println!("{:?}",doc);

        match doc {
            Some( doc ) => {
               match doc.get("awards") {
                Some(&Bson::Array(ref mongo_awards)) => {

                let mut awards=Vec::with_capacity(mongo_awards.len());
                for mongo_award in mongo_awards.iter() {
                    match *mongo_award {
                        Bson::Document( ref doc ) => {
                            let award:Award = bson::from_bson( bson::Bson::Document(doc.clone()) ).unwrap();//NOTE : problem, and I DO NOT WANT CLONE document!
```

Part of mongo document
```
"awards": Array([Document({ name: "abc", icon: "a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11", description: "def", id: 1 })])
```

If id of` struct Award` is String, all works, but if id is u32 or usize I get `InvalidType("u32")`
Note that Display trait for this error writes only `usize`,`u32` without any description like `InvalidType:usize`

UPD. Struct initialization in exemple gets &'static str instead String. This code do not works!
```
#[derive(Serialize, Deserialize, Debug)]
        pub struct Person {
            #[serde(rename = "_id")]  // Use MongoDB's special primary key field name when serializing
            pub id: String,
            pub name: String,
            pub age: u32
        }

        let person = Person {
            id: "12345".to_string(),
            name: "Emma".to_string(),
            age: 3
        };

        let serialized_person = bson::to_bson(&person)?;  // Serialize

        let person2:Person = bson::from_bson(serialized_person).unwrap();
}
```

Cargo.toml
```
serde="1.0.2"
serde_derive="*"
bincode="*"
redis = "*"
[...]

PR #71, opened 2017-05-21
Force convert unsigned type to Bson signed type

zonyitoo:
* Serialize `u32` to `Bson::I32`
* Serialize `u64` to `Bson::I64`
* Serialize `u16` to `Bson::I32`
* Serialize `u8` to `Bson::I32`

Ref: #70

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -85,7 +85,7 @@ impl Debug for Bson {
             &Bson::FloatingPoint(p) => write!(f, "FloatingPoint({:?})", p),
             &Bson::String(ref s) => write!(f, "String({:?})", s),
             &Bson::Array(ref vec) => write!(f, "Array({:?})", vec),
-            &Bson::Document(ref doc) => write!(f, "Document({})", doc),
+            &Bson::Document(ref doc) => write!(f, "Document({:?})", doc),
             &Bson::Boolean(b) => write!(f, "Boolean({:?})", b),
             &Bson::Null => write!(f, "Null"),
             &Bson::RegExp(ref pat, ref opt) => write!(f, "RegExp(/{:?}/{:?})", pat, opt),
--- a/src/decoder/serde.rs
+++ b/src/decoder/serde.rs
@@ -62,39 +62,73 @@ impl<'de> Visitor<'de> for BsonVisitor {
     }
 
     #[inline]
-    fn visit_bool<E>(self, value: bool) -> Result<Bson, E> {
+    fn visit_bool<E>(self, value: bool) -> Result<Bson, E>
+        where E: de::Error
+    {
         Ok(Bson::Boolean(value))
     }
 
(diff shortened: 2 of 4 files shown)


saghm:
Wouldn't the 32- and 64-bit cases have the potential to overflow? I'm not sure what the correct solution for u64 is, but at the very least, I think that u32 should be serialized to i64, or else numbers larger 2^31 - 1 will not work as expected.

zonyitoo:
Well, `i32` and `u32` has exactly the same size, so I think it is Ok to save `u32` value in a `i32` type. BSON doesn't support any unsigned types in specification, but just like other languages, such as Java, there are still lots of ways to work around it.

kyeah:
@saghm is correct; unsigned 32-bit ints can hold larger numbers than signed 32-bits, since no bits need to be reserved to indicate the sign value. Converting them directly to same-bit signed integers has the potential for overflow.

The root issue lies with the fact that any sort of support for unsigned values on the driver-side will require conversions to supported types, and we shouldn't be making implicit conversion choices on our end for the user. Even if we implement hacks like serializing u32s as i64s, it's unclear to the user that this is happening, and the Rust struct they are using is the only source of truth of whether that value should be a u32 or an i64. 

This is especially problematic when you consider that many different applications may read from that database/dataset; consider a user who may not know that mongodb doesn't support unsigned types, and does not realize they have been writing i64s to the database until they read that data back from another app.

The correct approach is to drop support for auto-serializing unsigned values entirely (including converting them to floating points); rather, we should provide informative errors + custom field serializers so that the user:

1. understands that unsigned values are not supported / directly written to mongodb.
2. is able to remediate and _opt in_ to a number of different conversion options, depending on their application context and requirements. Those options include: promotions to a signed value with a higher bit count, or using FloatingPoint if they have u64s and don't require exact-precision values.

```rust
#[derive(Serialize)]
struct MyThinger {
    #[serde(serialize_with = "bson::serde::serialize_i64")]
    id: u32,
}
```

See documentation for: [field attributes](https://serde.rs/field-attrs.html) and [type bound example using deserialize_with](https://serde.rs/attr-bound.html).

zonyitoo:
Hmm, so we should explicitly deny serializing unsigned types to BSON.

marcelgarus:
> ```rust
> #[serde(serialize_with = "bson::serde::serialize_i64")]
>     id: u32,
> ```

I'm trying to achieve exactly this. Is there an actual easy solution to serialize `u32` as `i64`? I can't find serializer methods anywhere in the `bson` crate…

flisky:
Are you looking for [serialize_u32_as_i64](https://docs.rs/bson/1.2.2/bson/serde_helpers/fn.serialize_u32_as_i64.html), @marcelgarus ?

marcelgarus:
Ah yes, thank you

PR #72, opened 2017-05-22
Disallow serde serialize/deserialize unsigned integer types

zonyitoo:
* Disallow `u16`, `u32` and `u64` to be serialize to / deserialize from BSON

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ This crate works with Cargo and can be found on
 
 ```toml
 [dependencies]
-bson = "0.7"
+bson = "0.8"
 ```
+
 ## Usage
 Link the library in _main.rs_:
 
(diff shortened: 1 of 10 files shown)


kyeah:
Should we replace these with functions that users can refer to, so people can still keep the original functionality if necessary?

```rust
// something like this -- pseudocode!

#[inline]
fn serialize_u64_into_f64(serializer: Encoder, value: u64) -> EncoderResult<Bson> {
    serializer.serialize_f64(value as f64)
}

#[derive(Serialize, Debug)]
struct Thing {
    #[serde(serialize_with = "bson::encoder::serialize_u64_into_f64")]
    s: u64,
}
```

zonyitoo:
Hmm, maybe. But I don't think it is correct to serialize `u64` to `f64`.

It is reasonable to do this for compatibility.

zonyitoo:
Added backward compatibility support mod `u2f`

```rust
#[test]
fn test_compat_u2f() {
    #[derive(Serialize, Deserialize, Eq, PartialEq, Debug)]
    struct Foo {
        #[serde(with = "bson::compat::u2f")]
        x: u32
    }

    let foo = Foo { x: 20 };
    let b = bson::to_bson(&foo).unwrap();
    assert_eq!(b, Bson::Document(doc! { "x" => (Bson::FloatingPoint(20.0)) }));

    let de_foo = bson::from_bson::<Foo>(b).unwrap();
    assert_eq!(de_foo, foo);
}
```

zonyitoo:
@kyeah How about this?

README.md
```diff
@@ -68,3 +69,29 @@ let person_document = mongoCollection.find_one(Some(doc! { "_id" => "12345" }),
 // Deserialize the document into a Person instance
 let person = bson::from_bson(bson::Bson::Document(person_document))?
 ```
+
+## Breaking Changes
+
+In BSON specification, it doesn't support any kind of _unsigned integer types_, for example, `u32`. In the older version of this crate (< `v0.8.0`), if you uses `serde` to serialize _unsigned integer types_ into BSON, it will store them with `Bson::FloatingPoint` type. From `v0.8.0`, we removed this behavior and simply returned an error when you want to serialize _unsigned integer types_ to BSON. [#72](https://github.com/zonyitoo/bson-rs/pull/72)
```
kyeah: "In the BSON specification, _unsigned integer types_ are unsupported; for example"...

README.md
```diff
@@ -68,3 +69,29 @@ let person_document = mongoCollection.find_one(Some(doc! { "_id" => "12345" }),
...
+
+For backward compatibility, we provided a mod `bson::compat::u2f` to explicitly serialize _unsigned integer types_ into BSON's floating point value as follows:
```
kyeah: `we` => `we've`

README.md
```diff
@@ -68,3 +69,29 @@ let person_document = mongoCollection.find_one(Some(doc! { "_id" => "12345" }),
...
+    let foo = Foo { x: 20 };
+    let b = bson::to_bson(&foo).unwrap();
+    assert_eq!(b, Bson::Document(doc! { "x" => (Bson::FloatingPoint(20.0)) }));
+
+    let de_foo = bson::from_bson::<Foo>(b).unwrap();
+    assert_eq!(de_foo, foo);
+}
+```
+
+In this example, we added an attribute `#[serde(with = "bson::compat::u2f")]` on field `x`, which will tell `serde` to uses the `bson::compat::u2f::serialize` and `bson::compat::u2f::deserialize` to process this field.
```
kyeah: `uses` => `use`

"the `bson::compat::u2f::serialize` and `bson::compat::u2f::deserialize` **methods**"

zonyitoo:
Already done.

kyeah:
awesome, LGTM!

zonyitoo:
Already publish to crates.io.

PR #76, opened 2017-06-17
add function to get length when decoding bson document

saghm:
Currently, there isn't any way to get the length of a BSON document when decoding it . I rediscovered that we're doing an [inefficient hack](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/58d28bbec006b42f509812cdfd7c92bfb18cb843/src/wire_protocol/operations.rs#L25) in the MongoDB driver to be able to get the byte length of documents, so adding this functionality should allow us to make some optimizations (i.e. not [re-encode everything we decode](https://github.com/mongodb-labs/mongo-rust-driver-prototype/blob/58d28bbec006b42f509812cdfd7c92bfb18cb843/src/wire_protocol/operations.rs#L529)).

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -28,6 +28,7 @@ pub use self::error::{DecoderError, DecoderResult};
 pub use self::serde::Decoder;
 
 use std::io::Read;
+use std::mem::size_of;
 
 use byteorder::{LittleEndian, ReadBytesExt};
 use chrono::UTC;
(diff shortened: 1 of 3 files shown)


zonyitoo:
I don't think this is a clean solution. If you want to know exactly how many bytes have been used for decoding, you can simply make a wrapper for the reader.

```rust
struct ReadWrapper<R: Read> {
    r: R,
    read: usize,
}

impl<R: Read> Read for ReadWrapper<R> {
    fn read(&mut self, buf: &mut self) -> io::Result<usize> {
        let size = self.r.read(buf)?;
        self.read += size;
        Ok(size)
    }
}
```

saghm:
Yep, looks like I completely overengineered this. Thanks for the help!

PR #83, opened 2017-10-01
Improve macro syntax support

kyeah:
## What

Overhauls the bson/doc macro definitions to support syntactic sugar available in the `json!` macro:

- [x] colon-separated key-value pairs
- [x] Non-parenthesized values with `.` characters
- [x] Trailing commas
- [x] `null` keyword

## How

By taking serde's `json!` macro and making a couple changes (including overloaded support for rocket separators for backwards-compatibility), and some simplifications (removing redundant boolean cases, etc). I ran the mongo driver tests with this branch of `bson` and they all passed, which is convincing enough for me (along with our bson-rs tests) that we don't have any regressions.

## Who

@zonyitoo, @saghm


src/bson.rs
```diff
@@ -397,19 +397,19 @@ impl Bson {
         match *self {
             Bson::RegExp(ref pat, ref opt) => {
                 doc! {
-                    "$regex" => (pat.clone()),
```
saghm: I'm so confused about git saying the first ')' was removed here but the second below

src/bson.rs
```diff
@@ -397,19 +397,19 @@ impl Bson {
...
-                    "$options" => (opt.clone())
+                    "$regex" => pat.clone(),
```
saghm: Out of curiosity, why remove the parens/add trailing commas but not switch the "=>" to ":"? Is this a rustfmt thing?

src/macros.rs
```diff
@@ -1,47 +1,250 @@
...
+
+/// Construct a bson::BSON value from a literal.
+///
+/// ```rust
+/// # #[macro_use]
+/// # extern crate bson;
+/// #
+/// # fn main() {
+/// let value = bson!({
+///     "code": 200,
```
saghm: Do we want to document the "=>" syntax, or are we considering that discouraged (but kept around for compatibility)?

src/macros.rs
```diff
@@ -1,47 +1,250 @@
...
+///           "loads",
+///       ]
+///     }
+/// });
+/// # }
+/// ```
 #[macro_export]
 macro_rules! bson {
-    ([]) => {{ $crate::Bson::Array(Vec::new()) }};
+    //////////////////////////////////////////////////////////////////////////
```
saghm: I assume these comments are also lifted from serde_json?

tests/lib.rs
```diff
@@ -27,27 +27,75 @@ fn test_format() {
     let doc = doc! {
         "float" => 2.4,
         "string" => "hello",
-        "array" => ["testing", 1],
+        "array" => ["testing", 1, true, [1, 2]],
         "doc" => {
             "fish" => "in",
             "a" => "barrel",
-            "!" => 1
+            "!" => 1,
```
saghm: Is the trailing comma required? If not, we should probably add a separate test for without it (both nested and top-level)

tests/lib.rs
```diff
@@ -27,27 +27,75 @@ fn test_format() {
...
+    let date = Utc::now();
+
+    let doc = doc! {
+        "float": 2.4,
+        "string": "hello",
+        "array" => ["testing", 1, true, [1, 2]],
+        "doc": {
+            "fish": "in",
+            "a": "barrel",
+            "!": 1,
```
saghm: Ditto above about additional tests for without trailing commas

zonyitoo:
LGTM

src/bson.rs
kyeah: nope, just made these changes before I re-added colon support. i'll change these.

src/macros.rs
kyeah: Considering it discouraged in favor of consistency, but here for backwards-compatibility.

saghm: Sounds good

src/macros.rs
kyeah: yeah; not sure where the line falls in terms of fair usage.

kyeah:
@zonyitoo ready for a publish. thanks 🙏

zonyitoo:
Published v0.10.0

PR #86, opened 2018-01-05
Add decode_document_utf8_lossy

thijsc:
MongoDB has a bug where it can return bson responses that contain strings with invalid UTF-8 encoding: https://jira.mongodb.org/browse/SERVER-24007

This bug has been open since 2016 and no fix is in sight, so we need a workaround in the `mongo_driver` crate. This commit adds a separate decode function that allows you to get lossy UTF-8 conversion. We will use this function when parsing error responses from Mongo.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -39,11 +39,18 @@ use oid;
 
 use serde::de::Deserialize;
 
-fn read_string<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<String> {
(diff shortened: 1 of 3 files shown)


thijsc:
Thanks! I saw you pushed a commit to update to 0.11, but it's not on crates.io yet. Could you publish?

kyeah:
@zonyitoo

zonyitoo:
Ah, of course. I will do it tonight.

zonyitoo:
Published `v0.11.0`

PR #87, opened 2018-01-28
add entry API for Document

saghm:
Since the last time the `linked-hash-map` dependency was updated, it added an entry API to mimic that of the standard library for modifying an entry in the map in-place to get around borrow checker issues when trying to get and insert at the same time. This PR adds a thin wrapper around this functionality for the underlying map used by a `Document`.

I attempted to implement the `entries` method as well, but due to [this](https://github.com/contain-rs/linked-hash-map/pull/89) not yet being released on crates.io, it wasn't working, so I didn't include it in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ rust-crypto = "0.2"
 serde = "1.0"
 serde_json = { version = "1.0.0", features = ["preserve_order"] }
 time = "0.1"
-linked-hash-map = "0.3"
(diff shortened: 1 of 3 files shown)


zonyitoo:
LGTM.

zonyitoo:
Should I publish these changes into crates.io?

saghm:
It's not urgent or anything, so whenever you're comfortable with pushing is
fine

PR #88, opened 2018-02-11
add TimeStamp type to facilitate deserialization

saghm:
Currently, there is no good way to deserialize the BSON timestamp type. The only thing that works is deserializing into the `Bson` enum, but that doesn't provide any compile-time guarantees that the value is actually a timestamp type. Following the example of `bson::UtcDateTime`, I implemented a timestamp type that can be used when deserializing.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -595,6 +595,27 @@ impl Bson {
     }
 }
 
+/// `TimeStamp` representation in struct for serde serialization
+///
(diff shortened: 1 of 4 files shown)


zonyitoo:
LGTM. cc @kyeah

kyeah:
@zonyitoo you can version-bump and publish at your convenience.

PR #90, opened 2018-03-17
Add feature-gated support for casting unsigned integers to signed integers

Ameobea:
When dealing with arbitrary inputs such as those from `serde_json::Value`, making use of the existing backwards compatibility APIs can be difficult or impossible.  This PR introduces the ability to cast unsigned integers to `i32` using the nightly `try_from` API.  It exposes this behind an opt-in feature and doesn't make any backwards-incompatible changes.

If you'd rather avoid this conversion strategy and stick to strict type checking, that's fine; I'm making use of these for my own project and understand if you'll want to keep them out of upstream.

I also changed all instances of `try!()` to `?`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,3 +29,6 @@ hex = "^0.2"
 
 [dev-dependencies]
 serde_derive = "1.0"
+
+[features]
+unsigned_conversion = []
--- a/README.md
+++ b/README.md
@@ -7,10 +7,12 @@
 Encoding and decoding support for BSON in Rust
 
 ## Useful links
-- [API Documentation](https://docs.rs/bson/)
-- [Serde](https://serde.rs/)
+
+* [API Documentation](https://docs.rs/bson/)
+* [Serde](https://serde.rs/)
 
 ## Installation
+
 This crate works with Cargo and can be found on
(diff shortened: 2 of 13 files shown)


zonyitoo:
LGTM. cc @kyeah

kyeah:
Hi, sorry for the delay @Ameobea, @zonyitoo!

My intuition is that users should opt into these conversions for unsigned fields, in the same way that we have this compat module for de/serializing unsigned integers: https://github.com/zonyitoo/bson-rs/blob/master/src/compat/u2f.rs.

That is, we should add `to_i32` and `from_i32` conversions (along with _i64) instead of adding a global conversion flag — and if we do decide on a global flag, we should provide options for TryInto<i64> and TryInto<f64>.

What do y'all think?

kyeah:
Hi @Ameobea! This question popped up again in #94, and it was noted that a global feature switch is useful to have when dealing with generated code. 

#95 was opened and merged before I remembered that this was still open — sorry! I'm going to close this for now since the feature has been merged as `u2i`, but the `try!` conversions are still useful, and would be welcome in a separate PR.

Note that the `u2i` implementation is a little different — in particular I believe it converts u64 to i64 if possible.

Thank you for putting this PR together!

Issue #94, opened 2018-06-26
Encode u32 as i64


marcelbuesing:
What do you think about encoding `u32` as `i64`?
I currently have the issue that I basically generate my Rust code from protobuf definitions, which I then want to decode and encode using bson. Now changing all protobuf defintions to `i32` just because bson does not support unsigned ints does not seem really reasonable to me. I know this does not cover the `u64`, downside I guess is here the inconsistency.

```
fn serialize_u32(self, value: u32) -> EncoderResult<Bson> {
        Ok(Bson::I64(value as i64))
}
```

kyeah:
Thanks for the issue! We want to avoid making implicit casting decisions like this, but I do agree that in some cases, it's valuable to have these conversions as optional features. We have a u2f compat module (see documentation [here](https://github.com/zonyitoo/bson-rs#breaking-changes)), and these could be expanded to `u2i`.

marcelbuesing:
Probably not a bad idea. I kind of need something though that requires no modification of the structs. Since the code is generated I can not add `#[serde(with = "bson::compat::u2i")]` to every field. 
Taking the already implemented `u2f` as an example is there already a way to encode unsigned as float without adding the attibute? It not, would it be feasible to add a function similar to `to_bson_compat(&x, CompatOpt::U2F)`which defines a "global" unsigned encoding policy?

kyeah:
Currently the `u2f` attribute is the only way to do this. A `to_bson_compat` function may be possible, but if not, perhaps a global feature flag is useful to enable this behavior by default.

zonyitoo:
Hmm, enabling it with a feature flag is better.

reitermarkus:
I'm facing a similar problem, but in the other direction, i.e. deserializing unsigned types to signed types. Could this be added as part of the `u2i` feature?

There is also different behaviour between `from_value` and `from_str`, here's a minimal example

```rust
use bson::Bson;
use serde_json::Value;

fn main() {
  assert_eq!(Bson::from(serde_json::from_str::<Value>("6").unwrap()), Bson::I64(6));
  assert_eq!(serde_json::from_str::<Bson>("6").unwrap(), Bson::I64(6));
}
```

which results in a `invalid type: integer `6`, expected a signed integer` error with the `from_str` call.

PR #95, opened 2018-07-03
Add i2u feature (#94)

marcelbuesing:
Basically as discussed in the issue this adds a feature flag for encoding unsigned types in signed types.
Open question: does this require further changes in the decoder. Given the tests it seems to work without changes in the decoder.
I also could not figure out how to conditionally add a test for the case that the "u2i" is not set. Hints welcome.

e.g.
```Rust
// this should not run when the u2i feature is set
#[test]
fn uint64() {
    let obj_min: EncoderResult<Bson> = to_bson(&u64::MIN);
    assert_matches!(obj_min, Err(EncoderError::UnsupportedUnsignedType));
}
```


zonyitoo:
LGTM. cc @kyeah @saghm

kyeah:
Looks great, thank you!

> Open question: does this require further changes in the decoder. Given the tests it seems to work without changes in the decoder.

unsigned integers will be stored as signed ints in the database, so they will be decoded as signed ints using the usual `deserialize_iXX` methods.

> I also could not figure out how to conditionally add a test for the case that the "u2i" is not set. Hints welcome.

You should be able to create default test cases, and ignore it if the feature is enabled:

```
#[test]
#[cfg_attr(feature = "u2i", ignore)]
```

should be good to merge after that 👍

marcelbuesing:
> You should be able to create default test cases, and ignore it if the feature is enabled:
Thanks! Just added the missing tests.

zonyitoo:
@kyeah Merge it if you think it is Ok.

zonyitoo:
@marcelbuesing You can try this in your project, looking for feedbacks.

PR #98, opened 2018-07-10
TimeStamp type uses incorrect endianness?

stefano-pogliani:
Hello,

I have recently attempted to use the new `bson::TimeStamp` type to decode an optime returned by the `replSetGetStatus` MongoDB command.
The value that I got was different from the result of `rs.status()`.

After a bit of investigation I noticed the timestamp is converted to big endian before it is masked and shifted: https://github.com/zonyitoo/bson-rs/blob/master/src/decoder/serde.rs#L595

I am not familiar with the protocol but other clients (including the bson package in the mongodb repo) seem to use little endian encoding for the values of the timestamp:

  * Go Client: https://github.com/mongodb/mongo-go-driver/blob/master/bson/value.go#L820
  * Comment in the official MongoDB codebase: https://github.com/mongodb/mongo/blob/master/src/mongo/bson/timestamp.cpp#L51-L52

Is this indeed a bug or am I doing something wrong?

### DIFF
--- a/src/decoder/serde.rs
+++ b/src/decoder/serde.rs
@@ -592,12 +592,12 @@ impl<'de> Deserialize<'de> for TimeStamp {
 
         match Bson::deserialize(deserializer)? {
             Bson::TimeStamp(ts) => {
-                let ts = ts.to_be();
+                let ts = ts.to_le();
 
                 Ok(TimeStamp { t: ((ts as u64) >> 32) as u32,
                                i: (ts & 0xFFFF_FFFF) as u32, })
             }
-            _ => Err(D::Error::custom("expecting UtcDateTime")),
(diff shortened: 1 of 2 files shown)


saghm:
This looks like a bug to me; the BSON spec indicates that the value of a Timestamp is the "basic type" uint64, which is described as little endian: http://bsonspec.org/spec.html

stefano-pogliani:
Unrelated to this behaviour I also have a commit to implement serde `Serialise` on the new `bson::TimeStamp` that I would like to submit for consideration.

Should that be a separate PR?

zonyitoo:
That must be a bug. Thanks for your contribution.

zonyitoo:
@stefano-pogliani Implement Serialize for bson::TimeStamp separately? How? You should open a PR if you want to.

PR #104, opened 2018-08-10
Fix serialization of newtype structs, simplify serialization of newtype and tuple variants

### LINKED ISSUE #103: Newtype struct and variant serializes to 1-element Array instead of forwarding to the wrapped type
As per serde's conventions, a newtype struct or variant is supposed to be treated as an insignificant wrapper around the inner type. `bson-rs` does not respect this and will serialize newtype structs and variants as 1-element arrays.

H2CO3:
Fixes #103.


zonyitoo:
LGTM. @kyeah Can I merge it? I should bump the minor version to fix this bug.

kyeah:
Thanks, LGTM @zonyitoo

PR #107, opened 2018-12-14
Fix parsing strict mode's `{ "$date": "<date>" }`

dpc:
https://docs.mongodb.com/manual/reference/mongodb-extended-json/#data_date
describes that primary date format is actually this one.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -457,6 +457,11 @@ impl Bson {
                 return Bson::JavaScriptCode(code.to_owned());
             } else if let Ok(hex) = values.get_str("$oid") {
                 return Bson::ObjectId(oid::ObjectId::with_string(hex).unwrap());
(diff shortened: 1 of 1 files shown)


patrickfreed:
The crate now properly parses both forms of extended JSON date as of v1.0.0:

```rust
let relaxed_date = json!({ "$date": "2012-12-24T12:15:30.501Z"});
let canonical_date = json!({ "$date": { "$numberLong": "1356351330501" } });

println!("{:?}", Bson::try_from(relaxed_date).unwrap());
println!("{:?}", Bson::try_from(canonical_date).unwrap());
```
yields:
```
DateTime(2012-12-24T12:15:30.501Z)
DateTime(2012-12-24T12:15:30.501Z)
```
Thank you for the PR nonetheless!

Issue #108, opened 2018-12-14
MongoDB import changes `somekey: 1` to `somekey: NumberLong(1)`


dpc:
I've been using `mongo_driver` that uses `bson` to import some data that I've exported with `mongoexport` (so in Strict mode). First I had problems with date parsing (#107), then I've noticed that normal numbers are needlessly converted to `NumberLong` version.

I might look into it tomorrow, but if anyone has any guesses where to look, it would help a lot.

BTW. I'm either confused, or the code is incorrectly calling "Strict Mode" , "the extended" one. From what I understand the "Extended JSON" is the one that MongoDB uses everywhere with `ObjectID` etc. It's extended because it's nor really correct JSON per-se. The one that converts these fancy type-wrapper to  proper JSON, is "Strict Mode". So `from_extended_document` should probably be called `from_strict_mode_document`.

patrickfreed:
If you check out the [extended JSON specification](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#special-rules-for-parsing-json-numbers), it mentions that the behavior this crate exhibits when deserializing JSON numbers is correct:

>If the number is a non-integer, parsers SHOULD interpret it as BSON Double.

>If the number is an integer, parsers SHOULD interpret it as being of the smallest BSON integer type that can represent the number exactly. If a parser is unable to represent the number exactly as an integer (e.g. a large 64-bit number on a 32-bit platform), it MUST interpret it as a BSON Double even if this results in a loss of precision. The parser MUST NOT interpret it as a BSON String containing a decimal representation of the number.

e.g.
```rust
let int32_json = json!(5);
let int64_json = json!(i32::MAX as i64 + 1);
let double_json = json!(i64::MAX as f64 + 1.0);

println!("{:?}", Bson::try_from(int32_json).unwrap());
println!("{:?}", Bson::try_from(int64_json).unwrap());
println!("{:?}", Bson::try_from(double_json).unwrap());
```
yields:
```
Int32(5)
Int64(2147483648)
Double(9223372036854776000.0)
```

PR #109, opened 2018-12-14
Fix JSON numbers being incorrectly set as integers.

dpc:
Issue #108

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -417,7 +417,7 @@ impl Bson {
             Bson::UtcDatetime(ref v) => {
                 doc! {
                     "$date": {
-                        "$numberLong" => (v.timestamp() * 1000) + v.nanosecond() as i64 / 1000000,
+                        "$numberLong" => ((v.timestamp() * 1000) + v.nanosecond() as i64 / 1000000).to_string(),
                     }
                 }
             }
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -457,6 +457,11 @@ impl Bson {
                 return Bson::JavaScriptCode(code.to_owned());
             } else if let Ok(hex) = values.get_str("$oid") {
                 return Bson::ObjectId(oid::ObjectId::with_string(hex).unwrap());
+            } else if let Ok(Ok(utc)) = values
+                .get_str("$date")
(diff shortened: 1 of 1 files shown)


dpc:
OK. So here I am confused. There are two contradicting tests.

* `test_ser_datetime`
* `oid`

The decision has to be made. Either (de-)serializating  is to Rust native types (`test_ser_date`) or to types where stuff was decomposed to types used in the "Strict Mode Json". I don't think it's possible to have both, without ambiguities. But I'm tired, and I might be confused.

It seems to me, it should be Rust native types, in which case `oid` test is wrong. If the user wants to get the `ObjectId`, they should have used that type, and not a map `{'$oid': "string"}`.

The way I see it, `Bson` can be converted back and forth between Rust types. The "Strict Mode JSON" is just a way to convert BSon to a JSON-compatible encoding, and back. They are separate concepts, for different applications. So the String-Mode encoding functions should never be called from within anything that has to do with `Bson <-> Rust Types` conversions.

dpc:
Oh... strike of enlightenment... So the whole thing is because `serde` has no way to express serialization and deserialization of types that are not native to Rust?

dpc:
It seems I digged myself out of a hole. The Travis-CI fail is nightly only.

I've ported my changed to 0.11.3 (for mongo_driver) and now the imported data is.. `0.0` instead of `0`. So I guess technically correct, but not what I expected. I thought that `0.0` is `0`...

dpc:
I should probably move the two other changes to a separate PR, but I'll wait to see what you think...

patrickfreed:
This works as designed. See my comment on #108. Thank you for submitting the pull request though!

patrickfreed:
I will add that the following is definitely true and is something we plan to improve in the future:
>The way I see it, Bson can be converted back and forth between Rust types. The "Strict Mode JSON" is just a way to encode BSon in a JSON, and back. These are separate concepts, for different use cases. So the Strict-Mode encoding functions should never be called from within anything that has to do with Bson <-> Rust Types conversions. It seems to me they are conflated now, so I can't complete this fix.

See RUST-426 for the ticket to track the work for investigating this.

PR #110, opened 2018-12-14
Fix bunch of issues with encoding to Strict Mode JSON

dpc:


dpc:
This contains #107 and #109

zonyitoo:
Requesting @kyeah for reviewing.

dpc:
@kyeah Thank you for looking into this.

I was just doing some MongoDB export and import. I don't have strong opinions about much, except I need a correct handling of "Strict Mode" BSON encoding to JSON, back and forth. I am exporting from one Db, importing to another, and I compare. So at very least a number `1` from Json must not become Bson's `NumberLong(1)`. Only a `{ $numberLong: 1 }` should become `NumberLong(1)`. It looks like Strict Mode encoding is used in this library in places where it's impossible to represent some Bson types in with serde-supported types, so it makes things more complex (at least for me).


I can't commit to driving this further, as I was barely able to get this far with a lot of guesswork, but I am at least going to test anything you come up with. I'll do what I can. :)

dpc:
The problem with treating Json 1 as integer is: you export series of records. One of them is some number: eg. height of a person. Everyone have height like 172.3, but one in ten people have just 172, and now they become `NumberLong(172)`. I am a bit confused here, because even with my PR, after importing some integers, they display at eg. 0.0, while they originally were just 0.

patrickfreed:
The parsing of extended JSON has been improved and updated to meet the [extended JSON specification](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst) in the 1.0 release, so I'm going to close this PR. Thank you for your contribution nonetheless!

PR #111, opened 2019-01-05
add wasm support

lrlna:
Hello!

I am using this module in a [library](https://github.com/lrlna/mongodb-schema-parser) that will be compiled to WASM. The current version doesn't quite work, as it makes system calls (`get_hostname()` and `libc::getpid()` specifically). 
You can try this yourself with `cargo build --target wasm32-unknown-unknown`.

This PR adds two replacement functions that will only be used in case of WASM, and in all other cases the current implementation will stay. The WASM-compatible functions just generate random numbers for machine id and process id; it's done similarly in the [js version of bson](https://github.com/mongodb/js-bson/blob/master/lib/objectid.js#L8).

Thank you for reviewing o/ !

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
 sudo: false
 
 script:
+    - rustup target add wasm32-unknown-unknown
     - cargo build -v
     - cargo check --target wasm32-unknown-unknown
     - cargo test -v --no-fail-fast
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ sudo: false
 
 script:
     - cargo build -v
+    - cargo check --target wasm32-unknown-unknown
     - cargo test -v --no-fail-fast
     - cargo test -v --no-fail-fast --features u2i
     - cd serde-tests && cargo test -v --no-fail-fast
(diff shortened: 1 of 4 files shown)


zonyitoo:
@kyeah Please review these commits.

saghm:
It's worth noting that MongoDB drivers as a whole are moving towards implementations independent of the hostname and PID, as described [here](It's worth noting that the the MongoDB drivers as a whole are moving towards using random bytes rather than using the hostname (see the spec [here](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#specification)). It might be worth just updating the implementation to follow that spec, which both saves us from needing to use conditional compilation (as the implementation would work for both standard Rust and wasm) and would bring us in line with what other drivers are doing.

lrlna:
Thanks @saghm ! 

If everyone is okay with using just the random bytes and having one compilation, I will gladly change this PR to accommodate that.

kyeah:
yep, let's move to the new spec with random initialized bytes. thanks @saghm + @lrlna, sorry for the delayed response!

lrlna:
@kyeah not delayed at all, don't worry!

let me fix this PR to just have the one compilation that uses random bytes.

lrlna:
closing in favour of #114 !

PR #114, opened 2019-01-14
use random byte array instead of process_id and machine_id

lrlna:
Hello o/

I made a PR last week to compat for WASM compilation, adding a WASM specific method to use random bytes arrays to substitute a call to `libc` and getting `hostname`. In the discussion there, @saghm pointed out MongoDB drivers are now transitioning to use 5 random bytes instead of `process_id` and `machine_id` ([spec](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#id3)). This PR implements that.

In addition to using random bytes, this PR also:
- [x] removes getter method for process_id (spec outlines that random value must not be accessible)
- [x] removes `process_id` and `machine_id` entirely
- [x] adds a check to travis for a WASM build 

This would be probably a semver major in semver terms, but I am not sure how you'd want to handle that (or deprecation for that matter).

Thank you for your time!

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -1,18 +1,21 @@
 //! ObjectId
 
-use libc;
-
 use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
 use std::{error, fmt, io, result};
 
 use byteorder::{BigEndian, ByteOrder, LittleEndian};
-use md5;
 
 use hex::{self, FromHexError};
 
 use rand::{thread_rng, Rng};
 
+#[cfg(not(target_arch = "wasm32"))]
 use hostname::get_hostname;
+#[cfg(not(target_arch = "wasm32"))]
+use libc;
+#[cfg(not(target_arch = "wasm32"))]
+use md5;
+
 use time;
 
 const TIMESTAMP_SIZE: usize = 4;
@@ -197,6 +200,7 @@ impl ObjectId {
 
     // Generates a new machine id represented as an MD5-hashed 3-byte-encoded hostname string.
     // Represented in Little Endian.
+    #[cfg(not(target_arch = "wasm32"))]
     fn gen_machine_id() -> Result<[u8; 3]> {
         // Short-circuit if machine id has already been calculated.
         // Since the generated machine id is not variable, arising race conditions
@@ -230,13 +234,41 @@ impl ObjectId {
         Ok(vec)
     }
 
+    // In case of wasm compilation generates a random 3-byte array.
+    #[cfg(target_arch = "wasm32")]
+    fn gen_machine_id() -> Result<[u8; 3]> {
+        // Short-circuit if machine id has already been calculated.
+        // Since the generated machine id is not variable, arising race conditions
+        // will have the same MACHINE_BYTES result.
+        unsafe {
+            if let Some(bytes) = MACHINE_BYTES.as_ref() {
+                return Ok(bytes.clone());
+            }
+        }
+
+        let mut rng = rand::thread_rng();
+        let vec: [u8; 3] = [rng.gen(), rng.gen(), rng.gen()];
+
+        unsafe { MACHINE_BYTES = Some(vec) };
+        Ok(vec)
+    }
+
     // Gets the process ID and returns it as a 2-byte array.
     // Represented in Little Endian.
+    #[cfg(not(target_arch = "wasm32"))]
     fn gen_process_id() -> [u8; 2] {
         let pid = unsafe { libc::getpid() as u16 };
         let mut buf: [u8; 2] = [0; 2];
         LittleEndian::write_u16(&mut buf, pid);
         buf
+    }
+
+    // In case of wasm return a random 2-byte array.
+    #[cfg(target_arch = "wasm32")]
+    fn gen_process_id() -> [u8; 2] {
+        let mut rng = rand::thread_rng();
+        let vec: [u8; 2] = [rng.gen(), rng.gen()];
+        vec
     }
 
     // Gets an incremental 3-byte count.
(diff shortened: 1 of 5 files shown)


lrlna:
@kyeah i opened this new PR instead of the previous wasm-only one. thanks!

lrlna:
@saghm would you have time to give this a review some time this week?

src/oid.rs
```diff
@@ -193,47 +168,11 @@ impl ObjectId {
...
-    // Represented in Little Endian.
-    fn gen_process_id() -> [u8; 2] {
-        let pid = unsafe { libc::getpid() as u16 };
-        let mut buf: [u8; 2] = [0; 2];
-        LittleEndian::write_u16(&mut buf, pid);
+    // Generate a random 5-byte array.
+    fn gen_process_id() -> [u8; 5] {
+        let rng = thread_rng().gen_range(0, MAX_U24) as u32;
+        let mut buf: [u8; 5] = [0; 5];
+        BigEndian::write_u32(&mut buf, rng);
```
kyeah: Actually @saghm do you know if there are any requirements here as far as Big vs Little endian here? The specifications say timestamp and counter are BigEndian but don't mention the random value, does that imply that this should still be LittleEndian?

saghm: Realistically, I'm not sure it matters (since the bytes are random, I can't see how the order they're inserted would make a difference), but from a quick check, it appears that Python [uses `">"` for the endianness](https://github.com/mongodb/mongo-python-driver/blob/357e3e3d6840f94794fb50ccf68734f483dae7d6/bson/objectid.py#L45), which from Googling appears to be [big endian](https://docs.python.org/2/library/struct.html#byte-order-size-and-alignment]

kyeah: cool yeah, I figured the randomness makes it moot but wanted to check. Thanks!

kyeah:
Two notes:

1. Looks like `cargo check --target wasm32-unknown-unknown` is failing. Not entirely sure why at first glance, but if anyone has leads be my guest!

2. Regarding deprecation:
    > This would be probably a semver major in semver terms, but I am not sure how you'd want to handle that (or deprecation for that matter).
   
    I think the ideal would have been a slow transition where we kept the old machine/process_id format and `#[deprecated]` the machine_id, process_id methods for a month or so before doing a major cut to the new format. /shrug

lrlna:
@kyeah thanks so much for the review! \o/

1. the cargo error comes from the `decimal` crate that does the `Decimal128` implementation. I _completely_ did not test that PR against the wasm target. That crate does a lot of its heavy lifting through `libc` which is very not wasm compatible. Let me do some more looking into this and get back to you!

2. If you'd like to do deprecation, we might have to add a wrapper method so people still can be creating `ObjectId` with `machine_id`/`process_id` (and therefore retrieving it with the two public methods) or the new way of random bytes. Otherwise you can't really retrieve machine_id if the ObjectId wasn't ever created with it. Is that along the lines you were thinking? Or did you have another idea?

kyeah:
@lrlna That's another way to do it! Or we could use random bytes only for WASM, and then make it the default everywhere after a deprecation period. I think it's unlikely that people are using the machine_id and process_id methods though, and we'll eventually need a major semver bump (or minor based on the lax versioning in the crates community), so I'd be fine doing a hard cut for this release. 

I'll let @zonyitoo chime in if there's thoughts on how careful we want to be here w/ the public API.

zonyitoo:
Of course, these changes has to be in a major version bump.

lrlna:
hello! Sorry I am the one who's been quiet now!!

The reason travis wasn't passing is because Decimal128 library upstream is a wrapper around C, and is therefore not compatible with WASM. Although I am currently working on a Decimal128 implementation that doesn't involve using C code, it might take a bit more time. And it would be beneficial for this code to be in master regardless of whether it's WASM compatible or not. Once I fix up my new Decimal128 implementation, I can put this check back into travis, and we should be back in fully operational in WASM land :tada:

saghm:
It looks like the nightly tests are failing due to the `serde-tests` subcrate specifying `serde` as a feature, which I think is a holdover from when we first introduced it in place of rustc-serialize. @kyeah can you confirm that removing the serde feature [here](https://github.com/zonyitoo/bson-rs/blob/b7a441a4470905d3ce8f77e0bc4b8ceca2130b1c/serde-tests/Cargo.toml#L8) is correct? I can open a PR for it assuming it's the right thing to do.

saghm:
Pinging @kyeah again?

kyeah:
@saghm yes I think that's right, thank you for the followup ping!

lrlna:
@saghm let me know if you end up making a PR. I will rebase

saghm:
Yep, I'll make a PR now. Thanks for the response Kevin!

saghm:
Just opened https://github.com/zonyitoo/bson-rs/pull/120; hopefully the CI will pass and then we can merge it

saghm:
Tests are passing on it! @kyeah @zonyitoo any chance we could get that merged to unblock this?

kyeah:
Merged into master, I'll let you merge/rebase @lrlna and then this should be good to go, thanks 🙏

zonyitoo:
Merged.

zonyitoo:
One test is failed in nightly:

```
0.77s$ cd serde-tests && cargo test -v --no-fail-fast
    Updating crates.io index
error: failed to select a version for `bson`.
    ... required by package `serde-tests v0.1.0 (/home/travis/build/zonyitoo/bson-rs/serde-tests)`
versions that meet the requirements `*` are: 0.13.0
```

Hmm..

zonyitoo:
Tell me if you think it is ready to be published to crates.io. @kyeah @saghm

lrlna:
woo! Thanks for merging!

I thought @saghm 's fix should have cleared that error up? That seems to be a similar error that was on master before 🤔

Issue #116, opened 2019-02-25
Malformed bson can force a fatal error


stusmall:
This came up from some fuzz testing of the crate.  I don't know bson's wire protocol but from the output of bsondump it looks like the issue is it declares an illegal BSONSize and the crate still tries to allocate it and crashes.

The offending binary is: [0x06, 0xcc, 0xf9, 0x0a, 0x05, 0x00, 0x00, 0x03, 0x00, 0xff, 0xff]

PR #117, opened 2019-02-25
Fix for #116

stusmall:
This PR adds a simple fuzzer, one failing test case I found with it and a what I think could be a fix.  I don't know BSON's wire protocol well, so my fix should be treated as suspect.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,26 @@
+
+[package]
+name = "bson-fuzz"
+version = "0.0.1"
+authors = ["Automatically generated"]
+publish = false
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies.bson]
+path = ".."
+[dependencies.libfuzzer-sys]
+git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
+
+# Prevent this from interfering with workspaces
+[workspace]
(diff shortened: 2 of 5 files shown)


stusmall:
Don't mind my amends.  :)  I wasn't sure on what the correct upper bound is, but I've got it now.  The spec didn't call it out so I ended up looking into some official mongo repos to see what they use for this case.

zonyitoo:
Why I saw some `fuzz` folders in this commit?

stusmall:
Those are generated by https://github.com/rust-fuzz/cargo-fuzz.  I thought it could be useful to have those tools available to continue to test with but I'd happily remove them.

zonyitoo:
Why you set `MAX_BSON_SIZE: i32 = 16 * 1024 * 1024`?

stusmall:
I originally pulled it from the source for bsondump from mongo, but they have this restraint in a lot of their code.  
https://github.com/mongodb/mongo-tools/blob/7722e31702b8d5f41520487a5197dd6202f75534/common/lldb/db.go#L38
https://github.com/mongodb/mongo-tools/blob/217728df376760cc66c93268fb50b7eca399572c/common/db/db.go#L38

This limit is what keeps bsondump from choking on this input.

It is also mentioned in the docs:  
https://docs.mongodb.com/manual/reference/limits/#BSON-Document-Size

This is a restraint set out by mongodb and not in the bson spec but AFAIK bson was made for mongo and mongo is its only user.

stusmall:
Let me know if you want me to remove the fuzz code.  I'm happy to.

zonyitoo:
Nope. Just leave it there.

PR #125, opened 2019-08-28
WIP: Switch to incomplete decimal128 crate

spease:
This is an ad hoc stuff to get some stuff to compile. Not sure if I'll be able to spend more time on it, but I figured I'd make you guys aware of it and see if people had any comments / alternative suggestions for enabling wasm with bson-rs.

PR #132, opened 2019-10-15
Fix building on ARM arch

Fenex:
Hello.

`bson-rs` has a depending crate `decimal` which may using with optional feature "decimal128". Building `decimal` (last version 2.0.4) is terminating with error on ARM arch when I run
```bash
cargo build --features="decimal128"
```
The reason is the difference representation `c_char` type by two different platforms: `c_char` is `i8` on `x86`\\`amd64` and is `u8` on `arm`.

I guess you want ask me why I write here and don't create an issue inside 'decimal' repository. The fact is `decimal` crate already has a merged PR alkis/decimal#42 for fix that, however fixed version hasn't released. The trouble is, @alkis, author of the crate, has no active since April and I have no idea how much time we will not be able to get a fixed version on crates.io.

So, I suggest change source of the depending crate from crates.io to git commit with fixed compile error. It is the fastest and more safety path to resolve the problem.

On the other hand it is possible to migrate on another crate, for example, [rust_decimal](https://crates.io/crates/rust-decimal) or other. It will be more complex and require some code changes for apply new library API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ time = "0.1"
 linked-hash-map = "0.5"
 hex = "0.3"
 md5 = "0.6"
-decimal = { version = "2.0.4", default_features = false, optional = true }
+decimal = { git = "https://github.com/alkis/decimal.git", rev = "5577a60b9b8860a322288856ff2419c191ed080f", optional = true }
 
 [dev-dependencies]
 assert_matches = "1.2"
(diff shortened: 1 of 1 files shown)


zonyitoo:
#125 I am hoping that in the future, we can switch to `decimal128` crate.

zonyitoo:
Using git path inside Cargo.toml will make this crate unpublishable to crates.io.

saghm:
Unfortunately we can't merge this due to not being able to publish to crates.io with a raw git dependency, and the decimal crate has still not been updated, so closing this out. You can follow https://jira.mongodb.org/browse/RUST-36, which is the ticket we'll use to track work on the decimal BSON type.

FollieHiyuki:
The `decimal` crate got updated to 2.1.0 recently.
I think we can try again

PR #135, opened 2019-12-20
 Replace 'linked-hash-map' with 'indexmap'

danclive:


src/doc.rs
```diff
@@ -86,100 +94,80 @@ impl Display for OrderedDocument {
     }
 }
 
-impl Debug for OrderedDocument {
+impl Debug for Document {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "OrderedDocument({:?})", self.inner)
```
nevi-me: shouldn't this be `write!(f, "Document({:?})", self.inner)`?

danclive: ok! thanks!

PR #137, opened 2020-01-05
Cleanup deps

AdminXVII:
- Remove unused dependencies (libc, md5)
- Replace the time crate with std equivalent
- Replace the byteorder crate with the std's `{to,from}_{le,be}_bytes` functions on integers
- Some formatting

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,15 +25,12 @@ decimal128 = ["decimal"]
 name = "bson"
 
 [dependencies]
-byteorder = "1"
 chrono = "0.4"
-libc = "0.2"
 rand = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order"] }
 linked-hash-map = "0.5.3"
 hex = "0.4.2"
-md5 = "0.7.0"
 decimal = { version = "2.0.4", default_features = false, optional = true }
 base64 = "0.12.1"
 lazy_static = "1.4.0"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -24,16 +24,12 @@ decimal128 = ["decimal"]
...
 [dependencies]
-byteorder = "1"
 chrono = "0.4"
-libc = "0.2"
 rand = "0.7"
 serde = "1.0"
 serde_json = { version = "1.0", features = ["preserve_order"] }
-time = "0.1"
 linked-hash-map = "0.5"
 hex = "0.3"
```
nevi-me: Hey @AdminXVII would you be able to update `hex` to 0.4, in case your PR gets merged? I wanted to update it, but then saw your PR

AdminXVII: Yeah I think I could update the PR during the week. I'll get back to you soon.

AdminXVII: It was easier than I thought, it's already rebased. Tell me if there is anything you want me to do.

saghm:
@AdminXVII due to the large number of code changes since this PR was filed, I went ahead and did the rebase myself rather than put you through it. Would you be able to take a look at the code as it is now to verify that I properly preserved all of the changes you made?

AdminXVII:
Yeah I think that does the same job.

saghm:
Awesome, thanks!

The lint failure should be addressed by #200, so I've added the rest of the team to the review so we can get this merged!

PR #139, opened 2020-01-11
Remove dependency on time crate

jhpratt:


jhpratt:
@saghm Looks like you're the one that's been doing work on this repo recently. Any chance this will be merged? Conflicts have arisen in the past 3+ months, though it looks like that's just imports.

saghm:
Hello! Apologies for the delay in responding. You're correct that I'm one of the maintainers of this project now; it's now maintained by the same team that makes the MongoDB Rust driver.

This seems like a reasonable change to make to me. If you can rebase to fix the conflicts and push up the change, I'll enable the CI run for it, and we can get it ready to be merged.

jhpratt:
@saghm Just rebased and force-pushed. Looks like CI was included as part of that.

saghm:
Looks like the linter caught a section that wasn't formatted properly. Could you run rustfmt on your changes?

jhpratt:
Just pushed up a commit that ran rustfmt. IIRC it's because I tried to match the format from before, which changed over the course of this PR being open.

saghm:
That makes sense; when we started maintaining the project, we switched the rustfmt config to match the one we use on the MongoDB Driver. Sorry about that!

src/oid.rs
```diff
@@ -134,8 +136,12 @@ impl ObjectId {
...
     // Represented in Big Endian.
     fn gen_timestamp() -> [u8; 4] {
-        let timespec = time::get_time();
-        let timestamp = timespec.sec as u32;
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("system clock is before 1970")
+            .as_secs()
+            .try_into()
+            .expect("FIXME before 2038");
```
patrickfreed, line 144: since this buffer is actually an unsigned 32 bit field, this should succeed until aboout Jan 6th, 2106 according to the [ObjectId spec](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#timestamp-field). I think we can update this "FIXME" to say something like "will succeed until 2106 due to unsigned" or something similar to reflect the low risk nature of the unwrap here.

patrickfreed:
Thanks for your contribution!

PR #144, opened 2020-02-07
Fix failing evergreen lint

saghm:
I've updated the BSON library to use the same rustfmt config as the driver and reformatted it, as well as fixed all of the clippy lints that don't require breaking changes and whitelisted the rest.

### DIFF
--- a/src/decimal128.rs
+++ b/src/decimal128.rs
@@ -1,7 +1,6 @@
 //! [BSON Decimal128](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst) data type representation
 
-use std::fmt;
-use std::str::FromStr;
+use std::{fmt, str::FromStr};
 
 use decimal::d128;
 
@@ -26,7 +25,9 @@ impl Decimal128 {
     /// let dec128 = Decimal128::from_str("1.05E+3");
     /// ```
     pub fn from_str(s: &str) -> Decimal128 {
-        Decimal128 { inner: s.parse::<d128>().expect("Invalid Decimal128 string"), }
+        Decimal128 {
+            inner: s.parse::<d128>().expect("Invalid Decimal128 string"),
+        }
     }
 
     /// Construct a `Decimal128` from a `i32` number.
@@ -38,7 +39,9 @@ impl Decimal128 {
     /// let dec128 = Decimal128::from_i32(num);
     /// ```
     pub fn from_i32(d: i32) -> Decimal128 {
-        Decimal128 { inner: From::from(d) }
+        Decimal128 {
+            inner: From::from(d),
+        }
     }
 
     /// Construct a `Decimal128` from a `u32` number.
@@ -50,7 +53,9 @@ impl Decimal128 {
     /// let dec128 = Decimal128::from_u32(num);
     /// ```
     pub fn from_u32(d: u32) -> Decimal128 {
-        Decimal128 { inner: From::from(d) }
+        Decimal128 {
+            inner: From::from(d),
+        }
     }
 
     /// Construct a `Decimal128` from a `i32` number.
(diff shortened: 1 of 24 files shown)


src/decimal128.rs
```diff
@@ -68,10 +69,25 @@ impl Decimal128 {
     /// let int = dec128.into_i32();
     /// assert_eq!(int, num);
     /// ```
+    #[deprecated(since = "0.15.0", note = "Replaced by `to_i32`")]
```
saghm: Clippy correctly noticed that this shouldn't be named `into_i32` due to it not taking ownership of `self`. I've deprecated this and `into_u32` and added `to_{u,i}32` to replace them.

src/encoder/serde.rs
```diff
@@ -76,6 +76,7 @@ pub struct Encoder;
 
 impl Encoder {
     /// Construct a new `Serializer`.
+    #[allow(clippy::new_without_default)]
```
saghm: In general, implementing `new` with no parameters means that you can also implement or derive `Default`. For now, I've just whitelisted each occurrence of this pattern.

src/oid.rs
```diff
@@ -101,15 +101,12 @@ impl ObjectId {
...
-        for i in 0..TIMESTAMP_SIZE {
-            buf[TIMESTAMP_OFFSET + i] = timestamp[i];
-        }
-        for i in 0..PROCESS_ID_SIZE {
-            buf[PROCESS_ID_OFFSET + i] = process_id[i];
-        }
-        for i in 0..COUNTER_SIZE {
-            buf[COUNTER_OFFSET + i] = counter[i];
-        }
+        buf[TIMESTAMP_OFFSET..(TIMESTAMP_SIZE + TIMESTAMP_OFFSET)]
```
saghm: Apparently clippy is able to detect [manual copies of bytes](https://rust-lang.github.io/rust-clippy/master/index.html#manual_memcpy) and suggests to replace them with a method call, which is awesome

src/ordered.rs
```diff
@@ -108,28 +108,30 @@ pub struct OrderedDocumentIterator<'a> {
     inner: linked_hash_map::Iter<'a, String, Bson>,
 }
 
+type DocumentMap<'a, T> = Map<OrderedDocumentIterator<'a>, fn((&'a String, &'a Bson)) -> T>;
```
saghm: Clippy doesn't like the really long type name in the two structs below (which is fair), so I abstracted it into a type alias.

tests/lib.rs
```diff
@@ -1 +1,3 @@
+#![allow(clippy::cognitive_complexity)]
```
saghm: Clippy doesn't like some of our longer test functions, so I just blanket whitelisted all of them in the tests

tests/modules/ser.rs
```diff
@@ -5,6 +5,7 @@ use bson::{from_bson, oid::ObjectId, to_bson, Bson, EncoderError, EncoderResult}
 use std::{collections::BTreeMap, u16, u32, u64, u8};
 
 #[test]
+#[allow(clippy::float_cmp)]
```
saghm: Clippy complains when we directly compare floats and suggests checking if they're within `std::f{32,64}::EPSILON` of each other. Since we're not doing any arithmetic with the floats, this shouldn't be an issue, so I've whitelisted each function that does this.

tests/modules/ser.rs
```diff
@@ -27,7 +28,7 @@ fn string() {
 #[test]
 fn arr() {
     let obj = Bson::Array(vec![Bson::I32(0), Bson::I32(1), Bson::I32(2), Bson::I32(3)]);
-    let arr: Vec<i32> = from_bson(obj.clone().clone()).unwrap();
+    let arr: Vec<i32> = from_bson(obj.clone()).unwrap();
```
saghm: I'm not actually sure why there were multiple calls to `Clone` here.

PR #147, opened 2020-03-12
RUST-305 Fix deserialization of non-generic binary

saghm:
[RUST-305](https://jira.mongodb.org/browse/RUST-305)

As discussed in person on Monday, this fixes a bug where the library deserializes all binary subtypes into subtype 0 (generic) regardless of the original subtype that was being deserialized. This occurred due to the fact that the deserialize implementation for the bson Decoder called `visit_bytes` on the visitor; while it allowed `Bson::Binary` to be deserialized into `Vec<u8>`, it also caused the subtype to be ignored in the case that `Bson` was getting deserialized back into `Bson` (which occurs most often in cases where a struct that derives `Deserialize` contains a field of type `Document` or `Bson`). This fixes that bug by calling `visit_bytes` in the case that the subtype of the `Binary` being deserialized is generic, and instead piggyback off of the extended JSON implementation to deserialize other binary subtypes. The downside of this solution is that it removes the functionality to deserialize non-generic subtypes into `Vec<u8>`, but I think this is worth it because a) deserializing non-generic binary without losing the subtype is more important than the non-essential quality of life improvement to deserialize into `Vec<u8>` and b) users can still implement a custom deserializer on a field of type `Vec<u8>` if they still want this functionality.

### DIFF
--- a/src/decoder/serde.rs
+++ b/src/decoder/serde.rs
@@ -312,10 +312,11 @@ impl<'de> Deserializer<'de> for Decoder {
             Bson::I32(v) => visitor.visit_i32(v),
             Bson::I64(v) => visitor.visit_i64(v),
             Bson::Binary(BinarySubtype::Generic, v) => visitor.visit_bytes(&v),
-            binary @ Bson::Binary(..) => visitor.visit_map(MapDecoder { iter: binary.to_extended_document()
-                                                                                    .into_iter(),
-                                                                        value: None,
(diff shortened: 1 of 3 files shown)


saghm:
The clippy failures are unrelated, so I'll submit a separate PR like I did for the driver

saghm:
Good call on the tests; it turns out that the encoder has the same issue, which I've fixed in the same way.

PR #148, opened 2020-03-12
minor: fix new clippy lint failures with rust 1.42.0

saghm:


src/decoder/error.rs
```diff
@@ -81,8 +81,16 @@ impl fmt::Display for DecoderError {
 impl error::Error for DecoderError {
     fn description(&self) -> &str {
         match *self {
-            DecoderError::IoError(ref inner) => inner.description(),
-            DecoderError::FromUtf8Error(ref inner) => inner.description(),
+            DecoderError::IoError(ref inner) =>
+            {
+                #[allow(deprecated)]
```
saghm: `Error::description` has been deprecated in Rust 1.42, and clippy suggests using `to_string()`, but unfortunately we aren't able to here since we're returning a reference, and the newly-allocated string returned by `to_string()` won't live long enough. We probably will want to get rid of our own `description` implementations before 1.0, but for now, I've just annotated each of these calls to suppress the deprecation warning.

patrickfreed: Can you file a ticket for dropping the description implementations?

saghm: Filed https://jira.mongodb.org/browse/RUST-312

PR #153, opened 2020-04-03
RUST-315 Remove ObjectId::with_timestamp

saghm:


tests/modules/oid.rs
```diff
@@ -14,24 +14,6 @@ fn deserialize() {
     assert_eq!(1_122_867 as u32, oid.counter());
 }
 
-#[test]
```
saghm, line 17: This test is used to verify `ObjectId::timestamp`, which will be removed as part of RUST-316

saghm:
To provide some context for why we're doing this: relying underlying format of an ObjectId is a bit of an antipattern, as the specification for how ObjectId are generated can change (and has done so in the past), and this is not considered a breaking change for the BSON library, the MongoDB driver, or the database itself. In particular, it's brittle to rely on the timestamp field due to the fact that it relies on the system time of the machine that generates the ObjectId; the time could be different across various machines due to time zone, differing NTP servers being synced to, or even just the machine being purposely configured to have an "incorrect" time. We standardize how ObjectIds are generated to ensure that they'll be unique across different machines and different drivers, but there are no user-facing guarantees about how they'll be generated. Users who need to perform queries over documents created in a certain time range should add a field with the time of insertion (and possibly update that on changes, if desired) and use that field when querying. See the documentation for [`$currentDate`](https://docs.mongodb.com/manual/reference/operator/update/currentDate/) for some details on how this can be done.

PR #154, opened 2020-04-03
RUST-316 Remove timestamp/counter getters on ObjectId

saghm:


tests/modules/oid.rs
```diff
@@ -1,27 +1,6 @@
...
-        0xEFu8, 0xCDu8, 0xABu8, 0xFAu8, 0x29u8, 0x11u8, 0x22u8,
-        0x33u8, // increment is 1122867
-    ];
-
-    let oid = ObjectId::with_bytes(bytes);
-    assert_eq!(3_735_928_559 as u32, oid.timestamp());
-    assert_eq!(1_122_867 as u32, oid.counter());
-}
-
-#[test]
```
saghm, line 17: This test is only used to verify `ObjectId::timestamp` with relation to `ObjectId::with_timestamp`, which is being removed as part of RUST-315

PR #156, opened 2020-04-03
RUST-318 Remove Document::insert_bson

saghm:


saghm:
Oops, didn't realize that rustfmt did a number on this file. I think I missed this when initially formatting the project because it's technically a separate crate from the rest of the library. I've added a minor commit that formats and fixes the clippy lints in serde-tests as well as adds it to the evergreen scripts so it will get verified in CI from now on.

PR #159, opened 2020-04-03
RUST-319 Convert variants of the BSON enum to use structs instead of multiple fields

saghm:


saghm:
I split the changes for each BSON type into separate commits to aid reviewability

src/bson.rs
```diff
@@ -53,13 +53,8 @@ pub enum Bson {
     Boolean(bool),
     /// Null value
     Null,
-    /// Regular expression - The first cstring is the regex pattern, the second is the regex
```
saghm, line 56: I moved the detailed documentation below to the struct itself

src/bson.rs
```diff
@@ -838,3 +845,19 @@ impl From<DateTime<Utc>> for UtcDateTime {
         UtcDateTime(x)
     }
 }
+
+/// Represents a BSON regular expression.
+#[derive(Debug, Clone, PartialEq)]
+pub struct RegExp {
```
saghm, line 867: I named this `RegExp` to match the existing name of the BSON variant. That being said, this would be a good opportunity to change it if we prefer something like `Regex`.

src/decoder/serde.rs
```diff
@@ -689,6 +689,18 @@ impl<'de> Deserialize<'de> for TimeStamp {
     }
 }
 
+impl<'de> Deserialize<'de> for RegExp {
```
saghm, line 698: This follows the pattern of the other deserializers defined here. Since the deserialize implementation for `Bson` implements deserializing from extjson, we piggyback off that to either get a RegExp or return an error.

src/encoder/serde.rs
```diff
@@ -496,6 +496,17 @@ impl Serialize for TimeStamp {
     }
 }
 
+impl Serialize for RegExp {
```
saghm, line 514: This follows the pattern of the other serializers defined here. Since the serialize implementation for Bson implements serializing into extjson, we piggyback off that to serialize a RegExp.

tests/serde.rs
```diff
@@ -94,6 +94,56 @@ fn test_de_timestamp() {
     assert_eq!(foo.ts, TimeStamp { t: 12, i: 10 });
 }
 
+#[test]
```
saghm, line 97: These follow the pattern of the TimeStamp tests above.

src/bson.rs
```diff
@@ -861,3 +871,9 @@ pub struct RegExp {
     /// unicode.
     pub options: String,
 }
+
+#[derive(Debug, Clone, PartialEq)]
```
saghm, line 881: This name is pretty verbose, but I don't mind that as a subtle way to discourage using this, since it's not really something that is needed

src/decoder/mod.rs
```diff
@@ -186,13 +186,16 @@ fn decode_array<R: Read + ?Sized>(reader: &mut R, utf8_lossy: bool) -> DecoderRe
 }
 
 fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8, utf8_lossy: bool) -> DecoderResult<Bson> {
-    use spec::ElementType::*;
```
saghm, line 189: The import of `ElementType::JavaScriptCodeWithScope` conflicted with the struct, so I changed this to only import `ElementType` itself.

src/decoder/mod.rs
```diff
@@ -205,34 +208,39 @@ fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8, utf8_lossy: bool) -> D
...
+        }
+        Some(ElementType::JavaScriptCodeWithScope) => {
             // disregard the length:
             //     using Read::take causes infinite type recursion
             read_i32(reader)?;
 
             let code = read_string(reader, utf8_lossy)?;
             let scope = decode_document(reader)?;
-            Ok(Bson::JavaScriptCodeWithScope(code, scope))
+            Ok(Bson::JavaScriptCodeWithScope(JavaScriptCodeWithScope {
```
saghm, line 235: This is the only change to method besides adding the `ElementType::` prefix to the cases

src/decoder/serde.rs
```diff
@@ -701,6 +701,18 @@ impl<'de> Deserialize<'de> for RegExp {
     }
 }
 
+impl<'de> Deserialize<'de> for JavaScriptCodeWithScope {
```
saghm, line 710: This follows the pattern of the other deserializers defined here. Since the deserialize implementation for Bson implements deserializing from extjson, we piggyback off that to either get a JavaScriptCodeWithScope or return an error.

src/encoder/serde.rs
```diff
@@ -507,6 +507,17 @@ impl Serialize for RegExp {
     }
 }
 
+impl Serialize for JavaScriptCodeWithScope {
```
saghm, line 525: This follows the pattern of the other serializers defined here. Since the serialize implementation for Bson implements serializing from extjson, we piggyback off that to serialize JavaScriptCodeWithScope.

tests/serde.rs
```diff
@@ -144,6 +144,56 @@ fn test_de_regex() {
     assert_eq!(foo.regex, regex);
 }
 
+#[test]
```
saghm, line 147: These follow the pattern of the TimeStamp test above.

src/bson.rs
```diff
@@ -211,9 +209,9 @@ impl From<bool> for Bson {
     }
 }
 
-impl From<(String, String)> for Bson {
```
saghm, line 214: I'm not thrilled with the previous implementation of converting from `(String, String)` for regex, `(String, Document)` for CodeWithScope, etc., as there might eventually be a new BSON type that's isomorphic to an existing one, so I replaced each of these with the conversions from the new structs.

src/decoder/serde.rs
```diff
@@ -713,6 +719,18 @@ impl<'de> Deserialize<'de> for JavaScriptCodeWithScope {
     }
 }
 
+impl<'de> Deserialize<'de> for Binary {
```
saghm, line 722: This follows the pattern of the other deserializers defined here. Since the deserialize implementation for Bson implements deserializing from extjson, we piggyback off that to either get a Binary or return an error.

src/encoder/serde.rs
```diff
@@ -518,6 +533,17 @@ impl Serialize for JavaScriptCodeWithScope {
     }
 }
 
+impl Serialize for Binary {
```
saghm, line 536: This follows the pattern of the other serializers defined here. Since the serialize implementation for Bson implements serializing from extjson, we piggyback off that to serialize Binary.

tests/serde.rs
```diff
@@ -269,13 +272,64 @@ fn test_binary_non_generic_roundtrip() {
...
-        &doc! {"data": Bson::Binary(BinarySubtype::BinaryOld, b"12345abcde".to_vec())}
+        &doc! {"data": Bson::Binary(Binary { subtype: BinarySubtype::BinaryOld, bytes: b"12345abcde".to_vec() })}
+    );
+
+    let f = bson::from_bson::<Foo>(b).unwrap();
+    assert_eq!(x, f);
+}
+
+#[test]
+fn test_binary_helper_generic_roundtrip() {
```
saghm, line 292: These follow the pattern of the `Bson::Binary` tests above.

src/bson.rs
patrickfreed: I would say I prefer `Regex` since it mirrors the `regex` crate's definition, which I imagine we would want to consider interop with in the future.

src/encoder/serde.rs
```diff
@@ -496,6 +496,17 @@ impl Serialize for TimeStamp {
     }
 }
 
+impl Serialize for RegExp {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let doc = Bson::RegExp(self.clone());
```
patrickfreed, line 520: I know this is used elsewhere in this file, but this variable probably shouldn't be named `doc`.

src/bson.rs
```diff
@@ -838,3 +845,19 @@ impl From<DateTime<Utc>> for UtcDateTime {
...
+    pub pattern: String,
+
+    /// The options for the regex.
+    ///
+    /// Options are identified by characters, which must be stored in
+    /// alphabetical order. Valid options are 'i' for case insensitive matching, 'm' for
+    /// multiline matching, 'x' for verbose mode, 'l' to make \w, \W, etc. locale dependent,
+    /// 's' for dotall mode ('.' matches everything), and 'u' to make \w, \W, etc. match
+    /// unicode.
+    pub options: String,
```
patrickfreed, line 878: would we ever want to consider introducing an options type here to enforce the alphabeticalness or so users wouldn't have to remember the various flags?

src/bson.rs
```diff
@@ -111,9 +111,12 @@ impl Debug for Bson {
 
                 write!(f, "TimeStamp({}, {})", time, inc)
             }
-            Bson::Binary(t, ref vec) => {
-                write!(f, "BinData({}, 0x{})", u8::from(t), hex::encode(vec))
-            }
+            Bson::Binary(Binary { subtype, ref bytes }) => write!(
+                f,
+                "BinData({}, 0x{})",
```
patrickfreed, line 116: Probably makes sense to just write `Binary` here instead. I think `BinData` is just the shell's output for this type. Also, if this is updated to be like all the others (e.g. `CaseName({:?})`), then we can just do away with this `Debug` implementation altogether.

src/bson.rs
```diff
@@ -361,11 +367,11 @@ impl From<Bson> for Value {
                     "i": inc
                 })
             }
-            Bson::Binary(t, ref v) => {
-                let tval: u8 = From::from(t);
+            Bson::Binary(Binary { subtype, ref bytes }) => {
+                let tval: u8 = From::from(subtype);
                 json!({
```
patrickfreed, line 372: This is also the legacy extJSON. This should be `{ "$binary": {"base64": <base64-encoded (with padding as =) payload as a string>, "subType": <BSON binary type as a one- or two-character hex string>}}`

src/bson.rs
```diff
@@ -53,25 +53,20 @@ pub enum Bson {
...
     JavaScriptCode(String),
     /// JavaScript code w/ scope
-    JavaScriptCodeWithScope(String, Document),
+    JavaScriptCodeWithScope(JavaScriptCodeWithScope),
     /// 32-bit integer
     I32(i32),
     /// 64-bit integer
     I64(i64),
     /// Timestamp
     TimeStamp(i64),
```
patrickfreed, line 67: I think it would be far less error-prone and more ergonomic to introduce a type for timestamp as well rather than bit shifting an integer.

saghm: Yeah, I think this is preferable too. The original implementation probably relied a little too heavily on mirroring the binary structure exactly. I'll add a follow-up commit for this.

src/bson.rs
saghm: I'll add a follow-up commit for this after I update the timestamp stuff as well (since that's currently being formatted as two integers like the new structure will be.

src/bson.rs
```diff
@@ -434,10 +432,13 @@ impl Bson {
     #[doc(hidden)]
     pub fn to_extended_document(&self) -> Document {
         match *self {
-            Bson::RegExp(ref pat, ref opt) => {
+            Bson::RegExp(RegExp {
+                ref pattern,
+                ref options,
+            }) => {
                 doc! {
```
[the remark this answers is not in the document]
saghm: When this implementation was originally written, legacy extended JSON was the only extended JSON :)

The BSON corpus tests have never been implemented for this library precisely because it's never been updated to the new extended JSON spec. If you're asking why we haven't implemented it since we took over stewardship of this library, then it's just due to having higher priority things to work on first.

src/bson.rs
saghm: I'm not convinced that we should be doing interop with the `regex` crate (mostly due to having misgivings about regex being a thing that should even exist at the BSON type level), but I also prefer the name `Regex`, so I'll update it.

src/bson.rs
saghm: I prefer leaving it like this; I don't think the BSON regex type is going to get used enough to warrant the extra work. If anything, I'm leaning towards just linking to the MongoDB documentation for details about the options rather than putting them here; I honestly don't know if they've changed at all since this comment was first written.

src/encoder/serde.rs
saghm: Yeah, I agree; I should have paid more attention to that. I'll update it in all of these implementations.

src/bson.rs
```diff
@@ -801,23 +770,29 @@ impl Bson {
...
-/// use bson::TimeStamp;
-///
-/// #[derive(Serialize, Deserialize)]
-/// struct Foo {
-///     timestamp: TimeStamp,
-/// }
-/// ```
+/// Represents a BSON timestamp value.
 #[derive(Debug, Eq, PartialEq, Clone, Copy, Hash)]
 pub struct TimeStamp {
```
saghm, line 775: This struct already existed, so I just reused it and changed the field names to be more descriptive.

src/bson.rs
```diff
@@ -801,23 +770,29 @@ impl Bson {
...
-    pub t: u32,
-    pub i: u32,
+    pub time: u32,
+    pub increment: u32,
+}
+
+impl TimeStamp {
+    pub(crate) fn to_le_i64(self) -> i64 {
+        let upper = (self.time.to_le() as u64) << 32;
```
saghm, line 782: This arithmetic is moved from [here](https://github.com/mongodb/bson-rust/blob/9055e1833f872eed6e5f5e116512e6d401787964/src/encoder/serde.rs#L493)

src/bson.rs
```diff
@@ -801,23 +770,29 @@ impl Bson {
...
+        let lower = self.increment.to_le() as u64;
+
+        (upper | lower) as i64
+    }
+
+    pub(crate) fn from_le_i64(val: i64) -> Self {
+        let ts = val.to_le();
+
+        TimeStamp {
+            time: ((ts as u64) >> 32) as u32,
```
saghm, line 792: This arithmetic is moved from [here](https://github.com/mongodb/bson-rust/blob/9055e1833f872eed6e5f5e116512e6d401787964/src/decoder/serde.rs#L682-L685)

tests/modules/macros.rs
```diff
@@ -28,7 +28,7 @@ fn standard_format() {
         "code": Bson::JavaScriptCode("function(x) { return x._id; }".to_owned()),
         "i32": 12,
         "i64": -55,
-        "timestamp": Bson::TimeStamp(229_999_444),
+        "timestamp": Bson::TimeStamp(TimeStamp { time: 0, increment: 229_999_444 }),
```
saghm, line 31: `229_999_444` is below `u32::MAX`, so this is the correct way to specify the old value

saghm:
I've updated this with the switch to using a struct for `Bson::TimeStamp`

src/bson.rs
```diff
@@ -556,12 +517,20 @@ impl Bson {
...
+                if time >= 0 && increment >= 0 {
+                    return Bson::TimeStamp(TimeStamp {
+                        time: time as u32,
+                        increment: increment as u32,
+                    });
+                }
+            } else if let (Ok(time), Ok(increment)) = (values.get_i64("t"), values.get_i64("i")) {
+                if time >= 0 && increment >= 0 {
+                    return Bson::TimeStamp(TimeStamp {
+                        time: time as u32,
```
patrickfreed, line 530: will this panic if either are too big?

src/bson.rs
```diff
@@ -739,23 +708,29 @@ impl Bson {
...
-/// struct Foo {
-///     timestamp: TimeStamp,
-/// }
-/// ```
+/// Represents a BSON timestamp value.
 #[derive(Debug, Eq, PartialEq, Clone, Copy, Hash)]
 pub struct TimeStamp {
-    pub t: u32,
-    pub i: u32,
+    pub time: u32,
```
patrickfreed, line 714: these two fields could use docstrings.

src/bson.rs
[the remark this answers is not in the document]
patrickfreed: Seems like something we will want to fix for 1.0, but not in this PR. Filed [RUST-346](https://jira.mongodb.org/browse/RUST-346).

src/bson.rs
saghm: Hmm, I thought we already had an auto-generated ticket for this, but I can't find it now.

src/bson.rs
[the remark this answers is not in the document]
saghm: Hmm, I thought we already had an auto-generated ticket for this, but I can't find it now.

src/bson.rs
saghm: Done

src/bson.rs
patrickfreed: the MAX values are inclusive right? So we probably need `<=`.

src/bson.rs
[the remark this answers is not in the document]
patrickfreed: Yeah doesn't look like we had one. The only things coming up for "corpus" and "extended json" are the ticket I filed and decimal 128 support.

src/bson.rs
saghm: Updated

PR #161, opened 2020-04-23
Added support for relaxed format of DateTime 

### LINKED ISSUE #160: Support for Relaxed Format of DateTime while creating Bson from Extended JSON
Extended Json specification allows to specify DateTime in format ISO-8601 format
https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson.Date

But currently the "relaxed" format doesn't seems to be supported in this crate.

    use bson::Bson;
    use serde_json::Value;
    use std::str::FromStr;

    fn main() {

        let x_json_1: Value = Value::from_str(r#"{
            "updatedTime": {"$date": "2020-04-01T00:00:00Z"}
        }"#).unwrap();

        let x_json_2: Value = Value::from_str(r#"{
            "updatedTime": {"$date": {"$numberLong": 1585699200000}}
        }"#).unwrap();

        let data1 = Bson::from(x_json_1);
        let data2 = Bson::from(x_json_2);

        // This fails with error.
        //
        // thread 'main' panicked at 'assertion failed: `(left == right)`
        //   left: `Document(OrderedDocument({"updatedTime": Document(OrderedDocument({"$date": String("2020-04-01T00:00:00Z")}))}))`,
        //  right: `Document(OrderedDocument({"updatedTime": UtcDatetime(2020-04-01T00:00:00Z)}))`', src/main.rs:21:5
        assert_eq!(data1, data2);
    }

VigneshChennai:
Added support for relaxed format of DateTime Extended JSON.
Fixes #160.

**Caveat**
Uses https://docs.rs/chrono/0.4.6/chrono/struct.DateTime.html#method.parse_from_rfc3339 to parse the date string which is stricter than ISO 8601 mention in https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson.Date specification

**Update:** The Mongodb Extended json specification also uses the stricter format as they were referring the RFC3339 only.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -559,6 +559,12 @@ impl Bson {
                 return Bson::UtcDatetime(
                     Utc.timestamp(long / 1000, ((long % 1000) * 1_000_000) as u32),
                 );
+            } else if let Ok(string) = values
+                .get_str("$date")
+            {
+                if let Ok(date) = chrono::DateTime::parse_from_rfc3339(string) {
+                    return Bson::UtcDatetime(date.into())
+                }
             } else if let Ok(sym) = values.get_str("$symbol") {
                 return Bson::Symbol(sym.to_owned());
             }


patrickfreed:
Thanks for the pull request! This functionality has actually been implemented separately, see my comment on #160.

PR #164, opened 2020-05-08
RUST-345 Improve naming of Bson and ElementType cases

patrickfreed:
[RUST-345](https://jira.mongodb.org/browse/RUST-345)

This PR renames `Bson` and `ElementType` enum cases to be more consistent with each other, with other MongoDB drivers, and with the Rust ecosystem as a whole, in that order of priority.

The names were chosen according to the following guidelines: What other drivers have settled on calling the various types defines the name of the "BSON type". From there, if there is a Rust type (including from popular community libraries) that shares a name with that type, name the case according to that. 

For instance, all drivers refer to what we used to call `Bson::FloatingPoint` as a "Double", which de-facto makes the name of the "BSON type" "Double". Since there is no Rust "double" type, we simply name the enum case `Bson::Double`. For regular expression, there is a popular `regex` crate which names the type `Regex`, so we name the case `Bson::Regex`. The same goes for `Bson::I32` and `Bson::I64`.

Changes to `Bson`:
- `FloatingPoint` -> `Double`
- `Utf8String` -> `String`
- `Boolean` -> `Bool`
- `TimeStamp` -> `Timestamp` (timestamp is rarely if ever camelCased)
- `UtcDatetime` -> `DateTime` (`chrono::DateTime` uses camelCase)

In the next step, I made sure each `ElementType` case shared the same name with its' corresponding `Bson` case:
- `FloatingPoint` -> `Double`
- `Utf8String` -> `String`
- `Boolean` -> `Bool`
- `UtcDatetime` -> `DateTime`
- `NullValue` -> `Null`
- `RegularExpression` -> `Regex`
- `Integer32Bit` -> `I32`
- `TimeStamp` -> `Timestamp`
- `Integer64Bit` -> `I64`
- `Decimal128Bit` -> `Decimal128`

One thing to note is that I did not rename `ElementType::EmbeddedDocument`, since a top level document by definition does not have an element type, since it isn't an element. Let me know if this should just be `Document` to reduce confusion.


src/ordered.rs
```diff
@@ -214,7 +214,7 @@ impl OrderedDocument {
     /// the correct type.
     pub fn get_f64(&self, key: &str) -> ValueAccessResult<f64> {
```
patrickfreed, line 215: One thing to note, I left these as `get_f64`, since they really are getting `f64` values, not `Bson::Double` values. I wasn't sure if this was exactly the right choice, and considered adding a `get_double` alias, or replacing this with `get_double`. In the end, I decided to just leave it as-is, but am open to suggestions.

cc @saghm

saghm: I'm fine leaving it as-is as well

src/bson.rs
```diff
@@ -58,18 +58,18 @@ pub enum Bson {
     JavaScriptCode(String),
     /// JavaScript code w/ scope
     JavaScriptCodeWithScope(JavaScriptCodeWithScope),
-    /// 32-bit integer
+    /// 32-bit signed integer
     I32(i32),
```
mbroadst, line 62: These types are named `Int32`, `Int64` and `Uint64` in the spec, it seems inconsistent to use something different just for them (I imagine they are named this for parity with the built-in types, but that itself might be somewhat confusing)

src/bson.rs
```diff
@@ -41,15 +41,15 @@ use crate::{
...
+    Double(f64),
     /// UTF-8 string
     String(String),
     /// Array
     Array(Array),
     /// Embedded document
     Document(Document),
     /// Boolean value
-    Boolean(bool),
+    Bool(bool),
```
mbroadst, line 52: Why not `Boolean`, it seems pretty consistently named that in the spec as well as other drivers.

src/bson.rs
```diff
@@ -373,24 +373,24 @@ impl Bson {
...
+            Bson::Double(..) => ElementType::Double,
+            Bson::String(..) => ElementType::String,
             Bson::Array(..) => ElementType::Array,
             Bson::Document(..) => ElementType::EmbeddedDocument,
-            Bson::Boolean(..) => ElementType::Boolean,
-            Bson::Null => ElementType::NullValue,
-            Bson::Regex(..) => ElementType::RegularExpression,
+            Bson::Bool(..) => ElementType::Bool,
+            Bson::Null => ElementType::Null,
+            Bson::Regex(..) => ElementType::Regex,
```
mbroadst, line 382: mild preference for naming the actual type `RegularExpression`, no need to conserve on bytes

src/bson.rs
patrickfreed: I gave a little bit of the rationale behind how the names were chosen in the PR's description. It basically boils down to this: 
- what other drivers / the spec refers to defines the type. 
- If there is an native Rust equivalent type that corresponds to it with the same or abbreviated name, use that. 
- Otherwise, use the full name. 

So for `Double`, since there is no Rust `doub` or `double`, we just use the full `Double`. For `Boolean`, we use the Rust `Bool` terminology.

[`serde_json` similarly refers to this case as `Bool`](https://docs.serde.rs/serde_json/enum.Value.html#variant.Bool), despite the generally accepted name for the type to be "boolean" in JSON.

src/bson.rs
patrickfreed: See my comment on `Bool` or the PR's description for the rationale behind the names for the integers.

`Timestamp` is a type we introduce ourselves, since it is a much easier API for users to use a struct instead of manually bit shifting an unsigned integer.

src/bson.rs
patrickfreed: Same rationale here again. It's less about conserving on bytes and more about consistency with the existing ecosystem. Both the crate and the type for regular expressions in Rust are `regex`.

src/bson.rs
mbroadst: According to the description your order of priority is consistency:
- with each other
- with other MongoDB drivers
- with the Rust ecosystem as a whole

I would argue that you probably want to add a "with the bson spec" somewhere next to "with other MongoDB drivers". 

Am I interpreting this list of priority backwards? Otherwise it seems like there's pretty overwhelming support for calling it `Boolean`, considering there is no mention of "bool" in the bsonspec, a quick glance at the top 4 most used drivers use the term `Boolean`. Additionally it is not a Rust `Bool`, it is a BSON wrapper of that type called `Boolean` - there is additional type information here that distinguishes it.

patrickfreed: The order of priorities you state is correct, but it's more of an "assembly line" type structure than discrete priorities. The "with the Rust ecosystem as a whole" part only affects the name with respect to spelling and abbreviations, not the actual name itself. It's why we name the case `Double` and not `f64` or `FloatingPoint`. It's also why I opted for `I32` over `Int32`--they're both abbreviations, so we might as well use the Rust abbreviation, right? (I do agree that it's unfortunate that the Rust language opted for nonstandard abbreviations).

So for boolean, both the spec and the other drivers refer to it as `Boolean`, which means that the type we're modeling here is named `Boolean`. Only then do we look to the Rust ecosystem to see if there's a generally accepted way to refer to boolean types (which there is, `bool`).

>Additionally it is not a Rust Bool, it is a BSON wrapper of that type called Boolean - there is additional type information here that distinguishes it.

I think this is a good point, and is sort of where I came up with my rationale as a compromise between consistency with other drivers and consistency with what Rust developers expect. (the spec/drivers defines the name, the ecosystem defines the spelling)

If that rationale all makes sense and you still think we should opt for the full types (`Boolean` / `Int32` / `Int64` / `RegularExpression`), I'm happy to make that change. I just wanted to make sure the reasoning behind the names was fully communicated before we moved off of them.

patrickfreed: After offline discussion, we opted for the longer names, both for clarity and to emphasize the distinction between them being BSON values rather than Rust values.

PR #166, opened 2020-05-13
RUST-346 Implement bson corpus spec tests

saghm:


etc/update-spec-tests.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
saghm: This is copied from the driver, with the exception of the path being changed from `src/test` to `tests`

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
         match a {
             Value::Number(x) => x
                 .as_i64()
-                .map(Bson::from)
+                .map(|i| {
+                    if i >= i32::MIN as i64 && i <= i32::MAX as i64 {
```
saghm, line 308: Previously, we were reading converting every JSON integer to an i64. The spec requires that we use the smallest size integer possible, so this logic checks to see if it can fit in an i32.

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
+            ),
             Value::Null => Bson::Null,
         }
     }
 }
 
 impl From<Bson> for Value {
     fn from(bson: Bson) -> Self {
-        match bson {
+        bson.into_relaxed_extjson()
```
saghm, line 330: The spec indicates that if there's a default conversion, it needs to use relaxed.

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
+    }
+}
+
+impl Bson {
+    pub fn into_relaxed_extjson(self) -> Value {
```
saghm, line 335: I'm on the fence about whether this should take ownership of `self` or if it should borrow and clone

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
+        match self {
+            Bson::FloatingPoint(v) if v.is_nan() => {
+                let s = if v.is_sign_negative() { "-NaN" } else { "NaN" };
```
saghm, line 338: I figured it was more straightforward to manually implement the stringification of +/-Nan and Infinity rather than trying to rely on the default formatting that Rust gives

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
+                Value::Object(v.into_iter().map(|(k, v)| (k, Value::from(v))).collect())
+            }
             Bson::Boolean(v) => json!(v),
             Bson::Null => Value::Null,
-            Bson::Regex(Regex { pattern, options }) => json!({
-                "$regex": pattern,
-                "$options": options
-            }),
+            Bson::Regex(Regex { pattern, options }) => {
+                let mut chars: Vec<_> = options.chars().collect();
```
saghm, line 360: Interestingly, the spec requires that the options string needs to be alphabetized

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
-                    "type": tval,
-                    "$binary": hex::encode(bytes),
+                    "$binary": {
+                        "base64": base64::encode(bytes),
+                        "subType": hex::encode([tval.into()]),
+                    }
+                })
+            }
+            Bson::ObjectId(v) => json!({"$oid": v.to_hex()}),
+            Bson::UtcDatetime(v) if v.timestamp_millis() >= 0 && v.year() <= 99999 => {
```
saghm, line 395: For relaxed extended JSON, the spec requires that dates from `[unix epoch, Y10000)` are formatted as strings rather than specified as integers since the epoch.

src/bson.rs
```diff
@@ -304,59 +304,108 @@ impl From<Value> for Bson {
...
+                let seconds_format = if v.timestamp_subsec_millis() == 0 {
```
saghm, line 396: While the spec indicates that millisecond precision should be used in the formatted string, it strangely leaves this off for the test with the unix epoch. I assume that this means that datetimes that are a whole number of seconds should omit the `.000`, so I do that here.

src/bson.rs
```diff
@@ -366,12 +415,54 @@ impl From<Bson> for Value {
...
+                        "$oid": id.to_hex()
+                    }
+                }
+            }),
+        }
+    }
+
+    pub fn into_canonical_extjson(self) -> Value {
+        match self {
+            Bson::I32(i) => json!({ "$numberInt": i.to_string() }),
```
saghm, line 431: Most cases are identical in canonical JSON to relaxed; the exceptions are numbers, datetimes, and from a recursive perspective, arrays and documents, since they could contain values that are numbers or datetimes.

src/bson.rs
```diff
@@ -366,12 +415,54 @@ impl From<Bson> for Value {
...
+            }
+            Bson::Array(arr) => {
+                Value::Array(arr.into_iter().map(Bson::into_canonical_extjson).collect())
+            }
+            Bson::Document(arr) => Value::Object(
+                arr.into_iter()
+                    .map(|(k, v)| (k, v.into_canonical_extjson()))
+                    .collect(),
+            ),
+            Bson::JavaScriptCodeWithScope(JavaScriptCodeWithScope { code, scope }) => json!({
```
saghm, line 457: To my surprise, there is actually test that checks that numbers are encoded canonically (rather than relaxed) inside the `$scope` document.

src/bson.rs
```diff
@@ -402,16 +493,22 @@ impl Bson {
 
     /// Converts to extended format.
     /// This function mainly used for [extended JSON format](https://docs.mongodb.com/manual/reference/mongodb-extended-json/).
-    #[doc(hidden)]
-    pub fn to_extended_document(&self) -> Document {
+    pub(crate) fn to_extended_document(&self) -> Document {
```
saghm, line 496: This function is a bit weird; it takes a BSON value and converts it into a document that represents the extended JSON, but still is BSON rather than JSON. I originally played around with getting rid of it, but it's used in deep parts of the serde implementation, so to avoid scope creep, I kept it around but converted it to mirror the new extended JSON implementation. I also figured that instead of just hiding it from the docs, it was better to just not export it. On the off-chance that we decide it needs to be part of the API again, changing it back to `pub` isn't a breaking change, so we could do that later.

src/bson.rs
```diff
@@ -486,184 +595,243 @@ impl Bson {
...
-                );
-            } else if let Ok(sym) = values.get_str("$symbol") {
-                return Bson::Symbol(sym.to_owned());
-            } else if let Ok(dec) = values.get_str("$numberDecimal") {
-                return Bson::Decimal128(dec.parse::<Decimal128>().unwrap());
-            } else if let Ok(undefined) = values.get_bool("$undefined") {
-                if undefined {
-                    return Bson::Undefined;
+    pub(crate) fn from_extended_document(doc: Document) -> Bson {
+        if doc.len() > 2 {
```
saghm, line 609: I found the previous implementation of this to be rather unwieldly, so I took a stab at cleaning it up. The one downside with this new implementation is that it needs to allocate a Vec for the keys to be able to pattern match them as a slice, but this only needs to happen if there are two or fewer keys in the document, and the values are just &str, which are not very large, so I think it's fine. If we're super worried about performance here, we could also look into using [`smallvec`](https://crates.io/crates/smallvec).

src/bson.rs
```diff
@@ -486,184 +595,243 @@ impl Bson {
...
+            ["$numberDecimal"] => {
+                if let Ok(d) = doc.get_str("$numberDecimal") {
+                    if let Ok(d) = d.parse() {
+                        return Bson::Decimal128(d);
+                    }
+                }
+            }
+
+            ["$binary"] => {
+                if let Ok(binary) = doc.get_document("$binary") {
```
saghm, line 668: Things like this are a bit messy, but I'm hesitant to use serde as part of the extended JSON parsing

src/bson.rs
```diff
@@ -486,184 +595,243 @@ impl Bson {
...
+                    }
+                }
+            }
+
+            ["$date"] => {
+                if let Ok(date) = doc.get_i64("$date") {
+                    let mut num_secs = date / 1000;
+                    let mut num_millis = date % 1000;
+
+                    // The chrono API only lets us create a DateTime with an i64 number of
```
saghm, line 763: This might be the largest ratio of size of comment to size of code the comment is documenting that I've eve written.

src/decoder/mod.rs
```diff
@@ -167,16 +167,8 @@ fn decode_array<R: Read + ?Sized>(reader: &mut R, utf8_lossy: bool) -> DecoderRe
...
-        match key.parse::<usize>() {
-            Err(..) => return Err(DecoderError::InvalidArrayKey(arr.len(), key)),
-            Ok(idx) => {
-                if idx != arr.len() {
-                    return Err(DecoderError::InvalidArrayKey(arr.len(), key));
-                }
-            }
-        }
+        // read key, but do no validate value
+        read_cstring(reader)?;
```
saghm, line 171: The corpus tests that "degenerate" BSON where the keys of an array document are not increasing integers is still parsed correctly, so I removed this validation section.

src/decoder/error.rs
```diff
@@ -8,7 +8,6 @@ pub enum DecoderError {
     IoError(io::Error),
     FromUtf8Error(string::FromUtf8Error),
     UnrecognizedElementType(u8),
-    InvalidArrayKey(usize, String),
```
saghm, line 11: This is no longer necessary;  see below

src/decoder/mod.rs
```diff
@@ -196,16 +188,25 @@ fn decode_bson<R: Read + ?Sized>(reader: &mut R, tag: u8, utf8_lossy: bool) -> D
         Some(ElementType::EmbeddedDocument) => decode_document(reader).map(Bson::Document),
         Some(ElementType::Array) => decode_array(reader, utf8_lossy).map(Bson::Array),
         Some(ElementType::Binary) => {
-            let len = read_i32(reader)?;
+            let mut len = read_i32(reader)?;
```
saghm, line 191: The tests check that we automatically handle the repeated length field in binary subtype 2, so I updated the logic here to do that

src/encoder/mod.rs
```diff
@@ -152,8 +155,19 @@ fn encode_bson<W: Write + ?Sized>(writer: &mut W, key: &str, val: &Bson) -> Enco
         Bson::I64(v) => write_i64(writer, v),
         Bson::TimeStamp(ts) => write_i64(writer, ts.to_le_i64()),
         Bson::Binary(Binary { subtype, ref bytes }) => {
-            write_i32(writer, bytes.len() as i32)?;
+            let len = if let BinarySubtype::BinaryOld = subtype {
```
saghm, line 158: Ditto above about the repeated length binary in subtype 2

tests/modules/bson.rs
```diff
@@ -131,12 +131,12 @@ fn from_impls() {
         Bson::Document(doc! {"a": "b"})
     );
 
-    let db_pointer = Bson::from_extended_document(doc! {
+    let db_pointer = Bson::from(json!({
```
saghm, line 134: Interestingly, now that `from_extended_document` is private, the only ways to create a `DbPointer` are through JSON or raw bytes. Since this is just a test, I picked the easier one.

tests/serde.rs
```diff
@@ -114,7 +115,7 @@ fn test_ser_regex() {
 
     let regex = Regex {
         pattern: "12".into(),
-        options: "10".into(),
+        options: "01".into(),
```
saghm, line 118: Since regex options are now alphabetized, the easiest way to fix these tests was to specify already-alphabetized options. These tests are just checking serde, so we can rely on the corpus tests to validate the alphabetization behavior.

tests/spec/corpus.rs
```diff
@@ -0,0 +1,226 @@
...
+        let description = format!("{}: {}", test.description, valid.description);
+
+        let bson_to_native_cb = bson::decode_document(
+            &mut hex::decode(&valid.canonical_bson)
+                .expect(&description)
+                .as_slice(),
+        )
+        .expect(&description);
+
+        let mut native_to_bson_bson_to_native_cv = Vec::new();
```
saghm, line 67: The long names like this map directly to the descriptions in the corpus test README; they're annoying to read in a vaccuum, but they make it much easier to verify that we're in fact testing what the corpus tests are asking us to.

tests/spec/corpus.rs
```diff
@@ -0,0 +1,226 @@
...
+        bson::encode_document(&mut native_to_bson_bson_to_native_cv, &bson_to_native_cb)
+            .expect(&description);
+
+        let cej: serde_json::Value =
+            serde_json::from_str(&valid.canonical_extjson).expect(&description);
+
+        // native_to_bson( bson_to_native(cB) ) = cB
+
+        if !description.contains("1.2345678921232E+18") {
```
patrickfreed, line 76: i think a comment explaining each of these skips would be helpful

src/bson.rs
saghm: Sounds good; I'll leave it as-is then

src/bson.rs
saghm: Filed RUST-426 and added the TODO.

src/bson.rs
```diff
@@ -486,184 +595,242 @@ impl Bson {
...
-                return Bson::Decimal128(dec.parse::<Decimal128>().unwrap());
-            } else if let Ok(undefined) = values.get_bool("$undefined") {
-                if undefined {
-                    return Bson::Undefined;
+    pub(crate) fn from_extended_document(doc: Document) -> Bson {
+        if doc.len() > 2 {
+            return Bson::Document(doc);
+        }
+
+        let keys: Vec<_> = doc.keys().map(|s| s.as_str()).collect();
```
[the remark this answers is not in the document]
saghm: Oh, good point; I was naively thinking of the pattern matching as doing set comparison, when in reality it won't do that at all. I've updated it to sort the keys before matching (the keys in the branches themselves happened to already be listed in proper order).

src/bson.rs
saghm: This is actually trickier than it looks for a few reasons. First, we're trying to get multiple keys from the document, which means we won't be able to use a single long chain. It would look more like this:

```rust
doc.get_document("$binary")
    .and_then(|binary| {
	    let bytes = binary.get_str("base64").and_then(|bytes|base64::decode(bytes));
		let subtype = binary.get_str("subType").and_then(|subtype| hex::decode(subtype));

        // etc.
	})
```

This isn't too bad, except the error types aren't the same; the first error in each subchain is a `bson::ordered::ValueAccessError`, but then they switch to errors from separate crates. `and_then` requires that the same error type is used for for `self` and for the return value. 

We also run into issues with the innermost `if` statement, which isn't checking for `Ok` but for the length being 1. We'd have to construct some sort of error for cases where the subtype is longer than one byte. Finally, we'd still need an `if let` at the top level, since we're not actually returning any error from this branch.

Overall, I'm not sure this is much cleaner than what we have here. I think dealing with all three of the above issues would end up with something that's harder to understand at a glance than what we have now, despite its ugliness. That said, there's still probably room for improvement. My first instinct is to abstract all this out to method that returns `Option<Bson>`, which would let us use `.ok()?` for all the result cases and return `None` if the subtype is not 1. That would let us get away with a single `if let` for the branch inside this method.

patrickfreed: >My first instinct is to abstract all this out to method that returns Option<Bson>, which would let us use .ok()? for all the result cases and return None if the subtype is not 1.

Yeah that seems like a good idea.

tests/spec/corpus.rs
```diff
@@ -0,0 +1,226 @@
...
+                    "{}",
+                    description,
+                );
+            }
+
+            // native_to_bson( json_to_native(dEJ) ) = cB (unless lossy)
+
+            if valid.lossy != Some(true) {
+                let mut native_to_bson_json_to_native_dej = Vec::new();
+                bson::encode_document(
```
patrickfreed, line 186: this seems like a helpful method to have on `Document` iteself. Like `to_vec` or `to_bytes` or something.

tests/spec/corpus.rs
```diff
@@ -0,0 +1,226 @@
...
+                        description,
+                    );
+                }
+            }
+        }
+
+        // native_to_relaxed_extended_json( json_to_native(rEJ) ) = rEJ
+
+        if let Some(ref rej) = valid.relaxed_extjson {
+            let json_to_native_rej: Bson = cej.clone().into();
```
patrickfreed, line 206: this should be `rej.clone().into()`

tests/spec/corpus.rs
saghm: I'm ambivalent about putting it in a public API; documents can get pretty large, and giving an API that's generic over `Write` (like `encode_document is) is generally the more idiomatic way to do something like this. Either way, I don't think it's a blocker for 1.0.

tests/spec/corpus.rs
saghm: Oh shoot, good catch! I knew that these long names were worth it

tests/spec/corpus.rs
saghm: Done

tests/spec/corpus.rs
patrickfreed: I was more referring to the namespacing than the output type. It seems strange to me that `encode_document` and `decode_document` are both free floating functions in the `bson` module rather than methods under the `Document` namespace it self.

e.g.
```rust
json_to_native_dej.as_document().encode(&mut native_to_bson_json_to_native_dej).unwrap()
let doc = Document::decode(&mut native_to_bson_json_to_native_dej).unwrap()
```
The current free floating way seems very C like to me, whereas this feels more in line with Rust.

tests/spec/corpus.rs
patrickfreed: Rather than outright skipping these tests, I think it'd be better to attempt to normalize the test files to match Rust's formatting as the spec suggests.

tests/spec/corpus.rs
saghm: I think it's intended to provide a similar API to `serde_json`, which uses free-floating functions for these things: https://docs.rs/serde_json/1.0.53/serde_json/#functions

patrickfreed: Those make sense as free floating because they're generic on the type being encoded/decoded; they can't be methods by nature of that. In this case, we have a concrete `Document` which we're encoding/decoding.

This convinces me further that, if we are to retain this method's functionality in the public API, we should restrict it to the `Document` namespace. That way, if/when we ever decide to match `serde_json`'s functionaltiy wrt to these pure generic functions (e.g. by adding `from_reader`), we won't have to deprecate the now redundant `from_document`.

This suggestion at this point certainly is out of scope for this PR, so I filed [RUST-429](https://jira.mongodb.org/browse/RUST-429) to discuss it further.

src/bson.rs
patrickfreed: lgtm

tests/spec/corpus.rs
saghm: Discussed offline

tests/spec/corpus.rs
saghm: Based on offline discussion, we'll manually inspect all $numberDouble values for either "e" or "E" and then convert it to a standard notation string

src/bson.rs
```diff
@@ -302,59 +302,108 @@ impl From<Value> for Bson {
...
 
 impl From<Bson> for Value {
     fn from(bson: Bson) -> Self {
-        match bson {
+        bson.into_relaxed_extjson()
+    }
+}
+
+impl Bson {
+    pub fn into_relaxed_extjson(self) -> Value {
```
patrickfreed, line 333: this could use a docstring

src/bson.rs
```diff
@@ -364,12 +413,54 @@ impl From<Bson> for Value {
...
+                    "$ref": namespace,
+                    "$id": {
+                        "$oid": id.to_hex()
+                    }
+                }
+            }),
+        }
+    }
+
+    pub fn into_canonical_extjson(self) -> Value {
```
patrickfreed, line 427: also docstring

tests/spec/corpus.rs
```diff
@@ -0,0 +1,253 @@
...
+use bson::Bson;
+use pretty_assertions::assert_eq;
+use serde_derive::Deserialize;
+
+use super::run_spec_test;
+
+#[derive(Debug, Deserialize)]
+struct TestFile {
+    description: String,
+    bson_type: String,
```
patrickfreed, line 12: could this be `ElementType`? Then below the skips could compare based on enum cases rather than strings.

tests/spec/corpus.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut native_to_bson_bson_to_native_cv = Vec::new();
+        bson::encode_document(&mut native_to_bson_bson_to_native_cv, &bson_to_native_cb)
+            .expect(&description);
+
+        let cej: serde_json::Value =
+            serde_json::from_str(&valid.canonical_extjson).expect(&description);
+
+        // native_to_bson( bson_to_native(cB) ) = cB
+
+        // Rust doesn't format f64 with exponential notation by default, and the spec doesn't give
```
patrickfreed, line 78: I think this comment doesnt need to be here.

tests/spec/corpus.rs
saghm: We can't put it in the struct directly since ElementType doesn't deserialize from strings. While we could parse it in the body of `run_test`, it would be tough to use it for decimal since currently `ElementType::Decimal128Bit` is only defined if the `decimal` feature is enabled, so it would require some boilerplate to define the checks properly.

tests/spec/corpus.rs
saghm: Ah, yep, that's leftover from when we were skipping.

src/bson.rs
saghm: Updated

src/bson.rs
saghm: Updated

src/bson.rs
patrickfreed: this and the other one could use a link to the official documentation: https://docs.mongodb.com/manual/reference/mongodb-extended-json/

saghm: Updated again

PR #169, opened 2020-05-14
RUST-313 Mark enums as non-exhaustive where appropriate

saghm:


saghm:
I think when we triaged this ticket, we decided not to put it on BSON. I think that bumping the major version if/when new BSON types are created is a small cost that's worth it to make it easier for users to use `Bson` in an exhaustive fashion.

saghm:
Despite it saying there were merge conflicts, rebasing locally didn't give any conflicts. Another point for team rebase.

PR #170, opened 2020-05-14
RUST-217 Rephrase unsupported/corrupt BSON messages

patrickfreed:
[RUST-217](https://jira.mongodb.org/browse/RUST-217)

This PR adds the key to the error returned when encountering an unsupported/corrupt BSON type.

### DIFF
--- a/src/decoder/error.rs
+++ b/src/decoder/error.rs
@@ -7,7 +7,7 @@ use serde::de::{self, Expected, Unexpected};
 pub enum DecoderError {
     IoError(io::Error),
     FromUtf8Error(string::FromUtf8Error),
-    UnrecognizedElementType(u8),
+    UnrecognizedElementType { key: String, element_type: u8 },
     InvalidArrayKey(usize, String),
     // A field was expected, but none was found.
     ExpectedField(&'static str),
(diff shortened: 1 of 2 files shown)


src/decoder/error.rs
```diff
@@ -7,7 +7,7 @@ use serde::de::{self, Expected, Unexpected};
 pub enum DecoderError {
     IoError(io::Error),
     FromUtf8Error(string::FromUtf8Error),
-    UnrecognizedElementType(u8),
+    UnrecognizedElementType { key: String, element_type: u8 },
```
patrickfreed, line 10: If we ever publicly support decoding `Bson` from a reader without it being part of a `Document`, this key field could present a problem. Though I honestly think the empty string or some other specific value like "TOP LEVEL" wouldn't be that bad. We could always add a `TopLevelUnrecognizedElementType` error as an alternative.

saghm: Could we just make key an `Option`? Alternately, we could name this variant something specific to the fact that it's for values in documents to leave room for us to add a separate variant for unnested values later.

patrickfreed: I'd rather use separate cases than `Option` personally. I renamed it to `UnrecognizedDocumentElementType`, let me know if you think that works.

saghm: Yep, works for me

PR #171, opened 2020-05-18
Map undefined to None

IcanDivideBy0:
Using bson with current mongodb driver, if a Deserialize target has an `Option<>` field, where the database holding an undefined value, it try to deserialise `undefined` ad the content of the Option.
I think most peopole will expect `undefined` to be deseralized to `None` when possible.

Also, can you consider making a release soon, having any `undefined` value in database makes it impossible to work with the current mongodb driver, thanks.

### DIFF
--- a/src/decoder/serde.rs
+++ b/src/decoder/serde.rs
@@ -312,6 +312,7 @@ impl<'de> Deserializer<'de> for Decoder {
             }
             Bson::Boolean(v) => visitor.visit_bool(v),
             Bson::Null => visitor.visit_unit(),
+            Bson::Undefined => visitor.visit_unit(),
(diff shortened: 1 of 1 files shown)


saghm:
Hello! While we understand that deserializing `undefined` into options might make things easier in some cases of existing data, `undefined` is a deprecated BSON type, and we don't want to encourage users to use the type in any new cases. Because of that, we've decided not to implement automatic deserialization of `undefined` into `Option` fields. As a workaround, you can use serde's [`deserialize_with` attribute](https://serde.rs/field-attrs.html#deserialize_with) to annotate fields with a custom function to deserialize undefined into `Option` types where you need it.

PR #173, opened 2020-05-21
Various error improvements

patrickfreed:
[RUST-430](https://jira.mongodb.org/browse/RUST-430) / [RUST-437](https://jira.mongodb.org/browse/RUST-437) / [RUST-443](https://jira.mongodb.org/browse/RUST-443) / [RUST-444](https://jira.mongodb.org/browse/RUST-444) / [RUST-312](https://jira.mongodb.org/browse/RUST-312)

This PR makes a number of improvements to the errors in the BSON library. It also fixes a bug where the serde `deny_unknown_fields` attribute was being ignored.

### DIFF
--- a/src/decoder/error.rs
+++ b/src/decoder/error.rs
@@ -100,27 +100,6 @@ impl fmt::Display for DecoderError {
 }
 
 impl error::Error for DecoderError {
-    fn description(&self) -> &str {
-        match *self {
-            DecoderError::IoError(ref inner) =>
-            {
-                #[allow(deprecated)]
-                inner.description()
-            }
-            DecoderError::FromUtf8Error(ref inner) =>
-            {
-                #[allow(deprecated)]
-                inner.description()
-            }
-            DecoderError::UnrecognizedDocumentElementType { .. } => "unrecognized element type",
-            DecoderError::InvalidArrayKey { .. } => "invalid array key",
-            DecoderError::SyntaxError { ref message } => message.as_str(),
-            DecoderError::EndOfStream => "end of stream",
-            DecoderError::DeserializationError { ref message } => message.as_str(),
-            DecoderError::InvalidTimestamp(..) => "no such local time",
-            DecoderError::AmbiguousTimestamp(..) => "ambiguous local time",
-        }
-    }
     fn cause(&self) -> Option<&dyn error::Error> {
         match *self {
             DecoderError::IoError(ref inner) => Some(inner),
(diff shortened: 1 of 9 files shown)


src/decoder/serde.rs
```diff
@@ -611,7 +623,6 @@ impl<'de> MapAccess<'de> for MapDecoder {
                 let de = Decoder::new(Bson::String(key));
                 match seed.deserialize(de) {
                     Ok(val) => Ok(Some(val)),
-                    Err(DecoderError::UnknownField(_)) => Ok(None),
```
patrickfreed, line 614: this allows `deny_unknown_fields` to be respected.

src/decoder/error.rs
```diff
@@ -1,39 +1,60 @@
...
+        /// The encountered element type.
+        element_type: u8,
+    },
+
+    /// There was an error with the syntactical structure of the BSON.
+    SyntaxError { message: String },
+
+    /// The end of the BSON input was reached too soon.
     EndOfStream,
-    // Invalid Type
```
patrickfreed, line 21: Removing these fields is part of the work of RUST-444. This ticket hadn't gone through triage yet, but seeing as the clock is ticking down I just went ahead and implemented it. I can always revert it if we end up thinking it wasn't the right idea, but I figured it was unlikely that would happen.

src/decoder/error.rs
```diff
@@ -1,39 +1,60 @@
...
-use serde::de::{self, Expected, Unexpected};
+use crate::Bson;
+use de::Unexpected;
+use serde::de;
 
 /// Possible errors that can arise during decoding.
 #[derive(Debug)]
 #[non_exhaustive]
 pub enum DecoderError {
+    /// A `std::io::Error` encountered while deserializing.
```
saghm, line 11: Should we make this (and the other std error types in the docs below) links?

src/decoder/error.rs
```diff
@@ -98,17 +113,10 @@ impl error::Error for DecoderError {
                 inner.description()
             }
             DecoderError::UnrecognizedDocumentElementType { .. } => "unrecognized element type",
-            DecoderError::InvalidArrayKey(..) => "invalid array key",
```
saghm, line 101: Should we take this time to remove the explicit `description` implementations on our errors given that it's deprecated (and has a default implementation)?

src/decoder/error.rs
patrickfreed: done

src/decoder/error.rs
patrickfreed: done. Doing this also made me realize I forgot to document / label the `oid` error cases, so I've done that as well.

PR #175, opened 2020-05-26
RUST-429 Move encode_document/decode_document to Document namespace

isabelatkinson:


src/document.rs
```diff
@@ -507,6 +513,43 @@ impl Document {
             inner: self.inner.entry(k),
         }
     }
+
+    /// Attempts to encode the `Document` into a byte stream.
+    pub fn encode_document<W: Write + ?Sized>(&self, writer: &mut W) -> EncoderResult<()> {
```
patrickfreed, line 518: given that these are now methods on `Document`, we can drop it from the function name. So we can just call `document.encode()` or `Document::decode(reader)`.

src/encoder/mod.rs
```diff
@@ -89,42 +89,19 @@ fn encode_array<W: Write + ?Sized>(writer: &mut W, arr: &[Bson]) -> EncoderResul
     Ok(())
 }
 
-/// Attempt to encode a `Document` into a byte stream.
-///
-/// Can encode any type which is iterable as `(key: &str, value: &Bson)` pairs,
```
patrickfreed, line 94: One thing I didn't notice when I filed this ticket is that `encode_document` is generic (kind of) over the input "document". Moving this to be a method on `Document` itself does lose this functionality, but is that really a problem? I think having an iterable over `(str, Bson)` that isn't a `Document` would be pretty rare. If we did want to have a generic "encode" function, we'd probably want to bound the input on `Serialize` anyways (like the `serde_json` [equivalent](https://docs.rs/serde_json/1.0.53/serde_json/fn.to_writer.html)), not on `IntoIterator`.

cc @saghm

saghm: I think this is yet another case of "let's wait until someone asks for it". Note that users can still replicate this functionality at the cost of some performance with `foo.cloned().collect::<Document>().encode(...)`, so I think it's reasonable to wait and see if anyone absolutely needs to avoid copying when encoding some non-Document iterator.

PR #176, opened 2020-05-26
RUST-438 Add more detailed documentation examples

patrickfreed:
[RUST-438](https://jira.mongodb.org/browse/RUST-438)

This PR updates the library documentation for the `bson` crate. It is heavily influenced by [the library docs for `serde_json`](https://docs.rs/serde_json/1.0.53/serde_json/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ let array = bson!([5, false]);
 ```
 [`bson!`](https://docs.rs/bson/latest/bson/macro.bson.html) has supports both array and object literals, and it automatically converts any values specified to [`Bson`](https://docs.rs/bson/latest/bson/enum.Bson.html), provided they are `Into<Bson>`.
 
-#### [`Bson`](enum.Bson.html) value unwrapping
+#### [`Bson`](https://docs.rs/bson/latest/bson/enum.Bson.html) value unwrapping
 
 [`Bson`](https://docs.rs/bson/latest/bson/enum.Bson.html) has a number of helper methods for accessing the underlying native Rust types. These helpers can be useful in circumstances in which the specific type of a BSON value
 is known ahead of time.
--- a/README.md
+++ b/README.md
@@ -108,10 +108,8 @@ BSON documents are ordered maps of UTF-8 encoded strings to BSON values. They ar
 [`Document`](https://docs.rs/bson/latest/bson/document/struct.Document.html)s can be created directly either from a byte
 reader containing BSON data or via the `doc!` macro:
 ```rust
-# use bson::doc;
-# use std::io::Read;
 let mut bytes = hex::decode("0C0000001069000100000000").unwrap();
-let doc = bson::decode_document(&mut bytes.as_slice()).unwrap(); // { "i": 1 }
+let doc = Document::decode(&mut bytes.as_slice()).unwrap(); // { "i": 1 }
 
 let doc = doc! {
    "hello": "world",
--- a/README.md
+++ b/README.md
@@ -7,9 +7,20 @@
 
 Encoding and decoding support for BSON in Rust
 
+## Index
+- [Overview of BSON Format](#overview-of-bson-format)
+- [Usage](#usage)
+    - [BSON Values](#bson-values)
+    - [BSON Documents](#bson-documents)
+    - [Modeling BSON with strongly typed data structures](#modeling-bson-with-strongly-typed-data-structures)
+- [Breaking Changes](#breaking-changes)
+- [Contributing](#contributing)
+- [Running the Tests](#running-the-tests)
+- [Continuous Integration](#continuous-integration)
+
 ## Useful links
 - [API Documentation](https://docs.rs/bson/)
-- [Serde](https://serde.rs/)
+- [Serde Documentation](https://serde.rs/)
 
 ## Installation
 This crate works with Cargo and can be found on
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -59,7 +205,7 @@ pub use self::{
 };
 
 #[macro_use]
-pub mod macros;
+mod macros;
```
patrickfreed, line 208: from looking at the rendered documentation, it didn't seem like this module needed to be exported.

src/lib.rs
```diff
@@ -19,25 +19,171 @@
...
+//! let array = bson!([5, false]);
+//! ```
+//! [`bson!`](macro.bson.html) has supports both array and object literals, and it automatically
+//! converts any values specified to [`Bson`](enum.Bson.html), provided they are either `Serialize`
+//! or `Into<Bson>`.
+//!
+//! ### [`Bson`](enum.Bson.html) value unwrapping
+//!
+//! [`Bson`](enum.Bson.html) has a number of helper methods for accessing the underlying native Rust
+//! types. These helpers can be useful in cirumstances in which a the specific type of a BSON value
```
isabelatkinson, line 76: extra "a" after which

src/lib.rs
```diff
@@ -19,25 +19,171 @@
...
+//!
+//! ### Creating [`Document`](document/struct.Document.html)s
+//!
+//! [`Document`](document/struct.Document.html)s can be created directly either from a byte
+//! reader containing BSON data or via the `doc!` macro:
+//! ```rust
+//! # use bson::doc;
+//! # use std::io::Read;
+//! let mut bytes = hex::decode("0C0000001069000100000000").unwrap();
+//! let doc = bson::decode_document(&mut bytes.as_slice()).unwrap(); // { "i": 1 }
```
isabelatkinson, line 105: Not sure if you'd want to fix this in this PR, but this should change to `Document::decode` since we're updating the name [here](https://github.com/mongodb/bson-rust/pull/175#pullrequestreview-418789367)

src/lib.rs
```diff
@@ -19,25 +19,171 @@
...
 //!
-//! let mut doc = Document::new();
-//! doc.insert("foo".to_owned(), Bson::String("bar".to_owned()));
+//! // attempt get values as untyped Bson
+//! let none = doc.get("asdfadsf"); // None
+//! let value = doc.get("string"); // Some(&Bson::String("string"))
 //!
-//! let mut buf = Vec::new();
-//! encode_document(&mut buf, &doc).unwrap();
+//! // attempt to get values with explcit typing
```
isabelatkinson, line 134: explicit sp.

src/lib.rs
```diff
@@ -19,25 +19,171 @@
...
+//! let int: Bson = bson!(5);
+//! let array = bson!([5, false]);
+//! ```
+//! [`bson!`](macro.bson.html) has supports both array and object literals, and it automatically
+//! converts any values specified to [`Bson`](enum.Bson.html), provided they are either `Serialize`
+//! or `Into<Bson>`.
+//!
+//! ### [`Bson`](enum.Bson.html) value unwrapping
+//!
+//! [`Bson`](enum.Bson.html) has a number of helper methods for accessing the underlying native Rust
```
saghm, line 75: Does the macro actually accept things that are `Serialize` but not `Into<Bson>`? I'm not sure I've ever tried this.

src/lib.rs
saghm: Assuming the macros are still getting exported fine, it's better to have it private

src/lib.rs
patrickfreed: fixed

src/lib.rs
patrickfreed: Once your PR gets merged, I'll update this to use the correct method. Until then I'll leave it as-is so the doc tests don't fail.

src/lib.rs
patrickfreed: fixed

src/lib.rs
patrickfreed: So it appears it actually doesn't. I was going off [this comment](https://github.com/mongodb/bson-rust/blob/master/src/macros.rs#L181) in the macro itself, but as you can see in the line below it only works on things that are `Into<Bson>`. Doing a little digging, it looks like the comment itself was lifted from [`serde_json`'s implementation of `json!`](https://github.com/serde-rs/json/blob/master/src/macros.rs#L270) way [back when](https://github.com/mongodb/bson-rust/pull/83/files#r142045639). I've updated this section to reflect that.

As a fun side note, this made me realize why `bson!` sometimes required a `Bson` type annotation (was calling `std::convert::From` in the macro instead of `Bson::from`), so I've fixed that.

patrickfreed:
Should I update the repos readme to contain these examples?

saghm:
Sure, sounds good

README.md
```diff
@@ -20,53 +31,166 @@ This crate works with Cargo and can be found on
...
-    pub id: bson::oid::ObjectId,
-    pub name: String,
-    pub age: i32
-}
+let string = Bson::String("hello world".to_string());
+let int = Bson::I32(5);
+let array = Bson::Array(vec![Bson::I32(5), Bson::Boolean(false)]);
+
+let string: Bson = "hello world".into();
+let int: Bson = (5 as i32).into();
```
saghm, line 78: Slight nit: instead of casting, you can do `5i32` to explicitly create the value as an i32. This also lets you omit the parentheses

src/lib.rs
isabelatkinson: just merged

README.md
patrickfreed: done

src/lib.rs
patrickfreed: done

PR #177, opened 2020-05-28
RUST-445 Round trip decimal128 values

patrickfreed:
[RUST-445](https://jira.mongodb.org/browse/RUST-455)

This PR adds basic support for the BSON decimal128 type outside of the `"decimal128"` feature flag. It exposes the `Decimal128` struct but with no functionality, so it can only be created by deserializing existing documents (e.g. from the DB).

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -328,6 +328,10 @@ impl From<Bson> for Value {
 
 impl Bson {
     /// Converts the Bson value into its [relaxed extended JSON representation](https://docs.mongodb.com/manual/reference/mongodb-extended-json/).
+    ///
+    /// Note: extended json encoding for `Decimal128` values is not supported without the
+    /// "decimal128" feature flag. If this method is called on a case which contains a
+    /// `Decimal128` value, it will panic.
     pub fn into_relaxed_extjson(self) -> Value {
         match self {
             Bson::FloatingPoint(v) if v.is_nan() => {
@@ -428,6 +432,10 @@ impl Bson {
     }
 
     /// Converts the Bson value into its [canonical extended JSON representation](https://docs.mongodb.com/manual/reference/mongodb-extended-json/).
+    ///
+    /// Note: extended json encoding for `Decimal128` values is not supported without the
+    /// "decimal128" feature flag. If this method is called on a case which contains a
+    /// `Decimal128` value, it will panic.
     pub fn into_canonical_extjson(self) -> Value {
         match self {
             Bson::I32(i) => json!({ "$numberInt": i.to_string() }),
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -29,12 +29,11 @@ use std::{
 use chrono::{DateTime, Datelike, SecondsFormat, TimeZone, Utc};
 use serde_json::{json, Value};
 
-#[cfg(feature = "decimal128")]
-use crate::decimal128::Decimal128;
 pub use crate::document::Document;
 use crate::{
     oid::{self, ObjectId},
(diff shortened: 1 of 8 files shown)


src/bson.rs
```diff
@@ -408,6 +405,11 @@ impl Bson {
             Bson::Symbol(v) => json!({ "$symbol": v }),
             #[cfg(feature = "decimal128")]
             Bson::Decimal128(ref v) => json!({ "$numberDecimal": v.to_string() }),
+            #[cfg(not(feature = "decimal128"))]
+            Bson::Decimal128(_) => panic!(
```
patrickfreed, line 409: this is a bit tricky since we cannot return an error here, but we also cannot come up with a reasonable value to return. I opted to panic, but we could also fail more silently here by returning an incomplete value (e.g. `{ "$numberDecimal": null }` or just `null`).

src/decoder/mod.rs
```diff
@@ -98,6 +97,16 @@ fn read_i64<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<i64> {
     reader.read_i64::<LittleEndian>().map_err(From::from)
 }
 
+/// Placeholder decoder for `Decimal128`. Reads 128 bits and just stores them, does no validation or
+/// parsing.
+#[cfg(not(feature = "decimal128"))]
+#[inline]
+fn read_f128<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<Decimal128> {
+    let mut buf = vec![0u8; 128 / 8];
```
saghm, line 105: Since it's only 16 bytes, it's probably fine to just use an array (which will use the stack rather than the heap):

```rust
let mut buf: [u8; 128 / 8] = [0; 128 / 8];
```

src/bson.rs
```diff
@@ -73,7 +72,6 @@ pub enum Bson {
     /// Symbol (Deprecated)
     Symbol(String),
     /// [128-bit decimal floating point](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst)
-    #[cfg(feature = "decimal128")]
     Decimal128(Decimal128),
```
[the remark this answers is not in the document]
patrickfreed: All other drivers with a type for it seem to call it `Decimal128` (go / c / java / python / node).

src/bson.rs
patrickfreed: I added a note in the docstring for this method to call it out so users won't be too surprised if they run into it.

src/decimal128.rs
```diff
@@ -1,15 +1,28 @@
...
-    inner: d128,
+    #[cfg(not(feature = "decimal128"))]
+    /// BSON bytes containing the decimal128. Stored for round tripping.
+    pub(crate) bytes: Vec<u8>,
+
+    #[cfg(feature = "decimal128")]
+    inner: decimal::d128,
 }
 
+#[cfg(feature = "decimal128")]
```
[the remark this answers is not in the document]
patrickfreed: yeah it's pretty cool. Hopefully the docs will render with the feature thing properly like this though.

src/decoder/mod.rs
patrickfreed: done

src/bson.rs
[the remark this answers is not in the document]
saghm: Sounds good

src/decoder/mod.rs
saghm: Ah, I didn't even think of doing it in the struct itself too for some reason

PR #178, opened 2020-05-28
RUST-453 bson-rust is missing binary subtype 6

markbenvenuto:
bson-rust is missing the new binary subtype 6. This adds support for recognizing BSON with the new subtype. The BSON library is not responsible for encryption so there is not a need for more support.

Ran
cargo test
bash .evergreen/check-clippy.sh
bash .evergreen/check-rustfmt.sh

### DIFF
--- a/tests/modules/macros.rs
+++ b/tests/modules/macros.rs
@@ -30,6 +30,7 @@ fn standard_format() {
         "i64": -55,
         "timestamp": Bson::TimeStamp(TimeStamp { time: 0, increment: 229_999_444 }),
         "binary": Binary { subtype: BinarySubtype::Md5, bytes: "thingies".to_owned().into_bytes() },
+        "encrypted": Binary { subtype: BinarySubtype::Encrypted, bytes: "secret".to_owned().into_bytes() },
         "_id": id,
         "date": Bson::UtcDatetime(date),
     };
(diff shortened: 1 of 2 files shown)


tests/modules/macros.rs
```diff
@@ -38,9 +39,10 @@ fn standard_format() {
         "{{ float: 2.4, string: \"hello\", array: [\"testing\", 1, true, [1, 2]], doc: {{ fish: \
          \"in\", a: \"barrel\", !: 1 }}, bool: true, null: null, regexp: /s[ao]d/i, \
          with_wrapped_parens: -20, code: function(x) {{ return x._id; }}, i32: 12, i64: -55, \
-         timestamp: Timestamp(0, 229999444), binary: BinData(5, 0x{}), _id: ObjectId(\"{}\"), \
-         date: Date(\"{}\") }}",
+         timestamp: Timestamp(0, 229999444), binary: BinData(5, 0x{}), encrypted: BinData(6, \
```
patrickfreed, line 42: this diff has made me realize that I think the `Display` implementation for `Binary` is wrong. The `0x` prefix should be on the subtype not the base64 payload. From looking at the implementation, it looks like its manually adding `0x` rather than using the `{:#x}` hexadecimal formatter. While we're here, can you fix that as part of this change?

patrickfreed: This change is actually being made as part of [RUST-457](https://jira.mongodb.org/browse/RUST-457)

PR #179, opened 2020-05-29
RUST-447 Standardize API with other serde libraries

isabelatkinson:
This PR updates the bson API language to use "serialize" and "deserialize" in place of "encode" and "decode" respectively.

### DIFF
--- a/src/de/error.rs
+++ b/src/de/error.rs
@@ -7,7 +7,7 @@ use serde::de;
 /// Possible errors that can arise during decoding.
 #[derive(Debug)]
 #[non_exhaustive]
-pub enum DecoderError {
+pub enum Error {
     /// A [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html) encountered while deserializing.
     IoError(io::Error),
 
@@ -45,24 +45,24 @@ pub enum DecoderError {
     },
 }
 
-impl From<io::Error> for DecoderError {
-    fn from(err: io::Error) -> DecoderError {
-        DecoderError::IoError(err)
+impl From<io::Error> for Error {
+    fn from(err: io::Error) -> Error {
+        Error::IoError(err)
     }
 }
 
-impl From<string::FromUtf8Error> for DecoderError {
-    fn from(err: string::FromUtf8Error) -> DecoderError {
-        DecoderError::FromUtf8Error(err)
+impl From<string::FromUtf8Error> for Error {
+    fn from(err: string::FromUtf8Error) -> Error {
+        Error::FromUtf8Error(err)
     }
 }
 
-impl fmt::Display for DecoderError {
+impl fmt::Display for Error {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            DecoderError::IoError(ref inner) => inner.fmt(fmt),
-            DecoderError::FromUtf8Error(ref inner) => inner.fmt(fmt),
-            DecoderError::UnrecognizedDocumentElementType {
+            Error::IoError(ref inner) => inner.fmt(fmt),
+            Error::FromUtf8Error(ref inner) => inner.fmt(fmt),
+            Error::UnrecognizedDocumentElementType {
                 ref key,
                 element_type,
             } => write!(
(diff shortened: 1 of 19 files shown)


src/de/serde.rs
```diff
@@ -4,7 +4,7 @@ use serde::de::{
     self,
     Deserialize,
     DeserializeSeed,
-    Deserializer,
+    Deserializer as SerdeDeserializer,
```
isabelatkinson, line 7: I think this rename needs to occur because the original name conflicts with the newly-named `Deserialize` struct, but the import needs to remain as-is for [this line](https://github.com/mongodb/bson-rust/blob/ed1740acde6b1f21007af8872d275404e6f415e1/src/decoder/serde.rs#L494), but let me know if there's a better workaround.

src/de/mod.rs
```diff
@@ -44,16 +44,16 @@ use crate::{
     Decimal128,
 };
 
-use ::serde::de::{Deserialize, Error};
+use ::serde::de::{self, Error as SerdeError};
```
isabelatkinson, line 47: Same conflict as other rename.

src/document.rs
```diff
@@ -510,11 +510,11 @@ impl Document {
         }
     }
 
-    /// Attempts to encode the `Document` into a byte stream.
-    pub fn encode<W: Write + ?Sized>(&self, writer: &mut W) -> EncoderResult<()> {
+    /// Attempts to serialize the `Document` into a byte stream.
+    pub fn serialize_doc<W: Write + ?Sized>(&self, writer: &mut W) -> crate::ser::Result<()> {
```
isabelatkinson, line 514: Since `Document` implements the `Serialize` trait [here](https://github.com/mongodb/bson-rust/blob/ed1740acde6b1f21007af8872d275404e6f415e1/src/encoder/serde.rs#L45), and calls `serialize` [here](https://github.com/mongodb/bson-rust/blob/ed1740acde6b1f21007af8872d275404e6f415e1/src/encoder/serde.rs#L69), the compiler wasn't happy with this method being named `serialize`. (I had no idea Rust doesn't allow method overloading, interesting.)

src/de/error.rs
```diff
@@ -7,7 +7,7 @@ use serde::de;
 /// Possible errors that can arise during decoding.
 #[derive(Debug)]
 #[non_exhaustive]
-pub enum DecoderError {
+pub enum Error {
```
saghm, line 10: Since we have a separate error type in the `de` module than at the top level (as opposed to something like `serde_json`, which has only one error type in the entire crate`, I think we still might want to have a prefix in the name of this type (and for the Result type below); it's a bit friendlier to users who need to import multiple Error types from our crate if we give them unique names.

src/document.rs
saghm: Good catch, I hadn't thought of this conflict!. Technically you _could_ have two different methods with the same name if one of them is from a trait, but it's not traditional overloading because there's no disambiguation at the call site based on the parameters; you'd only have access to one of methods for a given variable based on what type it is. Here's an example of that: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=996a1b2bab9c7a8602de0098de00012c

I think I might prefer the name `serialize_to` here; when calling it, I think it would read well, e.g. `doc.serialize_to(&mut writer)?`.

src/document.rs
```diff
@@ -526,8 +526,8 @@ impl Document {
         Ok(())
     }
 
-    /// Attempts to decode a `Document` from a byte stream.
-    pub fn decode<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<Document> {
+    /// Attempts to deserialize a `Document` from a byte stream.
+    pub fn deserialize<R: Read + ?Sized>(reader: &mut R) -> crate::de::Result<Document> {
```
saghm, line 530: If you go with my `serialize_to` suggestion above, I think this should be renamed to `deserialize_from` for parity.

src/ser/error.rs
```diff
@@ -7,7 +7,7 @@ use crate::bson::Bson;
 /// Possible errors that can arise during encoding.
 #[derive(Debug)]
 #[non_exhaustive]
-pub enum EncoderError {
+pub enum Error {
```
saghm, line 10: Ditto above suggestion to prefer the name of the Error/Result types in this module

src/de/serde.rs
saghm: Since you don't need to use the name `SerdeDeserializer` at all, I think you can actually use `Deserializer as _`, which will include it for the purposes of values that implement that trait but won't put any name into scope.

src/document.rs
isabelatkinson: I agree, less redundant than having `doc` in the name

src/de/error.rs
isabelatkinson: @patrickfreed is there precedence in other serde libraries for this naming style when there's more than one error/result defined? I agree that it's a little less user friendly to use something like `crate::de::Error` rather than `DeserializerError`.

patrickfreed: I think it's pretty standard to rely on the module path for these kinds of things. All the serialization libraries that I can find which have separate errors for serialization / deserialization simply name each `Error` ([`toml`](https://docs.rs/toml/0.5.6/toml/de/index.html), [`xml`](https://docs.rs/xml-rs/0.8.3/xml/reader/index.html), [`serde_urlencoded`](https://docs.rs/serde_urlencoded/0.6.1/serde_urlencoded/de/struct.Error.html)), and `serde` itself also follows this pattern with its [`ser::Error`](https://docs.rs/serde/1.0.110/serde/ser/trait.Error.html) and [`de::Error`](https://docs.rs/serde/1.0.110/serde/de/trait.Error.html) traits. I think we should keep to this pattern for consistency with other libraries.

src/document.rs
patrickfreed: I like `serialize_to`, but we might also want to consider something like `to_writer` and `from_reader` like in [`serde_json`](https://docs.rs/serde_json/1.0.53/serde_json/fn.to_writer.html) and [`serde-xml-rs`](https://docs.rs/serde-xml-rs/0.4.0/serde_xml_rs/ser/fn.to_writer.html).

src/de/error.rs
isabelatkinson: That makes sense, I think consistency with `serde` would be good. @saghm what do you think?

saghm: I guess it makes sense to be consistent with our `oid::Error`. I'm not sure what you're referring to with the `xml-rs` crate, though; as it doesn't seem to have `ser` or `de` modules; do you mean [`serde_xml`](https://docs.rs/serde_xml/)?

src/document.rs
saghm: I'm fine with `from_reader` and `to_writer`.

isabelatkinson: I'm fine with both, I'll leave as-is with `serialize_to` unless anyone feels super strongly either way.

saghm: You can change it to Patrick's suggestions; sorry if I was unclear

isabelatkinson: oops sorry I didn't see that first comment! I had this window sitting open for a while and didn't realize you need to refresh to see new comments. will do

saghm: No worries; github does require a lot of refreshes that it should probably trigger on its own

src/de/error.rs
patrickfreed: The `xml` crate I was referring to appears to actually just be a parsing library, not a serialization one. It has `reader` and `writer` modules with their own `Error`s which are somewhat analogous to `de` and `ser`. `serde_xml` appears to be the actual corresponding serialization crate for xml.

PR #180, opened 2020-05-29
RUST-457 Fix Display implementation for Binary type

isabelatkinson:


patrickfreed:
change looks good, just need to rebase off of the subtype 6 commit to fix that one up too.

PR #181, opened 2020-06-01
RUST-454 Implement error case BSON corpus tests

patrickfreed:
[RUST-454](https://jira.mongodb.org/browse/RUST-454)

This PR implements the "decode errors" and "parse errors" sections of the BSON corpus spec tests.

As part of implementing the parse tests, there ended up being far more changes (both in implementation and API) than I would have liked, but I believe they were all necessary. The changes stem from the fact that our current extJSON parser would not fail if extJSON wasn't formatted properly--it would just parse it as a document. The spec's indicate we should actually reject such inputs, which required adding failability to all the extJSON parsing mechanisms (`From<Value> for Bson => TryFrom<Value> for Bson`).

To implement these changes, I started by trying to modify `from_extended_document` to shake the boat as little as possible, but I quickly realized that doing the error checking there would require a ton of boilerplate. I instead opted to leverage `serde` as much as possible by defining models for the various extJSON representations of things, while still leaving `from_extended_document` around so as to not break the `serde` implementation.

From an API perspective, I needed a new error type for extJSON parsing errors. To that end, I added an `extjson` and `extjson::de` module to house that type. In the future, I think we could add further generic functions for extJSON deserialization/serialization in there too.

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -1,6 +1,7 @@
 //! A BSON document represented as an associative HashMap with insertion ordering.
 
 use std::{
+    convert::{TryFrom, TryInto},
     error,
     fmt::{self, Debug, Display, Formatter},
     io::{Read, Write},
@@ -100,6 +101,29 @@ impl Debug for Document {
         write!(f, "Document({:?})", self.inner)
     }
 }
+
+impl Document {
+    pub(crate) fn from_ext_json(
+        obj: serde_json::Map<String, serde_json::Value>,
+    ) -> DecoderResult<Self> {
+        Ok(obj
+            .into_iter()
+            .map(|(k, v)| -> DecoderResult<(String, Bson)> {
+                let value: Bson = v.clone().try_into()?;
+                Ok((k.clone(), value))
+            })
+            .collect::<DecoderResult<Vec<(String, Bson)>>>()?
+            .into_iter()
+            .collect())
+    }
+}
+
+// impl TryFrom<serde_json::Value> for Document {
+//     type Error = crate::DecoderError;
+//     fn try_from(value: serde_json::Value) -> Result<Self, Self::Error> {
+//         unimplemented!()
+//     }
+// }
 
 /// An iterator over Document entries.
 pub struct DocumentIntoIterator {
--- a/src/document.rs
+++ b/src/document.rs
@@ -1,7 +1,6 @@
 //! A BSON document represented as an associative HashMap with insertion ordering.
 
 use std::{
-    convert::{TryFrom, TryInto},
     error,
     fmt::{self, Debug, Display, Formatter},
     io::{Read, Write},
@@ -100,22 +99,6 @@ impl Display for Document {
 impl Debug for Document {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         write!(f, "Document({:?})", self.inner)
-    }
-}
-
-impl TryFrom<serde_json::Map<String, serde_json::Value>> for Document {
-    type Error = extjson::de::Error;
-
-    fn try_from(obj: serde_json::Map<String, serde_json::Value>) -> extjson::de::Result<Self> {
-        Ok(obj
-            .into_iter()
-            .map(|(k, v)| -> extjson::de::Result<(String, Bson)> {
-                let value: Bson = v.try_into()?;
-                Ok((k, value))
-            })
-            .collect::<extjson::de::Result<Vec<(String, Bson)>>>()?
-            .into_iter()
-            .collect())
     }
 }
 
--- a/src/document.rs
+++ b/src/document.rs
@@ -21,7 +21,7 @@ use serde::de::{self, Error, MapAccess, Visitor};
 use crate::decimal128::Decimal128;
 use crate::{
     bson::{Array, Binary, Bson, Timestamp},
-    de::{deserialize_bson_kvp, read_i32, MIN_BSON_DOCUMENT_SIZE},
+    de::{deserialize_bson_kvp, ensure_read_exactly, read_i32, MIN_BSON_DOCUMENT_SIZE},
     extjson,
     oid::ObjectId,
     ser::{serialize_bson, write_i32},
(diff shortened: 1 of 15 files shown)


src/extjson/de.rs
```diff
@@ -0,0 +1,230 @@
+//! Deserializing [MongoDB Extended JSON v2](https://docs.mongodb.com/manual/reference/mongodb-extended-json/)
```
patrickfreed, line 1: This module is largely empty for the moment, but in the future I imagine all the various `from_value` / `from_reader` / etc. functions for extjson could live here. Likewise, we could introduce a `ser` module that has the counterparts to those.

e.g.
```rust
pub(crate) fn from_value<T>(ext_json: serde_json::Value) -> Result<T>
where
    T: DeserializeOwned,
{
    let bson: crate::Bson = ext_json.try_into()?;
    let result = crate::from_bson(bson)?;
    Ok(result)
}
```

src/extjson/de.rs
```diff
@@ -0,0 +1,230 @@
...
+impl From<oid::Error> for Error {
+    fn from(err: oid::Error) -> Self {
+        Self::InvalidObjectId(err)
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// This converts from the input JSON object as if it were [MongoDB Extended JSON v2](https://docs.mongodb.com/manual/reference/mongodb-extended-json/).
+impl TryFrom<serde_json::Map<String, serde_json::Value>> for Bson {
```
patrickfreed, line 81: this method could be simplified greatly through clever uses of macros and traits or something, but for the time being I just wanted to keep it simple, with the idea being we can restructure it after we ship 1.0

src/extjson/mod.rs
```diff
@@ -0,0 +1,81 @@
+//! Deserialization and serialization of [MongoDB Extended JSON v2](https://docs.mongodb.com/manual/reference/mongodb-extended-json/)
```
patrickfreed, line 1: this module is also largely empty, but it could export the various functions we eventually define in the `ser` and `de` submodules like we do for BSON.
e.g.
```rust
// in `extjson/mod.rs`
pub use de::from_value;
pub use ser::to_value;

// in users' code
let ext_json = extjson::to_value(5).unwrap();
let x: i32 = extjson::from_value(json!({"$numberInt: "5" })).unwrap();
```
The way I see it, extJSON is basically its own serialization format, so it makes sense to have a similar module hierarchy and functionality for it.

tests/spec/corpus.rs
```diff
@@ -247,6 +250,40 @@ fn run_test(test: TestFile) {
             );
         }
     }
+
+    for decode_error in test.decode_errors {
+        // No meaningful definition of "byte count" for an arbitrary reader.
```
patrickfreed, line 255: Perhaps I'm misunderstanding this test, but I don't see why there can't be other data after the section identified as being the document.

src/decoder/mod.rs
```diff
@@ -120,20 +130,29 @@ fn read_f128<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<Decimal128> {
 
 fn decode_array<R: Read + ?Sized>(reader: &mut R, utf8_lossy: bool) -> DecoderResult<Array> {
     let mut arr = Array::new();
+    let length = read_i32(reader)?;
 
-    // disregard the length: using Read::take causes infinite type recursion
-    read_i32(reader)?;
+    let mut buf = vec![0u8; (length as usize) - 4];
```
saghm, line 135: I assume we can't use an array here since the size is dynamic?

src/decoder/mod.rs
```diff
@@ -120,20 +130,29 @@ fn read_f128<R: Read + ?Sized>(reader: &mut R) -> DecoderResult<Decimal128> {
...
+    reader.read_exact(&mut buf)?;
 
+    let mut cursor = std::io::Cursor::new(buf);
```
saghm, line 138: This cursor loop seems like it's used in very similar ways a couple times below. Do you think it's worth abstracting in some way?

src/extjson/models.rs
```diff
@@ -0,0 +1,371 @@
...
+impl DateTime {
+    pub(crate) fn parse(self) -> extjson::de::Result<crate::DateTime> {
+        match self.body {
+            DateTimeBody::Canonical(date) => {
+                let date = date.parse()?;
+
+                let mut num_secs = date / 1000;
+                let mut num_millis = date % 1000;
+
+                // The chrono API only lets us create a DateTime with an i64 number of seconds
```
saghm, line 225: Nice to see that my giant comment survived the refactor

src/decoder/mod.rs
patrickfreed: done

src/document.rs
```diff
@@ -542,24 +541,19 @@ impl Document {
         let mut buf = vec![0u8; (length as usize) - 4];
         reader.read_exact(&mut buf)?;
 
-        let mut cursor = std::io::Cursor::new(buf);
-        loop {
-            let tag = cursor.read_u8()?;
+        ensure_read_exactly(buf, "document length longer than contents", |cursor| {
```
saghm, line 544: We can probably move the reading of the bytes into the buffer above into the function too, right?

tests/spec/corpus.rs
saghm: Since this is in the top-level file, I think the idea is that if that the user tries to decode a single document and then provides more bytes than the document indicates, it should fail. I'm not really sure how we could represent that in our tests though, since our readers are lazy.

src/document.rs
patrickfreed: good idea, done

tests/spec/corpus.rs
patrickfreed: Is it okay to just continuing to skip this test then?

saghm: I'm fine with skipping it

PR #182, opened 2020-06-01
RUST-390 Implement ordering for Timestamp

isabelatkinson:


saghm:
Could you add a couple simple tests for this behavior? I honestly can't remember off the top of my head how deriving Ord prioritizes the fields in a struct, so I think it makes sense to add a sanity check that the time field takes precedence over the increment field.

isabelatkinson:
Will do! The [docs](https://doc.rust-lang.org/std/cmp/trait.Ord.html) indicate that it prioritizes top-down so time should precede increment but I agree it wouldn't hurt to sanity check.

PR #186, opened 2020-06-04
fix: Don't use the type for constants; use std

### LINKED ISSUE #185: bson 0.15 build error!!!

fjah:
There were multiple cases where a primitive type was being used as a module, like `f64::NAN`. This was fixed by using the `std` module, like `std::f64::NAN`.

fixes #185

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -636,9 +636,9 @@ impl Bson {
             }
 
             ["$numberDouble"] => match doc.get_str("$numberDouble") {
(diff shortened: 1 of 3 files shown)


saghm:
It looks like the rustfmt lint is failing; the instructions for setting up this test locally are [here](https://github.com/mongodb/bson-rust#linter-tests).

fjah:
Oh - apologies. Should be fixed.

saghm:
No worries!

saghm:
Thanks for the pull request!

PR #188, opened 2020-06-05
Bump dependencies for hex and md5 crates

jpgneves:
- hex -> 0.4.2
- md5 -> 0.7.0

The dependency updates should be innocuous and bring us to the latest version of these crates.


patrickfreed:
Thanks for your contribution!

PR #193, opened 2020-06-23
RUST-475 adding getter for timestamp from ObjectId

utkmehta:
added a getter for timestamp and a test for the same

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -13,6 +13,8 @@ use hex::{self, FromHexError};
 
 use rand::{thread_rng, Rng};
 
+use chrono::Utc;
+
 const TIMESTAMP_SIZE: usize = 4;
 const PROCESS_ID_SIZE: usize = 5;
 const COUNTER_SIZE: usize = 3;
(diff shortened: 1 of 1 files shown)


src/oid.rs
```diff
@@ -114,6 +114,11 @@ impl ObjectId {
         }
     }
 
+    /// Retrieves the timestamp (seconds since epoch) from an ObjectId.
+    pub fn timestamp(&self) -> u32 {
```
patrickfreed, line 118: I think if it would be better to return a `DateTime` here instead of a number of seconds. This would be in line with the other drivers (Go / Ruby / Node / Python). Java opts to have methods for both, but I think that would probably be unnecessary given that a user could easily [get the number of seconds from the `DateTime`](https://docs.rs/chrono/0.4.11/chrono/struct.DateTime.html#method.timestamp).

cc @saghm

saghm: Interesting; the version of this method we had before just returned the number of seconds, but I wouldn't be opposed to returning a DateTime here

src/oid.rs
```diff
@@ -204,16 +214,38 @@ fn count_generated_is_big_endian() {
...
-fn test_debug() {
-    let id = ObjectId::with_string("53e37d08776f724e42000000").unwrap();
+    #[test]
+    fn test_debug() {
+        let id = super::ObjectId::with_string("53e37d08776f724e42000000").unwrap();
 
-    assert_eq!(format!("{:?}", id), "ObjectId(53e37d08776f724e42000000)")
+        assert_eq!(format!("{:?}", id), "ObjectId(53e37d08776f724e42000000)")
+    }
+    use chrono::{offset::TimeZone, Utc};
```
saghm, line 232: Small nit: could you move this import to the top of the test module and separate the test by a single blank line? Something like this:

```rust
#[cfg(test)]
mod test {
    use chrono::{offset::TimeZone, Utc};

    #[test]
    fn test_display() {
       ...
    }

    #[test]
    fn test_debug() {
       ...
    }

    #[test]
    fn test_timestamp() {
       ...
    }
}
```

src/oid.rs
utkmehta: fixed

src/oid.rs
utkmehta: fixed

saghm: Thanks!

PR #194, opened 2020-06-24
RUST-413 Implementing ObjectId spec tests

utkmehta:
1. Timestamp field already tested.
2. Counter overflow tested by checking if 0xFFFFFF overflows to 0
3. Fork doesn't have to be tested in Rust

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -1,5 +1,4 @@
 //! ObjectId
-
 use std::{
     error,
     fmt,
@@ -13,6 +12,7 @@ use hex::{self, FromHexError};
 
 use rand::{thread_rng, Rng};
 
+use crate::tests::LOCK;
 use chrono::Utc;
 
 const TIMESTAMP_SIZE: usize = 4;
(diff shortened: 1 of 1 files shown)


src/oid.rs
```diff
@@ -250,4 +251,11 @@ mod test {
         // "Feb 7th, 2106 06:28:15 UTC"
         assert_eq!(Utc.ymd(2106, 2, 7).and_hms(6, 28, 15), id.timestamp());
     }
+
+    #[test]
+    fn test_counter_overflow() {
```
patrickfreed, line 256: This test is supposed to be verifying that the counter used for ObjectID generation overflows after an ObjectID is generated while the counter is the maximum unsigned 24-bit integer, but as-written it is just instantiating a single ObjectID directly from bytes and then reading from it.

What we need from this test is to set the counter to `std::u24::MAX` and to generate two`ObjectId`s via `ObjectId::new`, verifying the first one's counter is `std::u24::MAX` and the next one's is 0 by reading the last 3 bytes of each as big endian `u24`s. Note that the test is currently reading the first 3 bytes as little endian `u24`s instead, which will need to be fixed. The constant `COUNTER_OFFSET` can be used to index into the buffer until the last 3 bytes (e.g. `oid.bytes[COUNTER_OFFSET..]`) and `OID_COUNTER` can be set directly, but the test will need to be moved out of the test module in order to access those two. For an example, see the `count_generated_is_big_endian` test in `oid.rs`.

PR #195, opened 2020-07-02
RUST-484 Add a test lock and move integration tests

utkmehta:
This PR adds a test lock because of concurrency issues (multiple tests reading/writing from/to counter in an object and all running on the same thread). To be able to use the same lock in src/oid.rs and the tests in src/tests, the tests module had to be moved to src/tests


Cargo.toml
```diff
@@ -37,6 +37,7 @@ hex = "0.3"
 md5 = "0.6"
 decimal = { version = "2.0.4", default_features = false, optional = true }
 base64 = "0.12.1"
+lazy_static = "1.4.0"
```
saghm, line 40: Since this is only needed for the tests, we can put it under `dev-dependencies` rather than `dependencies` so that users of the library don't need to pull it in.

src/tests/mod.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(clippy::cognitive_complexity)]
```
saghm, line 1: Since this is in the `src` directory now, which has this attribute in `lib.rs`, we can remove it from here

src/tests/modules/lock.rs
```diff
@@ -0,0 +1,20 @@
...
+    inner: RwLock<()>,
+}
+
+impl TestLock {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn run_concurrently(&self) -> LockResult<RwLockReadGuard<'_, ()>> {
+        self.inner.read()
```
saghm, line 14: We should `unwrap` the calls to `read` and `write` here so that the tests fail loudly if we can't acquire the lock

Cargo.toml
utkmehta: makes sense. fixed

src/tests/mod.rs
utkmehta: fixed

src/tests/modules/lock.rs
utkmehta: makes sense. fixed

src/tests/modules/ser.rs
```diff
@@ -163,6 +163,7 @@ fn int64() {
 
 #[test]
 fn oid() {
+    let _guard = LOCK.run_concurrently();
```
patrickfreed, line 166: The serde tests are run separately from the `u2i` and regular tests, so this lone concurrent assurance won't do anything on its own. For this reason, I think it's probably fine if we just don't bother acquiring the lock in the serde tests at all.

utkmehta: makes sense. fixed

src/tests/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod modules;
```
patrickfreed, line 1: Every test within `src/tests/*` needs to acquire the lock, either concurrently or exclusively, in order to guarantee that the ones which acquire it exclusively are actually executing alone. Currently, there are several tests in these modules that do not acquire the lock.

utkmehta: Why would we need to acquire a lock if the test does not interfere with the counter field? I added locks to all tests that create a new object using Object::New() or interfere with counter otherwise.

patrickfreed: If those tests do not acquire the lock to run concurrently, then the tests that acquire the lock exclusively will be running concurrently with them, which is what is supposed to be avoided by the test lock. In practice, it may not have an effect because you manually vetted those tests, but there's no telling that in the future they won't make a call to `ObjectId::new` or that there are other concurrent conditions to take into account. For those reasons, we need to ensure full usage of the lock across all tests, including the concurrent ones.

src/tests/modules/ser.rs
patrickfreed: So I just noticed that my comment was wrong in some ways. The serde tests in `ser/tests/modules/ser.rs` _do_ need the lock (as was mentioned in my comment below), but the tests in `serde-tests/*` are the ones that don't need it, since they're run separately.

The `u2i` tests also need the locks, since they are run via the same `cargo test` invocation as the rest of the regular tests (see `run-tests-u2i.sh`).

src/tests/mod.rs
utkmehta: Thanks, fixed!

src/tests/modules/bson.rs
```diff
@@ -49,19 +51,23 @@ fn to_json() {
...
+    let _guard = LOCK.run_concurrently();
     let doc1 = Document::default();
     assert_eq!(doc1.keys().count(), 0);
     assert_eq!(doc1, Document::new());
 }
 
 #[test]
 fn from_impls() {
+    let _guard = LOCK.run_concurrently();
+    let _guard = LOCK.run_concurrently();
```
isabelatkinson, line 70: Does this need to happen twice?

src/tests/modules/ordered.rs
```diff
@@ -48,17 +51,16 @@ fn ordered_insert_shorthand() {
 
 #[cfg(feature = "decimal128")]
 fn test_decimal128(doc: &mut Document) {
```
isabelatkinson, line 53: I think `#[test]` needs to be added above this test -- clippy is complaining that it doesn't get used.

src/tests/mod.rs
patrickfreed: Looks good, I think just a single test still needs the lock (`src/tests/spec/corpus.rs`).

utkmehta: added!

src/tests/modules/bson.rs
utkmehta: oof my bad. fixed!

src/tests/modules/ordered.rs
utkmehta: I think clippy was complaining since I removed the call to this test. Added it back with the decimal128 feature enabled. So it shouldn't complain now

PR #196, opened 2020-07-02
RUST-413 Implementing ObjectId spec tests

utkmehta:
1. Timestamp already tested.
2. counter overflow tested by creating an objectID with MAX counter and making sure the counter is MAX. Then created another objectID which incremented the counter to MAX + 1 and made sure it is reset to 0. Found a bug in the code. Fixed that.
3. No fork in Rust.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -16,6 +16,7 @@ use hex::{self, FromHexError};
 use rand::{thread_rng, Rng};
 
 use chrono::Utc;
+use lazy_static::lazy_static;
 
 const TIMESTAMP_SIZE: usize = 4;
 const PROCESS_ID_SIZE: usize = 5;
@@ -27,7 +28,9 @@ const COUNTER_OFFSET: usize = PROCESS_ID_OFFSET + PROCESS_ID_SIZE;
 
 const MAX_U24: usize = 0xFF_FFFF;
 
-static OID_COUNTER: AtomicUsize = AtomicUsize::new(0);
+lazy_static! {
+    static ref OID_COUNTER: AtomicUsize = AtomicUsize::new(thread_rng().gen_range(0, MAX_U24 + 1));
+}
 
 /// Errors that can occur during OID construction and generation.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/oid.rs
```diff
@@ -168,7 +168,7 @@ impl ObjectId {
         // Static mutexes are currently unstable; once they have been
         // stabilized, one should be used to access OID_COUNTER and
         // perform multiple operations atomically.
-        let u = u_counter % MAX_U24;
+        let u = u_counter % (MAX_U24 + 1);
```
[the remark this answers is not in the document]
utkmehta: We want counter to overflow to 0 when it reaches MAX+1, not MAX

src/oid.rs
```diff
@@ -168,7 +168,7 @@ impl ObjectId {
         // Static mutexes are currently unstable; once they have been
```
patrickfreed, line 168: The counter is stored as a `usize`, so a second "overflow event" will occur when that reaches its max. Also, from looking at the implementation, once the counter reaches usize's max, it will be randomly generated again and not overflow properly. To verify this, I'd add another test case that sets the counter to `usize`'s max and see what happens. It should fail I think, and so `gen_count` will need to be updated to fix it.

utkmehta: makes sense. fixed by adding a boolean variable and making sure counter resets to 0 in case of usize_max overflow (instead of random generation)

patrickfreed: The atomic boolean approach is susceptible to race conditions. In order to maintain consistency between the two different variables you'd need a lock. Unfortunately, we can't use locks either (as the comment in the code suggests) since it would needlessly inhibit performance and prevent `ObjectId::new` from being used in non-blocking contexts.

To do this without locks, we could use `lazy_static` to ensure the random generation only happens the first time, and all subsequent rollovers will just happen normally via addition without the usage of a separate atomic variable. As part of that, you can just remove the comment about adding locks in the future.

src/oid.rs
```diff
@@ -157,22 +159,16 @@ impl ObjectId {
...
         // Init oid counter
-        if OID_COUNTER.load(Ordering::SeqCst) == 0 {
-            let start = thread_rng().gen_range(0, MAX_U24 + 1);
-            OID_COUNTER.store(start, Ordering::SeqCst);
-        }
 
         let u_counter = OID_COUNTER.fetch_add(1, Ordering::SeqCst);
 
         // Mod result instead of OID_COUNTER to prevent threading issues.
         // Static mutexes are currently unstable; once they have been
```
patrickfreed, line 166: Oh can you remove the comment from "Static mutexes..." and onward? It's out of date now that we're using `lazy_static`.

utkmehta: fixed

PR #198, opened 2020-07-12
Fix code snippet in README

jduan:
The code in the README doesn't work because of the breaking change in 0.8.0. This PR fixes the code so causes less confusion. 

I tested the change by running the code in a small program.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,6 @@ e.g.:
 #[derive(Serialize, Deserialize)]
 struct Person {
     name: String,
-    // Unsigned integer types are not supported. See the "Breaking Changes" section below.
     age: i32,
     phones: Vec<String>,
 }
@@ -240,4 +239,3 @@ bash .evergreen/check-clippy.sh && bash .evergreen/check-rustfmt.sh
 
 ## Continuous Integration
 Commits to master are run automatically on [evergreen](https://evergreen.mongodb.com/waterfall/rust-bson).
-
--- a/README.md
+++ b/README.md
@@ -154,7 +154,8 @@ e.g.:
 #[derive(Serialize, Deserialize)]
 struct Person {
     name: String,
-    age: u8,
+    // Unsigned integer types are not supported. See the "Breaking Changes" section below.
+    age: i32,
     phones: Vec<String>,
 }
 
--- a/README.md
+++ b/README.md
@@ -159,29 +159,27 @@ struct Person {
     phones: Vec<String>,
 }
 
-fn typed_example() {
-    // Some BSON input data as a `Bson`.
-    let bson_data: Bson = bson!({
-        "name": "John Doe",
-        "age": 43,
-        "phones": [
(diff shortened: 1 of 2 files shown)


jduan:
The failing CI check `evergreen — patch must be manually authorized` has a link that leads to a login page. I don't know what to do there.

saghm:
Sorry about the confusion; evergreen patches have to be authorized by one of us, since they run on company resources. Github unfortunately shows the "pending authorization" status as a test failure. I've authorized the patch to run now.

patrickfreed:
Thanks for catching this! Could you update the same example used in the documentation in `lib.rs` as well?

src/lib.rs
```diff
@@ -177,6 +177,8 @@
 //!     // Get a serialized version of the input data as a `Bson`.
 //!     let redacted_bson = bson::to_bson(&person).unwrap();
 //! }
+//!
+//! typed_example();
```
jduan, line 181: @patrickfreed I fixed the doc test. I think this line is needed, otherwise the code won't be exercised. Let me know if you see a problem. Thanks!

patrickfreed: Oh good point. In fact, I think this example should probably just not be included in a `fn` at all, since I don't think it really adds anything. So could you just remove the `typed_example` function wrapping here and in the README?

saghm: Are these examples getting run, or just compiled? I thought they were just compiled by default

patrickfreed: They're wrapped in an implicit `fn main() {}`, compiled, and then run. If they don't panic, they pass.

ref: https://doc.rust-lang.org/rustdoc/documentation-tests.html#passing-or-failing-a-doctest

jduan: That's a good idea. I've removed the wrapping function in 0885ae0.

saghm: Hmm, why do the driver doc tests work when no database is running then?

saghm: Oh wait, for the same reason as here; we don't call the functions that we define in the examples

jduan: @saghm I think fixing the other code in doc tests is beyond the scope of this PR. Does this PR look good for merging?

README.md
```diff
@@ -154,33 +154,32 @@ e.g.:
 #[derive(Serialize, Deserialize)]
 struct Person {
     name: String,
-    age: u8,
+    // Unsigned integer types are not supported. See the "Breaking Changes" section below.
```
saghm, line 157: I'm not sure we need to have a comment here talking about unsigned integers

jduan, line 157:
```suggestion
```

jduan: I've removed it in 7b0bcb3.

jduan:
@patrickfreed can you merge this PR? I don't have permissions.

patrickfreed:
@jduan done. Thanks for your contribution!

PR #203, opened 2020-07-23
Exclude unneeded files from built package

darnuria:
By default cargo publish all the content versionned by git.
following insight given by cargo diet I used an explicit approach
to exclude test dataset etc. from the final package.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,16 @@ homepage = "https://github.com/mongodb/bson-rust"
 documentation = "https://docs.rs/crate/bson"
 edition = "2018"
 
+# By default cargo include everything git include
+# cargo diet can help to manage what's not useful.
+include = [
+    "Cargo.toml",
+    "src/**/*",
+    "LICENSE",
+    "README.md",
+    "!**/tests/**/*"
+]
+
 [features]
 # no features by default
 default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,12 +15,16 @@ edition = "2018"
 
 # By default cargo include everything git include
 # cargo diet can help to manage what's not useful.
-include = [
-    "Cargo.toml",
-    "src/**/*",
-    "LICENSE",
-    "README.md",
-    "!**/tests/**/*"
+exclude = [
+    "etc/**",
+    "examples/**",
+    "fuzz/**",
+    "serde-tests/**",
+    "src/tests/**",
+    "rustfmt.toml",
+    ".travis.yml",
+    ".evergreen/**",
+    ".gitignore"
 ]
 
 [features]


Cargo.toml
```diff
@@ -13,6 +13,16 @@ homepage = "https://github.com/mongodb/bson-rust"
 documentation = "https://docs.rs/crate/bson"
 edition = "2018"
 
+# By default cargo include everything git include
+# cargo diet can help to manage what's not useful.
+include = [
+    "Cargo.toml",
+    "src/**/*",
```
saghm, line 20: Will this end up adding `src/test` even though it's ignored in `!**/tests/*`? I'm not sure what `include` does in the case of conflicting instructions

darnuria: I was wondering about it it's why I published it as a draft I am investigating it.

saghm: Oops, sorry, I missed that this was still a draft. Sorry for jumping the gun!

darnuria:
Seems like the files are included, look like a bug/edge case in the reporting of `cargo-diet` I fill a bug in the repository of the project because, the line suggested by `cargo diet`do not have the same semantics in cargo itself than in the size analisys of `cargo diet`.

I will push soon something to avoid test folder in each sub-folder or at least in src it seems to work. (Need some test to be sure)

Cargo.toml
saghm: I think it might be easier to just use `exclude` instead of `include`: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

Cargo.toml
```diff
@@ -13,6 +13,16 @@ homepage = "https://github.com/mongodb/bson-rust"
...
+    "LICENSE",
+    "README.md",
+    "!**/tests/**/*"
```
darnuria, line 23: This line seems to do the trick.

darnuria:
> I think it might be easier to just use exclude instead of include: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

Just made a commit with explicit exclusion if it's okay, I can rebase it, if it not I can change the Pull-request.

## Quick analysis of the size of the package

I checked the package size by doing:
```
cargo clean
cargo publish --dry-run
ls -al ./target/package
```
Results:
```
# Before 
78972 juil. 24 23:46 bson-1.0.0.crate
# After 
43993 juil. 24 23:33 bson-1.0.0.crate
```

Something like 34979 bytes saved!

Sorry for my English I am a not an English native writer and also thanks for your patience! :)

darnuria:
There was a effectively a bug in `crago-diet`about the interpretation of globing in cargo manifest format: https://github.com/the-lean-crate/cargo-diet/issues/6 seems solved.

saghm:
Any chance you'd be able to rebase this branch with master? I think that should fix the lint failures, and then this should be review for the rest of the team to review!

darnuria:
Hi, sure thanks for the follow-up!

I rebased it onto master, but I didn't squashed the two commit to keep information while reviewing of the first approach but if the exclusion one is selected, I will rebase my commits to keep just one.

saghm:
Sounds good! I agree there's no reason to modify the commits you made until we're ready to merge.

I just authorized the CI run; once it passes, I'll mark this as ready for review and add the others!

darnuria:
Thanks I took the liberty to make it available for review since the result of the CI test were OK.

image

saghm:
Just merged; thanks for the PR!

darnuria:
Thanks for time and review! :)

PR #204, opened 2020-07-27
Add ObjectIdRepr that deconstruct an ObjectId

Stargateur:
I have the need to use `ObjectId` counter so maybe this could be useful to add some helper. As `ObjectId` is documented to have a specific representation I see no reason to not add this. https://docs.mongodb.com/manual/reference/method/ObjectId/#description

Note: I add byteorder as dependencies to make it simple, I think I could remove this dependencies but before doing some annoying code I prefer to see if this is useful or not.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ decimal128 = ["decimal"]
 name = "bson"
 
 [dependencies]
+byteorder = "1"
 chrono = "0.4"
 rand = "0.7"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -15,6 +15,10 @@ use rand::{thread_rng, Rng};
 
 use lazy_static::lazy_static;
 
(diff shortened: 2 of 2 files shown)


patrickfreed:
Hi @Stargateur, thanks for the pull request! Unfortunately, we won't be able to merge this since providing such a representation would violate the [ObjectID driver specification](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#specification). Specifically, the specification forbids providing accessors for either the random value or counter fields. The rationale is that these portions of the ObjectID are only meant to ensure randomness and uniqueness and aren't supposed to be queryable themselves.

All that being said, we do provide the bytes that make up the `ObjectId` via [`ObjectId::bytes`](https://docs.rs/bson/1.1.0-beta/bson/oid/struct.ObjectId.html#method.bytes), so you should be able to implement this in your application if necessary.

Stargateur:
I could argue that you are talking about mongodb driver whereas here we are on bson. I don't see obvious reason to not to expect it's written. I would prefer rational of this choice.

PR #210, opened 2020-08-06
RUST-503 Support serializing directly to Document

isabelatkinson:
[RUST-503](https://jira.mongodb.org/browse/RUST-503)

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -197,8 +197,11 @@ where
 {
     match to_bson(value)? {
         Bson::Document(doc) => Ok(doc),
-        _ => Err(Error::SerializationError {
-            message: "Cannot be serialized to Document".to_string(),
+        bson => Err(Error::SerializationError {
+            message: format!(
+                "Could not be serialized to Document, got {:?} instead",
+                bson.element_type()
+            ),
         }),
     }
 }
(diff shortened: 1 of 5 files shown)


src/ser/mod.rs
```diff
@@ -189,3 +189,16 @@ where
...
     value.serialize(ser)
 }
+
+/// Encode a `T` Serializable into a BSON `Document`.
+pub fn to_document<T: ?Sized>(value: &T) -> Result<Document>
+where
+    T: Serialize,
+{
+    let bson = to_bson(value)?;
+    bson.as_document()
```
saghm, line 199: I think it would be more efficient to just pattern match the value and return it directly instead of cloning it. I know that using combinators is more fun though, but it's worth it to keep things as quick as possible for users :/

src/ser/mod.rs
```diff
@@ -189,3 +189,16 @@ where
...
+
+/// Encode a `T` Serializable into a BSON `Document`.
+pub fn to_document<T: ?Sized>(value: &T) -> Result<Document>
+where
+    T: Serialize,
+{
+    match to_bson(value)? {
+        Bson::Document(doc) => Ok(doc),
+        _ => Err(Error::SerializationError {
+            message: "Cannot be serialized to Document".to_string(),
```
patrickfreed, line 201: can we also include what T was serialized to in this message?

PR #214, opened 2020-09-23
RUST-364 Cache unique process value for oids

saghm:
We already implemented the ObjectId spec tests in RUST-484, so no new tests were added here. The only deviation from the spec I noticed was that we were generating new random bytes for each ObjectId rather than caching them per process, so I've updated the implementation here to cache them.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -162,7 +162,7 @@ impl ObjectId {
             };
         }
 
-        BUF.clone()
+        *BUF
     }
 
     // Gets an incremental 3-byte count.
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -153,10 +153,16 @@ impl ObjectId {
 
     // Generate a random 5-byte array.
     fn gen_process_id() -> [u8; 5] {
-        let rng = thread_rng().gen_range(0, MAX_U24) as u32;
-        let mut buf: [u8; 5] = [0; 5];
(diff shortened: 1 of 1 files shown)


patrickfreed (approved):
code lgtm but linter seems upset (I don't need to re-review once that's fixed)

saghm:
Interesting, I didn't realize that certain-sized arrays of Copy types are also Copy

src/oid.rs
```diff
@@ -153,10 +153,16 @@ impl ObjectId {
 
     // Generate a random 5-byte array.
     fn gen_process_id() -> [u8; 5] {
-        let rng = thread_rng().gen_range(0, MAX_U24) as u32;
-        let mut buf: [u8; 5] = [0; 5];
-        buf[0..4].copy_from_slice(&rng.to_be_bytes());
-        buf
+        lazy_static! {
```
isabelatkinson, line 156: When does the code in the `lazy_static` block actually execute?

isabelatkinson (approved):
looks good, just a question for my own understanding!

src/oid.rs
saghm: It will execute the first time that the reference it defines (i.e. `BUF`) is dereferenced, which will be the first this the method is called. [`lazy_static`'s documentation](https://docs.rs/lazy_static/1.4.0/lazy_static/#semantics) has more details on this.

PR #221, opened 2020-12-01
RUST-507 Deserialize unsigned integers

isabelatkinson:


src/de/serde.rs
```diff
@@ -230,6 +218,22 @@ impl<'de> Visitor<'de> for BsonVisitor {
...
 
+fn convert_unsigned_to_signed<E>(value: u64) -> Result<Bson, E>
+where
+    E: Error,
+{
+    if let Ok(int32) = i32::try_from(value) {
+        Ok(Bson::Int32(int32))
+    } else if let Ok(int64) = i64::try_from(value) {
+        Ok(Bson::Int64(int64))
+    } else {
```
isabelatkinson, line 229: I decided to remove the `f64` comparison since all of the conversion strategies seem to allow for some lossy conversion (see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f404eec1051ebe6eba01a9d5ff269907)). `i32` and `i64` both implement `TryFrom` for `u64`, but `f64` does not. The strategy used in the `bson_util::get_int` method and the `FromPrimitive` trait in `num` both result in inaccuracies. It would be possible to do some string comparison, but I didn’t think that was worth what would probably be a performance hit — let me know if you think it would be worth it, though.

saghm: Does this cause any behavioral changes where something that would work before no longer works?

isabelatkinson: Nope, previously every method that calls this new function would just error out (by remove I mean it was originally included in the design).

saghm: Ah, okay. This seems fine with me. We can always add something opt-in for floats later if users ask for it.

saghm (approved):

src/de/serde.rs
patrickfreed: What is the error bound on the `num` conversions? If its something _really_ small, then I'd personally be okay with the possible lossy-ness.

isabelatkinson: I'm not sure, but in the example I linked above it allows for a difference of 385 so I don't think we should rely upon it to convert accurately for large numbers.

patrickfreed: Oh yeah, we definitely can't have that.

patrickfreed (approved):

Issue #122, opened 2019-07-31
Fail to deserialize `TimeStamp` .


weiby3:
I want to deserialize str to my struct, but it gets error. The sample code is:
```rust
use serde::{Serialize, Deserialize};
#[derive(Serialize, Deserialize, Debug)]
struct Foo {
    ts: bson::TimeStamp,
}
fn main() -> Result<(), Box<dyn Error>> {
    let original_str = "{\"ts\":1403964864000}";
    let decoded_json = dbg!(serde_json::from_str::<Foo>(original_str)?);
    Ok(())
}
```
Its output is:
```
Error: Error("invalid type: integer `1403964864000`, expected a signed integer", line: 1, column: 19)
```

zonyitoo:
`bson::TimeStamp` is only for converting from/to `Bson`.

saghm:
Hello @weiby3! To clarify further, the `TimeStamp` type is for the the timestamp type as described [here](http://bsonspec.org/spec.html); this is not a type that exists in JSON, so you would need to use extended JSON to deserialize it into Bson. Although we don't currently have a spec-compliant extended JSON implementation in this library, as a workaround you can use a JSON object with the fields `"t"` and `"i"`, which will be converted to a BSON timestamp when the `From<serde_json::Value>` implementation on `Bson` is used:

```rust
use bson::Bson;

fn main() {
    let json = serde_json::json!({
        "t": 12345,
        "i": 54321
    });

    println!("{}", Bson::from(json));
}
```

patrickfreed:
In the 1.0.0 release, a spec compliant extended JSON implementation was completed, so the following should work now (note that `try_from` must be used instead of `from`):
```rust
let json = json!({ "$timestamp": { "t": 12345, "i": 54321 }});
println!("{:?}", bson::Bson::try_from(json).unwrap());
```
I filed RUST-507 to track work for enabling the functionality requested in the original issue (deserializing numbers from JSON or other formats).

agolin95:
Work in RUST-507 is complete! This is available as part of the 1.2.0 release. Please let us know if you encounter any issues!

PR #222, opened 2020-12-04
RUST-506 Add serde helper functions

isabelatkinson:


src/serde_helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+    match bson {
+        Bson::ObjectId(oid) => Ok(oid),
+        _ => Err(de::Error::custom(
+            "cannot convert extended JSON to ObjectId",
+        )),
+    }
+}
+
+/// Serializes a hex string as an ObjectId.
+pub fn serialize_hex_string_as_object_id<S: Serializer>(
```
isabelatkinson, line 106: No corresponding deserialize function for this one since checking for a hex string is already default behavior.

src/serde_helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+pub fn deserialize_uuid_from_binary<'de, D>(deserializer: D) -> Result<Uuid, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let binary = Binary::deserialize(deserializer)?;
+    if binary.subtype == BinarySubtype::Uuid {
+        if binary.bytes.len() == 16 {
+            let mut bytes = [0u8; 16];
+            for (i, byte) in bytes.iter_mut().enumerate() {
+                *byte = binary.bytes[i];
```
isabelatkinson, line 141: There is a `TryFrom` implementation that makes this conversion a lot simpler, but it's only available in 1.48+. Since we support lower compiler versions this roundabout way is necessary -- the index is safe since the size of the vector is guaranteed in the previous check.

src/serde_helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+        ))
+    }
+}
+
+/// Serializes a u64 as a Timestamp.
+pub fn serialize_u64_as_timestamp<S: Serializer>(
+    val: &u64,
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    let time = (*val >> 32) as u32;
```
isabelatkinson, line 161: Based the math for these functions on [this documentation](https://docs.mongodb.com/manual/reference/bson-types/#timestamps) but it could definitely use a sanity check.

src/serde_helpers.rs
saghm: I think it might be more idiomatic to use [`copy_from_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) here rather than copying each byte manually. Clippy sometimes warns about things like this, although I'm not sure if it can detect it all of the time.

src/serde_helpers.rs
```diff
@@ -0,0 +1,175 @@
...
+    let increment = *val as u32;
```
saghm, line 162: Do we need a bitmask here? I'm honestly not sure what the semantics are for casting a u64 higher than `u32::MAX` to a `u32`.

src/tests/serde.rs
```diff
@@ -594,3 +600,188 @@ fn test_serialize_deserialize_unsigned_numbers() {
...
+        #[serde(serialize_with = "serde_helpers::serialize_u64_as_i32")]
+        num_2: u64,
+    }
+
+    let a = A { num_1: 1, num_2: 2 };
+    let doc = to_document(&a).unwrap();
+    assert!(doc.get_i32("num_1").unwrap() == 1);
+    assert!(doc.get_i32("num_2").unwrap() == 2);
+
+    let a = A {
```
saghm, line 621: I think this should be split into two separate failure cases, one where only `num_1` is too high and the other where only `num_2` is. As it's written right now, assuming `serialize_u2_as_i32` works properly, I don't think `serialize_u64_as_i32` is being tested.

src/tests/serde.rs
```diff
@@ -594,3 +600,188 @@ fn test_serialize_deserialize_unsigned_numbers() {
...
+    let json: Value = serde_json::from_str(&date).unwrap();
+    let b: B = serde_json::from_value(json).unwrap();
+    let expected: chrono::DateTime<chrono::Utc> =
+        chrono::DateTime::from_str("2020-06-09 10:58:07.095 UTC").unwrap();
+    assert_eq!(b.date, expected);
+
+    #[derive(Deserialize, Serialize)]
+    struct C {
+        #[serde(
+            serialize_with = "serde_helpers::serialize_iso_string_as_datetime",
```
saghm, line 692: Serde provides a way to specify both serialization and deserialization with the same attribute [through modules](https://serde.rs/field-attrs.html#with). I think this is a good case for it; in addition to defining the functions that already exist, we could make a module that just defines thin wrappers around each of those two functions so that users who want full roundtripping between ISO strings and DateTime can do that with one attribute.

src/tests/serde.rs
```diff
@@ -594,3 +600,188 @@ fn test_serialize_deserialize_unsigned_numbers() {
...
+}
+
+#[test]
+fn test_uuid_helpers() {
+    let _guard = LOCK.run_concurrently();
+
+    #[derive(Serialize, Deserialize)]
+    struct A {
+        #[serde(
+            serialize_with = "serde_helpers::serialize_uuid_as_binary",
```
saghm, line 746: Ditto above about considering a module for allowing roundtripping with a single attribute.

src/tests/serde.rs
```diff
@@ -594,3 +600,188 @@ fn test_serialize_deserialize_unsigned_numbers() {
...
+}
+
+#[test]
+fn test_timestamp_helpers() {
+    let _guard = LOCK.run_concurrently();
+
+    #[derive(Deserialize, Serialize)]
+    struct A {
+        #[serde(
+            serialize_with = "serde_helpers::serialize_u64_as_timestamp",
```
saghm, line 773: Ditto above about considering a module for allowing roundtripping with a single attribute.

isabelatkinson: Not sure I understand the differentiation between `u64`/`Timestamp` and `Uuid`/`Binary` here -- wouldn't it be possible to implement a simple helper there as well? I think the benefit here is similar to that of the previous serde work to accept user types into crud methods in that it allows users not to do any conversions/not need to think about `Timestamp` (besides adding the attribute) and instead working directly with `u64`.

patrickfreed: I think one issue here is that timestamps, when represented as integers, are typically the number of seconds since the epoch, but the BSON timestamp type supports sub-second accuracy by means of the ordinal. This means using the `deserialize_u64_from_timestamp` is lossy for any timestamp with an ordinal assuming we're using the unix timestamp as the format of our `u64`. We could opt to use the `u64` as the storage of the timestamp as it's used in BSON, but I think that would be confusing to users who are likely expecting the number to be a unix timestamp. 

Given these concerns, I don't think we necessarily want to encourage users to use `u64` in place of `Timestamp` when deserializing from BSON or extended JSON. The serialization case seems ok, and both serialization to / deserialization from `u64` for `Timestamp` should be fine as well, since it is the more precise type. In the UUID case, the conversion is lossless, so it's okay to use the `Uuid` type as a drop-in replacement for BSON binary. If a `u64` (as a unix timestamp) could be losslessly converted to/from BSON timestamp, then I'd agree that there is no meaningful difference though.

src/serde_helpers.rs
```diff
@@ -0,0 +1,226 @@
...
+
+/// Attempts to serialize a u64 as an i64. Errors if an exact conversion is not possible.
+pub fn serialize_u64_as_i64<S: Serializer>(val: &u64, serializer: S) -> Result<S::Ok, S::Error> {
+    match i64::try_from(*val) {
+        Ok(val) => serializer.serialize_i64(val),
+        Err(_) => Err(ser::Error::custom(format!("cannot convert {} to i64", val))),
+    }
+}
+
+pub mod chrono_datetime_to_bson_datetime {
```
patrickfreed, line 38: I think instead of `thing1_to_thing2`, we should name these something that works for both serialization and deserialization. Maybe `thing1_as_thing2`?

patrickfreed, line 38: Can we document each of these modules + include code samples for them?

src/serde_helpers.rs
```diff
@@ -0,0 +1,226 @@
+//! Collection of helper functions for serializing to and deserializing from BSON using Serde
+
+use std::{convert::TryFrom, result::Result};
+
+use serde::{de, ser, Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::{oid::ObjectId, Bson};
+
+/// Attempts to serialize a u32 as an i32. Errors if an exact conversion is not possible.
+pub fn serialize_u32_as_i32<S: Serializer>(val: &u32, serializer: S) -> Result<S::Ok, S::Error> {
```
patrickfreed, line 10: for the ones now included as part of modules, it would probably be nice to have aliases at the top level to match the format of these.

src/tests/serde.rs
isabelatkinson: I believe `u64 -> Timestamp` can also be lossy, since the seconds since the Unix epoch is stored in a `u32` in `Timestamp` (so anything greater than `u32::MAX` would be truncated). We do have [one user](https://github.com/mongodb/bson-rust/issues/122) issue asking for this functionality though, so it might be best to assume the `u64` is seconds from the Unix epoch and error out in the case of a lossy conversion (and make sure that behavior is well-documented). Alternatively, we could include sets of functions for both seconds from the Unix epoch and the BSON storage including an increment, but I think it would be best to hold off on the latter unless someone asks for it.

patrickfreed: Oh yeah that's a good point. It seems like we actually just want conversions between `u32` and what not then, if we're going to be erroring on anything greater anyways.

And yeah I agree. I doubt anyone is going to ask for having the BSON storage as an actual integer.

src/serde_helpers.rs
patrickfreed: Yeah you can via the `use` keyword. It's the same mechanism by which we move our exports around:

```rust
pub use some_module::some_fn as toplevel_some_fn;
```

patrickfreed (approved):
LGTM!

saghm (approved):
LGTM! I assume the lint errors are the ones that your PR from earlier this week fixed

PR #224, opened 2021-01-04
minor: fix clippy

isabelatkinson:


src/tests/modules/oid.rs
```diff
@@ -27,6 +27,7 @@ fn byte_string_oid() {
 }
 
 #[test]
+#[allow(clippy::eq_op)]
```
isabelatkinson, line 30: [eq_op](https://rust-lang.github.io/rust-clippy/master/index.html#eq_op) produces a warning when something is compared to itself, but we need to do that in this test.

src/de/mod.rs
```diff
@@ -212,7 +212,7 @@ pub(crate) fn deserialize_bson_kvp<R: Read + ?Sized>(
         Some(ElementType::Array) => deserialize_array(reader, utf8_lossy).map(Bson::Array)?,
         Some(ElementType::Binary) => {
             let mut len = read_i32(reader)?;
-            if len < 0 || len > MAX_BSON_SIZE {
+            if !(0..=MAX_BSON_SIZE).contains(&len) {
```
saghm, line 215: Interesting, I find this far less readable, and it seems like it would be slower (since it has to do a linear search rather than two comparisons), but if clippy wants it this way, who am I to argue?

saghm (approved):

patrickfreed (approved):

PR #227, opened 2021-01-26
RUST-642 Remove travis-ci

isabelatkinson:


saghm (approved):
I think we'll also need to change a repository setting somewhere to turn off the trigger on pull requests; I _think_ you should have permissions to do this, but let me know if you aren't able to

patrickfreed (approved):

saghm (approved):

PR #228, opened 2021-01-26
RUST-647 Sync spec tests for invalid relaxed UUIDs

isabelatkinson:


src/extjson/models.rs
```diff
@@ -173,42 +173,21 @@ pub(crate) struct Uuid {
 }
 
 impl Uuid {
-    pub(crate) fn parse(mut self) -> extjson::de::Result<crate::Binary> {
-        if !valid_uuid_format(&self.value) {
```
isabelatkinson, line 177: The checks in `valid_uuid_format` do not account for extraneous hyphens (covered in [this new test](https://github.com/mongodb/specifications/blob/0a03c879e6277e38284c73f9b424b30fb9f8d7be/source/bson-corpus/tests/binary.json#L109)). Since we added the `uuid` crate as a dependency when implementing the serde helpers, I switched this method to parse into a `uuid::Uuid` rather than do the work manually.

patrickfreed (approved):

saghm (approved):

PR #229, opened 2021-02-07
Incorporate rawbson code from rawbson = "0.2.1" in mod raw

jcdyer:
Also adds criterion benchmarks


saghm:
Thanks for the pull request @jcdyer! Is it all right with you if I start pushing commits to this branch? We're interested in getting this out to users as soon as possible (specifically in our 2.0 release, which we just released an alpha for), so it will go more quickly if we can write the code for changes we want to make ourselves. We definitely would love your input as a reviewer for the code we write!

jcdyer:
by all means! Use it as you see fit.  I'm happy to answer questions, discuss implementation ideas, review code, or whatever helps.

src/raw/mod.rs
```diff
@@ -0,0 +1,1849 @@
...
+    /// ```
+    /// # use bson::raw::{DocBuf, DocRef, RawError};
+    /// let docbuf = DocBuf::new(b"\x13\x00\x00\x00\x02hi\x00\x06\x00\x00\x00y'all\x00\x00".to_vec())?;
+    /// let docref: DocRef = docbuf.as_docref();
+    /// # Ok::<(), RawError>(())
+    /// ```
+    #[deprecated(since = "0.2.0", note = "use docbuf.as_ref() instead")]
+    pub fn as_docref(&self) -> &Doc {
+        self.as_ref()
+    }
```
jcdyer, line 331: I recommend yanking out all mentions of DocRef. It doesn't need to exist anymore.

src/raw/mod.rs
```diff
@@ -0,0 +1,1849 @@
...
+    /// ```
+    ///
+    /// # Safety
+    ///
+    /// The provided bytes must have a valid length marker, and be NUL terminated.
+    pub unsafe fn new_unchecked(data: Vec<u8>) -> DocBuf {
+        DocBuf {
+            data: data.into_boxed_slice(),
+        }
+    }
```
jcdyer, line 312: This function doesn't actually do anything unsafe.  I marked it that way out of a suspicion that it might someday in the future, so it's marked to prevent backwards compatibility semver issues, but everything interacts with the underlying slice in a completely safe manner.

src/raw/mod.rs
```diff
@@ -0,0 +1,1849 @@
...
+pub mod de;
+pub mod elem;
+
+#[cfg(test)]
+mod props;
+
+/// Error to indicate that either a value was empty or it contained an unexpected
+/// type, for use with the direct getters.
+#[derive(Debug, PartialEq)]
+pub enum RawError {
```
jcdyer, line 164: It would be nice to merge this with the primary crate Error type. The main difference is needing to be able to signal malformed bsom when accessing elements.

src/raw/mod.rs
```diff
@@ -0,0 +1,1849 @@
...
+    /// Found a Bson value with the specified key, but not with the expected type
+    UnexpectedType,
```
jcdyer, line 166: This should be updated to contain the spec::ElementType that was actually found

saghm:
Thanks so much for this PR! We're going to hold off for a couple of weeks before merging the serde implementation in order to focus on the API that we need to be able to update the driver for the 2.0 release, but we definitely plan on using it once we're ready, and we'll make sure that you get proper credit for it.

I've pushed some small API and documentation changes as a separate commit and added the rest of the team as reviewers to the pull request.

src/raw/mod.rs
saghm: Yep, that makes sense. I removed this method in my follow-up commit.

src/raw/mod.rs
saghm: Interesting; I was wondering why it was marked as unsafe. I think if it ever does need to become unsafe, we would probably prefer to just add a separate method and deprecate this one rather than marking it as unsafe even though it currently doesn't need to be, so I've removed that annotation in my follow-up commit.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+    ///
+    /// let document = doc! {
+    ///     "_id": ObjectId::new(),
+    ///     "name": "Herman Melville",
+    ///     "title": "Moby-Dick",
+    /// };
+    /// let doc = RawDocument::from_document(&document);
+    /// # Ok::<(), RawError>(())
+    /// ```
+    pub fn from_document(doc: &Document) -> RawDocument {
```
patrickfreed, line 242: should we just `impl From<Document> for RawDocument`?

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+        // SAFETY:
+        //
+        // Dereferencing a raw pointer requires unsafe due to the potential that the pointer is
+        // null, dangling, or misaligned. We know the pointer is not null or dangling due to the
+        // fact that it's created by a safe reference. Converting &[u8] to *const [u8] will be
+        // properly aligned due to them being references to the same type, and converting *const
+        // [u8] to *const RawDocumentRef is aligned due to the fact that the only field in a
+        // RawDocumentRef is a [u8], meaning the structs are represented identically at the byte
+        // level.
+        unsafe { &*(data.as_ref() as *const [u8] as *const RawDocumentRef) }
```
patrickfreed, line 421: Would it not be easier / avoid `unsafe` if we just added a lifetime annotation to `RawDocumentRef` and stored an `&'a [u8]` instead of just a `[u8]`? I'm not sure I understand why this is preferable.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+
+    fn into_iter(self) -> RawDocumentIter<'a> {
+        RawDocumentIter {
+            doc: self,
+            offset: 4,
+        }
+    }
+}
+
+pub struct RawDocumentIter<'a> {
```
patrickfreed, line 792: can we add a docstring for this type?

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+                    self.doc.data[self.offset]
+                ))))
+            }
+        };
+
+        let element_size = match element_type {
+            ElementType::Double => 8,
+            ElementType::String => {
+                let size =
+                    4 + i32_from_slice(&self.doc.data[valueoffset..valueoffset + 4]) as usize;
```
patrickfreed, line 833: this may panic if the data is too short, since we never check that valueoffset is less than `data.len()`. This applies in a lot of other places in this function.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+        // SAFETY:
+        //
+        // Dereferencing a raw pointer requires unsafe due to the potential that the pointer is
+        // null, dangling, or misaligned. We know the pointer is not null or dangling due to the
+        // fact that it's created by a safe reference. Converting &RawDocumentRef to *const
+        // RawDocumentRef will be properly aligned due to them being references to the same type,
+        // and converting *const RawDocumentRef to *const RawArray is aligned due to the fact that
+        // the only field in a RawArray is a RawDocumentRef, meaning the structs are represented
+        // identically at the byte level.
+        unsafe { &*(doc as *const RawDocumentRef as *const RawArray) }
```
patrickfreed, line 965: i think this `unsafe` can similarly be avoided by making `RawDocumentRef` have a lifetime annotation.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+
+fn try_to_str(data: &[u8]) -> RawResult<&str> {
+    match std::str::from_utf8(data) {
+        Ok(s) => Ok(s),
+        Err(_) => Err(RawError::Utf8EncodingError(data.into())),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
patrickfreed, line 1162: can we break this module off into its own file?

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+    type Item = RawResult<elem::RawBson<'a>>;
+
+    fn into_iter(self) -> RawArrayIter<'a> {
+        RawArrayIter {
+            inner: self.doc.into_iter(),
+        }
+    }
+}
+
+pub struct RawArrayIter<'a> {
```
patrickfreed, line 1083: can we add a docstring?

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+
+        Some(Ok((
+            key,
+            elem::RawBson::new(element_type, &self.doc.data[valueoffset..nextoffset]),
+        )))
+    }
+}
+
+/// A BSON array referencing raw bytes stored elsewhere.
+pub struct RawArray {
```
patrickfreed, line 946: can we break `RawArray` and `RawDocument` + `RawDocumentRef` into their own files in the `raw` module?

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+/// let doc = RawDocument::new(b"\x13\x00\x00\x00\x02hi\x00\x06\x00\x00\x00y'all\x00\x00".to_vec())?;
+/// assert_eq!(doc.get_str("hi")?, Some("y'all"));
+/// # Ok::<(), RawError>(())
+/// ```
+#[derive(Clone, Debug)]
+pub struct RawDocument {
+    data: Box<[u8]>,
+}
+
+impl RawDocument {
```
patrickfreed, line 188: I think we'll want some methods that support appending to this type, otherwise users will have to route through `Document`, negating many of the performance benefits of sticking with raw.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+/// requires iterating through the document from the beginning to find the requested key.
+///
+/// ```
+/// # use bson::raw::{RawDocument, RawError};
+/// let doc = RawDocument::new(b"\x13\x00\x00\x00\x02hi\x00\x06\x00\x00\x00y'all\x00\x00".to_vec())?;
+/// assert_eq!(doc.get_str("hi")?, Some("y'all"));
+/// # Ok::<(), RawError>(())
+/// ```
+#[derive(Debug)]
+pub struct RawDocumentRef {
```
patrickfreed, line 366: We'll probably also want a mutable version of this type to support appending to subdocuments without copying, ditto for arrays. I don't think there's a way for us to have an impl block conditional on whether the underlying reference is mutable or not, which would let us have a single type.

src/raw/props.rs
```diff
@@ -0,0 +1,59 @@
+use crate::{spec::BinarySubtype, Binary, Bson, Document, JavaScriptCodeWithScope, Regex};
```
patrickfreed, line 1: I'm not sure I quite understand exactly what this is doing, could we add some comments inline to explain? Also, can we move this to the `test` module when created?

patrickfreed (commented):
Hi @jcdyer, thanks again for the PR on this--it looks great so far! To avoid possibly duplicating any work though, @saghm will handle making any code changes @isabelatkinson and I request going forward, but feel free to chime in during the review as you see fit.

src/raw/elem.rs
```diff
@@ -0,0 +1,409 @@
...
+};
+use crate::{
+    oid::ObjectId,
+    spec::{BinarySubtype, ElementType},
+    Bson,
+};
+
+/// A BSON value referencing raw bytes stored elsewhere.
+#[derive(Clone, Copy, Debug)]
+pub struct RawBson<'a> {
```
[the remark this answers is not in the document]
saghm: Hmm, I have mixed feelings on this. It definitely would be more ergonomic for users, but it semantically doesn't seem quite the same to me. With the existing Document and Bson types, the Document very explicitly contains each Bson value, and with RawDocument and RawBson as written currently in the API, this is also true modulo the ElementType. With the enum as you've written it, the `f64` or `bool` contained in the `RawBson` don't actually exist as part of the RawDocument. I somewhat prefer the idea of having the RawBson values that are accessed through the RawDocument actually being subsets of the RawDocument like with Document and Bson, although I'm having trouble articulating exactly why that is.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+/// An error that occurs when attempting to parse raw BSON bytes.
+#[derive(Debug, PartialEq)]
+pub enum RawError {
+    /// A BSON value did not fit the expected type.
+    UnexpectedType,
+
+    /// A BSON value did not fit the proper format.
+    MalformedValue(String),
+
+    /// Improper UTF-8 bytes were found when proper UTF-7 was expected. The error value contains
```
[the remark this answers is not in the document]
saghm: Are you sure? We could always decide to just roll with it and change BSON to use that format, which apparently actually exists: https://en.wikipedia.org/wiki/UTF-7

src/raw/elem.rs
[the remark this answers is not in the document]
jcdyer: Indeed, I was actually thinking about an alternative version of RawBson where, from a data standpoint, it's just a slice of a bson document (or array) from the element type byte to the start of the next element type byte.  So a document would be `length + element* + \0`.  Then RawBson could also be made an unsized by-reference type, but then its data would also include the key used to reference it, which is kind of a weird mismatch with how the parsed type works or how people  would expect it to work, based on their experience with JSON.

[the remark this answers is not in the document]
jcdyer: One use case that the current design unlocks that the straight-enum version might break is dealing with not-quite valid BSON data.  I've encountered mongo databases in the wild that contain strings with invalid utf-8 data.  (Specifically it was an ObjectId that somehow got entered as a string.  It was a well-formed string type (had a four-byte length marker), but the data was binary, not utf8.  It was impossible to use python to access that database, because it raised exceptions whenever hitting that data.  Go could handle it because go only pretends that its strings are UTF-8.  The current rust implementation also failed, because it coerced the data to a String immediately on fetching the data.

The current RawBson design decodes the data on request, and provides a way to access the raw bytes as such, specifically to help people work around issues like the one I faced.  It's not particularly ergonomic (kind of intentionally so), but it's *possible*, and that's very valuable.

src/raw/elem.rs
```diff
@@ -0,0 +1,409 @@
...
+        if let ElementType::Symbol = self.element_type {
+            read_lenencoded(self.data)
+        } else {
+            Err(RawError::UnexpectedType)
+        }
+    }
+
+    /// Gets the BSON JavaScript code with scope that's referenced or returns an error if the value
+    /// isn't BSON JavaScript code with scope.
+    pub fn as_javascript_with_scope(self) -> RawResult<(&'a str, &'a RawDocumentRef)> {
```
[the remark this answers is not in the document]
jcdyer: :+1:

src/raw/elem.rs
```diff
@@ -0,0 +1,409 @@
...
+        } else {
+            Err(RawError::UnexpectedType)
+        }
+    }
+
+    /// Gets the BSON binary value that's referenced or returns an error if the value a BSON binary.
+    pub fn as_binary(self) -> RawResult<RawBinary<'a>> {
+        if let ElementType::Binary = self.element_type {
+            let length = i32_from_slice(&self.data[0..4]);
+            let subtype = BinarySubtype::from(self.data[4]); // TODO: This mishandles reserved values
```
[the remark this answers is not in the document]
jcdyer: Note that [the spec](http://bsonspec.org/spec.html) is inconsistent here.  The grammar defines the values `\x01`-`\x06` and `\x80` (user defined), but the accompanying explanatory text says:

> The BSON "binary" or "BinData" datatype is used to represent arrays of bytes. It is somewhat analogous to the Java notion of a ByteArray. BSON binary values have a subtype. This is used to indicate what kind of data is in the byte array. Subtypes from zero to 127 are predefined or reserved. Subtypes from 128-255 are user-defined.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+//! A [`RawDocument`] can be created from a [`bson::document::Document`]. Internally, this
+//! serializes the `Document` to a `Vec<u8>`, and then includes those bytes in the [`RawDocument`].
+//!
+//! ```rust
+//! use bson::{
+//!     raw::RawDocument,
+//!     doc,
+//! };
+//!
+//! let document = doc! {
```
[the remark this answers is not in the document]
jcdyer: I don't remember if I got around to adding a `fn insert(&mut self, &str, Element)` method yet, but that would probably be a useful step for creating a rawdoc macro.  If it doesn't exist yet, a first version could just wrap doc!, though it would be less performant than direct insert.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+pub enum RawError {
+    /// A BSON value did not fit the expected type.
+    UnexpectedType,
+
+    /// A BSON value did not fit the proper format.
+    MalformedValue(String),
+
+    /// Improper UTF-8 bytes were found when proper UTF-7 was expected. The error value contains
+    /// the malformed data as bytes.
+    Utf8EncodingError(Vec<u8>),
```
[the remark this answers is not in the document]
jcdyer: Sounds like a good idea to me.

src/raw/mod.rs
```diff
@@ -0,0 +1,1634 @@
...
+    /// Return the contained data as a `Vec<u8>`
+    ///
+    /// ```
+    /// # use bson::raw::RawDocument;
+    /// use bson::doc;
+    ///
+    /// let doc = RawDocument::from_document(&doc!{});
+    /// assert_eq!(doc.into_inner(), b"\x05\x00\x00\x00\x00".to_vec());
+    /// ```
+    pub fn into_inner(self) -> Vec<u8> {
```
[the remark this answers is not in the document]
jcdyer: .into_inner is a naming convention I [borrowed from the standard library](https://doc.rust-lang.org/std/?search=into_inner), but into_vec() is also common.  I do think it should exist as an inherent method by one of those names, since it's fundamental to the data structure itself that it is a wrapper around heap allocated bytes, and not just some arbitrary conversion.

src/raw/mod.rs
jcdyer: This makes it possible to implement Deref.  The unsafe pattern is taken straight from the way Path is implemented.

Before this, all the methods on RawDocumentRef were duplicated on RawDocument with .as_ref().method() implementations.

jcdyer: The other benefit is that having a Buffer type and an unsized reference type is a familiar pattern from Vec<T> / [T], String / str, PathBuf / Path , OsString / OsStr, etc.

src/raw/mod.rs
jcdyer: I had started some work on this that I guess never made it in.  One thing to note is that keeping the methods performant necessitates not checking for duplicate keys, since finding a key is O(N) on the length of the document.  Similarly for finding the length of a RawArray (which is also a big ergonomic hit, since `array.append(x)` becomes `array.insert(the_right_index, x)`.  One option for arrays would be a RawArrayAppender type, that lets you append multiple entries while only checking the current length once, when the appender is created.  I don't love that either, though.

src/raw/mod.rs
jcdyer: That won't work, since the subdocument is part of a larger Vec, so even if it were made mutable, the mutability would be strictly limited to changing values that are fixed size.  For those kinds of modifications, you really do need to parse the document into a more flexible type, and take the performance hit.

src/raw/props.rs
jcdyer: proptests generate a large number of random inputs, run tests against those inputs, then minimize the input until it finds the smallest failing example it can.  These are the functions for generating arbitrary BSON objects, so proptest can verify that arbitrary BSON round-trips properly.  

Further reading:
* [A blog post introducing property testing in rust](https://blog.logrocket.com/property-based-testing-in-rust-with-proptest/)
* [The guide to the proptest library](https://altsysrq.github.io/proptest-book/intro.html)

src/raw/mod.rs
jcdyer: Finally, storing an &'a [u8] forces this type to only be used as a reference type.  An usized RawDocumentRef type can also be stored in a Box / Rc / Arc / other smart pointer, and be passed around as an owned type, saving an extra pointer indirection over boxing a RawDocument.

jcdyer:
To summarize some of my comments from the other day, I'd caution against thinking about (or advertising) `RawDocument` simply as a *faster* `Document`, or about trying to force the two to look exactly the same at the API level.  It is a different data structure with different performance characteristics.  In many cases, it is faster, but not all.

In some cases, as the benchmarks show, `RawDocument` is significantly slower than `Document`. For instance, reading an element out of a document by name is `O(n)` in RawDocument,  since  you have to start at the beginning and read each element out of the `RawDocument` until you find the one you want, while `Document` is O(1), since it is just a hashmap.  If you only need a single element, `RawDocument` is faster since the parsing time overwhelms the measurement, but as the number of elements  accessed goes up, raw document becomes O(M * N) (M= number of accesses, N=size of document), while Document is O(M * 1).  

Where there is clear similarity between the two types, it makes sense to keep a similar API, but document mutation is not one of those situations.  Raw document can support some mutations fairly simply, but others are quite complex, and should be discouraged:

### RawDocument append

Easy. O(1) performance (assuming a reasonably sized element) 
1. Remove the trailing \0 
2. add the new element
3. append a trailing \0
4. rewrite the length marker at the beginning of the document.

Requires an O(n) read to guarantee non-duplicate keys, because you have to iterate over the document to compare with the existing keys.

### RawArray append

Note: there currently isn't an owning RawArray type, since that's not part of the bson spec.  If such a type were created, this could be implemented.

Easy, O(n) performance:

1. Iterate the document to get the current length
2. append as above with length as the key.

### Non-size-changing element modification 

(e.g., changing the value of a fixed size element, without changing its type)

1. Access the desired element (O(n))
2. Modify the element.

### General element modification / element replacement

Slow, moderately complicated

1. Find the desired element (O(n))
2. Reserve all following elements to a buffer. (O(n) write)
3. Modify / replace the element.
4. Append the reserved Elements (O(n) write)
5. Rewrite the length marker at the beginning of the document.

### Modifying a nested document: 

Slow and complicated (O(N * depth of nesting)).  

For a nested key x.y.z:
1. Find element x.  Copy all following elements to a buffer as for general modification
2. Find element y in embedded document x.  Copy all following elements to a buffer as for general element modification.
3. Find element z in embedded document y: Copy all following elements to a buffer as for general element modification.
4. Modify element z.
5. Append reserved elements for document y.
6. Rewrite the length marker at the beginning of y.
7. Append reserved elements for document x.
8. Rewrite the length marker at the beginning of x.
9. Append reserved elements for top-level document.
10. Rewrite the length marker for top-level document.

Note, this algorithmic complexity is exacerbated by the fact that the data is owned by the top-level document, so inner documents aren't even capable of resizing themselves to accomodate modified elements.

### Summary

I would recommend supporting the following operations:

1. non-key-checking (O(1)) push and extend.
2. key-checking (O(n)) push and extend.
3. `impl FromIterator<Item=(&str, &RawBson)> for RawDocument`

Then document that other modifications should be made either by iterating over the existing document and collecting to a fresh RawDocument, or by deserializing to a more appropriate data structure to perform mutations (a Document or custom type), then serializing back to RawDocument.

src/raw/mod.rs
patrickfreed: >Finally, storing an &'a [u8] forces this type to only be used as a reference type. An usized RawDocumentRef type can also be stored in a Box / Rc / Arc / other smart pointer, and be passed around as an owned type, saving an extra pointer indirection over boxing a RawDocument.

How could we create e.g. a `Box<RawDocumentRef>`? And what's the use case for it?

The first two reasons seem valid, though I'd personally prefer to avoid using `unsafe` / stick with simpler reference based code and just have to deal with the boilerplate of re-implementing all those methods.

Any thoughts on this? @saghm @isabelatkinson

src/raw/props.rs
patrickfreed: Very cool! This seems like a great way to test this.

src/raw/elem.rs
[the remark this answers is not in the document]
saghm: That's a really good point; we actually just added back in support to the library for decoding documents with invalid UTF-8 strings due to a long-standing server bug. I definitely think it's worth giving users a way to deal with data like that in the raw BSON API.

src/raw/elem.rs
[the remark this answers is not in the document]
saghm: Done

src/raw/mod.rs
saghm: Done

patrickfreed (commented):
Thanks for the detailed overview @jcdyer ! I definitely agree that `RawDocument` is definitely not a strict improvement performance-wise for all use-cases. I think the lion's share of the perf benefits will come from within the driver actually, since it should allow us to skip an intermediate serialization step for insertions and potentially reduce a lot of heap allocations in other places. It's unclear if deserialization will improve significantly though, since that requires mostly key lookups.

I also agree with the summary / suggested operations. One additional operation that may be worth including is entry deletion, since that should be pretty straightforward as well (will be O(n) as with all other key-based operations though). This would make it easier for a user to do a replace should they have to, though again we should try to encourage them to use a different type if they plan to do mutations of that sort. I guess they could also trivially implement it via a filter + collect + non-key-checking append, though, so maybe its not necessary.

src/raw/elem.rs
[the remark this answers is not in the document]
saghm: > For well-formed BSON, this difference doesn't actually matter in practice, since we return a copy of the various Copy types anyways from the struct, and the lifetime annotation on the enum guarantees that the RawBson instance itself will not outlive the RawDocument it was derived from, even if it doesn't actually contain any reference to it.

I understand that it doesn't matter in terms of safety, but it still feels a little weird. Like I said, I'm having trouble articulating exactly why it bothers me, but it still does.

src/raw/error.rs
```diff
@@ -0,0 +1,27 @@
+/// An error that occurs when attempting to parse raw BSON bytes.
+#[derive(Debug, PartialEq)]
+pub enum Error {
```
saghm, line 3: As part of refactoring this into a separate submodule, I removed the `Raw` prefix from the `Error` and `Result` types defined here. This is consistent with how we define errors elsewhere in the library (e.g. `bson::de::Error` and `bson::ser::Error`), and users will need to import it with `bson::raw::{Error, Result}` anyhow.

As for RawDocument and friends, I do not think we should remove the prefix here. `raw::Error` and `raw::Result` are essentially semantically equivalent to other `Error` and `Result` types both within this library and externally, `Document` and `RawDocument` are semantically different in significant ways. Put another way, I don't think it would be confusing for readers of code using this library to gloss over what the exact Result or Error type is given that their expectations of how they work will be correct, but glossing over whether a regular Document or a RawDocument is used seems much more dangerous.

src/raw/mod.rs
saghm: I've split what previously existing in the top-level `raw` module into the `doc`, `array`, and `test` (private) submodules, leaving behind only re-exports of the public items from those submodules and a few helper functions to be used by the submodules.

src/raw/props.rs
saghm: See also [`quickcheck`](https://hackage.haskell.org/package/QuickCheck) for info on the Haskell library that popularized this idea

src/raw/mod.rs
saghm: I think a compelling example use case would be to be able to do something like `Cow<RawDocumentRef>` in the same way that one could do `Cow<str>`. By giving `RawDocumentRef` the same semantics as existing owned/borrowed pairs like `Vec<T>`/`[T]` and, `String`/`str`, we allow users to re-use the same patterns that they're used to. I'm not thrilled with the idea of having to include unsafe code in the library, but a single line of unsafe limited to very straightforward casting seems like it wouldn't be the end of the world.

jcdyer: > How could we create e.g. a Box<RawDocumentRef>? And what's the use case for it?

I think, given that `Deref<Target=RawDocumentRef>` is implemented, you can get a `Box<RawDocumentRef>` by using `Box::new(*raw_document)`.  The usecase, particularly for Rc/Arc, is to get multiple ownership semantics without introducing an extra pointer indirection from inside RawDocument.  The usecase would be for a minor performance and size improvement when passing between tasks/threads in an async/multithreaded context, or storing in data structures where you want need multiple ownership.

I also personally find it harder to read struct lifetimes than simple reference lifetimes.

It's definitely a maintenance question for y'all to answer primarily.  One line of unsafe to audit vs maintaining duplicate APIs.

patrickfreed: `Box::new(*raw_document)` doesn't work since `RawDocumentRef` is dynamically sized. That being said, `Cow` does seem to be a pretty good reason to model this type that way. 

If we're going to go this route, we'll need to add a manual `From` implementation for our type, [like they do for other similar reference / slice types](https://doc.rust-lang.org/std/borrow/enum.Cow.html#trait-implementations). Otherwise, creating a `Cow` for our document/reference type can be a little thorny.

src/raw/mod.rs
patrickfreed: sp: "borrwed" -> "borrowed"

src/raw/props.rs
patrickfreed: Does the prop definitions need to be in the actual library? Or can we move them into the `test` module?

src/raw/elem.rs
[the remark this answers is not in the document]
patrickfreed: Well anyways, I think this would be a huge usability / familiarity win for users, so unless we have strong reasons to opt for the struct, I would really prefer for us to go the enum route. As I mentioned before, the paths aren't mutually exclusive, so we could do the struct approach too if we really needed to.

patrickfreed (commented):
It looks like the tests aren't compiling after the module reorganization.

src/raw/elem.rs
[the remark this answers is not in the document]
isabelatkinson: +1 for the enum route; it's a much more intuitive design IMO, and I think it'll be nice to have some consistency between `Document` and `RawDocument` implementations

jcdyer:
Tangential to this, are there any plans to derive(Copy) for ObjectId?  Its current impl is stack allocated, fixed size, and smaller than the pointer for a &[u8] on 64 bit platforms, and would clean up a lot of code.

src/raw/doc.rs
```diff
@@ -0,0 +1,824 @@
...
+/// happen at any time during use.
+///
+/// Iterating over a RawDocument yields either an error or a key-value pair that borrows from the
+/// original document without making any additional allocations.
+///
+/// ```
+/// # use bson::raw::{RawDocument, Error};
+/// let doc = RawDocument::new(b"\x13\x00\x00\x00\x02hi\x00\x06\x00\x00\x00y'all\x00\x00".to_vec())?;
+/// let mut iter = doc.iter();
+/// let (key, value) = iter.next().unwrap()?;
```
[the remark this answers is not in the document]
jcdyer: Yes. The return type is Option<Result<_,_>>.  It could also be .unwrap().unwrap().

saghm:
> Tangential to this, are there any plans to derive(Copy) for ObjectId? Its current impl is stack allocated, fixed size, and smaller than the pointer for a &[u8] on 64 bit platforms, and would clean up a lot of code.

Yeah, I think this probably is worth doing. Feel free to open a separate PR with this change!

jcdyer:
Where does this stand?  From what I can see, it seems to have stalled out again.  Is there anything I can do to help get it moving again?

patrickfreed:
We had to put this on the backburner for a little while to focus on getting the 2.0 release of the driver out as well as a few other internal priorities. We'll be picking this up again in the near term future after the 2.0 stable release is out. Thanks again for your continued patience and all the work you've put into this so far, and sorry about the delay!

patrickfreed:
Hey @jcdyer, just wanted to let you know that we've picked this up again now that the 2.0 has been released. As a first step, I'll be  tidying things up and addressing any of the outstanding comments from the last round of reviews. After that's done, I'll open it up to the team for another review. As with last time, feel free to chime in with any comments as you see fit.

Thanks again for getting the ball rolling on this!

src/raw/elem.rs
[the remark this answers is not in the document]
patrickfreed: Per the latest commits, `RawBson` is now an enum. That being said, there is still room for a struct-style `RawBson` via an `Entry`-like API, for example.

src/raw/test/mod.rs
```diff
@@ -0,0 +1,497 @@
...
+        "symbol": Bson::Symbol(String::from("artist-formerly-known-as")),
+        "javascript_with_scope": Bson::JavaScriptCodeWithScope(JavaScriptCodeWithScope{ code: String::from("console.log(msg);"), scope: doc!{"ok": true}}),
+        "int32": 23i32,
+        "timestamp": Bson::Timestamp(Timestamp { time: 3542578, increment: 0 }),
+        "int64": 46i64,
+        "end": "END",
+    });
+
+    let rawdoc = RawDoc::new(&docbytes).expect("invalid document");
+    let _doc: crate::Document = rawdoc.try_into().expect("invalid bson");
```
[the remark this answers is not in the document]
patrickfreed: good idea, updated to test this.

src/raw/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            message: "expected string to be null-terminated".to_string(),
+        }));
+    }
+
+    // exclude null byte
+    try_to_str(&buf[4..(end - 1)])
+}
+
+fn try_to_str(data: &[u8]) -> Result<&str> {
+    match std::str::from_utf8(data) {
```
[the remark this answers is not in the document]
patrickfreed: updated.

src/raw/error.rs
```diff
@@ -0,0 +1,139 @@
...
+pub struct Error {
+    /// The type of error that was encountered.
+    pub kind: ErrorKind,
+
+    /// They key associated with the error, if any.
+    pub(crate) key: Option<String>,
+}
+
+impl Error {
+    pub(crate) fn new_with_key(key: impl AsRef<str>, kind: ErrorKind) -> Self {
```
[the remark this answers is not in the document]
patrickfreed: updated

src/raw/array.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+}
+
+/// An iterator over borrowed raw BSON array values.
+pub struct RawArrIter<'a> {
+    inner: Iter<'a>,
+}
+
+impl<'a> Iterator for RawArrIter<'a> {
```
isabelatkinson, line 232: One validation we're missing here is checking to make sure that the keys in the array are well-formed (i.e. incrementing integers starting with 0). This could lead to some confusing behavior in which malformed BSON bytes that include an "\x04" tag followed by a regular document are successfully interpreted as an array. I think we should add some checks here to ensure that the document we're dealing with is a proper array-style document.

Cargo.toml
```diff
@@ -55,10 +55,12 @@ serde_bytes = "0.11.5"
 
 [dev-dependencies]
 assert_matches = "1.2"
-serde_bytes = "0.11"
+criterion = "0.3.0"
 pretty_assertions = "0.6.1"
-chrono = { version = "0.4", features = ["serde"] }
 uuid = { version = "0.8.1", features = ["serde", "v4"] }
+proptest = "0.10"
```
isabelatkinson, line 61: looks like this crate is now at 1.0.0 -- should we bump the dependency?

src/raw/doc.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// following invariants:
+    ///   * `data` is at least five bytes long (the minimum for a valid BSON document)
+    ///   * the initial four bytes of `data` accurately represent the length of the bytes as
+    ///     required by the BSON spec.
+    ///   * the last byte of `data` is a 0
+    ///
+    /// Note that the internal structure of the bytes representing the
+    /// BSON elements is _not_ validated at all by this method. If the
+    /// bytes do not conform to the BSON spec, then method calls on
+    /// the RawDocument will return Errors where appropriate.
```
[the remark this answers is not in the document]
patrickfreed: Fixed this and a few of the other non-linkified references to types in this file.

Cargo.toml
patrickfreed: Good catch, fixed.

src/raw/array.rs
patrickfreed: Surprisingly, accepting array documents with messed up indexes is actually required by the bson corpus tests. e.g. the "Single Element array with index set incorrectly to ab" test is included under "valid": https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/array.json#L23

isabelatkinson: Huh, that's interesting.

abr-egn:
> Yeah the name is derived from the `String`/`str` pairing, but it's not clear if that's the right choice. Even within the Rust stdlib this sort of breakdown isn't consistent. For example, there's also `PathBuf` / `Path`, which we could also mimic here via `RawDocumentBuf` / `RawDocument`, if that seems clearer. I couldn't decide myself between the two so I just picked the shorter one, but I think both would work. Any thoughts on names here @abr-egn @isabelatkinson ?

I'm of the opinion that the core types should have been `StringBuf` / `String` personally, so I'd lean that way here as well.  That said, this is a _very_ mild leaning and I don't think it's critical either way.

kmahar (approved):

Issue #226, opened 2021-01-24
Decoding a a document with lossy utf8 conversion


thijsc:
I'm bumping the bson version in the Mongo C driver wrapper to the latest version in thijsc/mongo-rust-driver/pull/58.

We currently decode a few responses from Mongo with lossy utf8 conversion since we found out that sometimes invalid characters kan make it into error responses. We used `bson::decode_document_utf8_lossy` for this.

It looks like `bson::Document::from_reader` is the replacement. I don't see a way to have it use lossy encoding. Is there a way to achieve this?

agolin95:
Tracked in: https://jira.mongodb.org/browse/RUST-648

patrickfreed:
Fixed in #232, and will be included in the next release.

PR #232, opened 2021-02-10
RUST-648 Reintroduce Document decoding w/ lossy UTF-8 conversion

patrickfreed:
RUST-648 and #226 

This PR reintroduces support for decoding a `Document` from a stream of bytes that may contain invalid UTF-8 strings. This is useful when reading raw BSON returned from the server, which in rare cases may invalidly truncate it ([SERVER-24007](https://jira.mongodb.org/browse/SERVER-24007)).

### DIFF
--- a/src/document.rs
+++ b/src/document.rs
@@ -524,8 +524,7 @@ impl Document {
         Ok(())
     }
 
-    /// Attempts to deserialize a `Document` from a byte stream.
-    pub fn from_reader<R: Read + ?Sized>(reader: &mut R) -> crate::de::Result<Document> {
+    fn decode<R: Read + ?Sized>(reader: &mut R, utf_lossy: bool) -> crate::de::Result<Document> {
         let mut doc = Document::new();
 
         let length = read_i32(reader)?;
@@ -550,7 +549,7 @@ impl Document {
                         break;
                     }
 
-                    let (key, val) = deserialize_bson_kvp(cursor, tag, false)?;
+                    let (key, val) = deserialize_bson_kvp(cursor, tag, utf_lossy)?;
                     doc.insert(key, val);
                 }
                 Ok(())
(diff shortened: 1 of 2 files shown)


src/document.rs
```diff
@@ -559,6 +558,21 @@ impl Document {
...
+        Self::decode(reader, false)
+    }
+
+    /// Attempt to deserialize a `Document` that may contain invalid UTF-8 strings from a byte
+    /// stream.
+    ///
+    /// This is mainly useful when reading raw BSON returned from a MongoDB server, which
+    /// in rare cases can contain invalidly truncated strings (https://jira.mongodb.org/browse/SERVER-24007).
+    /// For most use cases, `Document::from_reader` can be used instead.
+    pub fn from_reader_utf8_lossy<R: Read + ?Sized>(reader: &mut R) -> crate::de::Result<Document> {
```
patrickfreed, line 573: If we anticipate there being more options to add here, maybe we should consider `from_reader_with_options`? Or do we think this will be it?

src/tests/modules/serializer_deserializer.rs
```diff
@@ -64,6 +64,21 @@ fn test_serialize_deserialize_utf8_string() {
     assert_eq!(deserialized, doc);
 }
 
+#[test]
+fn test_encode_decode_utf8_string_invalid() {
```
patrickfreed, line 68: This was the existing test that was removed when we eliminated this API.

src/document.rs
saghm: I think it's fine to leave like this. If it becomes necessary, we can add a `with_options` variant in addition to these two.

saghm (approved):

isabelatkinson (approved):

thijsc:
Thanks!

PR #239, opened 2021-02-25
RUST-687 Support deserializing binary of UuidOld

kenkoooo:
Since I have a lot of existing binary data of subtype=0x03, which is a subtype of old UUID, I added support to deserialize such binary data to UUID.


src/serde_helpers.rs
```diff
@@ -215,10 +215,16 @@ pub mod uuid_as_binary {
...
-        if binary.subtype == BinarySubtype::Uuid {
+        if binary.subtype == BinarySubtype::Uuid || binary.subtype == BinarySubtype::UuidOld {
             if binary.bytes.len() == 16 {
                 let mut bytes = [0u8; 16];
                 bytes.copy_from_slice(&binary.bytes);
+
+                if binary.subtype == BinarySubtype::UuidOld {
+                    bytes[0..8].reverse();
+                    bytes[8..16].reverse();
+                }
```
jcdyer, line 226: I'm not sure baking this assumption into the deserializer is the right approach.  The correct logic for this depends on what client created the data, which is why the new UUID subtype was created.  The logic here is only correct for data created by the Java client.  

Specifics on the correct deserialization rules are here: https://github.com/mongodb/mongo-python-driver/blob/master/bson/binary.py#L251-L262

Since the data format itself provides no way to know which format to use, this would have to be handled either by providing an argument to the deserializer itself (probably not a great idea), or deserializing to a target type that understands the appropriate byte order.  One alternative would be to create separate types for each format: 

```rust
struct PythonLegacyUuid(uuid::Uuid);
struct JavaLegacyUuid(uuid::Uuid);
struct CSharpLegacyUuid(uuid::Uuid);
```

and then impl Deref<Target=uuid::Uuid>` (& also DerefMut).

Another alternative would be to create a single LegacyUuid type that stores the bytes untouched, and allows the user to choose the appropriate ordering by method call.
```rust
struct LegacyUuid([u8;16]);
impl LegacyUuid {
    fn java_uuid(&self) -> Uuid { todo!() }
    fn python_uuid(&self) -> Uuid { todo!() }
    fn csharp_uuid(&self) -> Uuid { todo!() }
}
```

These types can be created outside of the bson create, but I would also love to see one or both alternatives within as well, since they solely exist to deal with legacy mongodb issues.

agolin95:
Tracked in https://jira.mongodb.org/browse/RUST-687

src/serde_helpers.rs
kenkoooo: Thank you for your suggestion. I will add `JavaLegacyUuid`, `PythonLegacyUuid`, and `CSharpLegacyUuid` as you suggested.

patrickfreed (commented):
Hi @kenkoooo! Thank you or the PR, and thank you @jcdyer for bringing up the UUID spec! I think this implementation looks great so far, but I think we'll want to go for a slightly different API to be consistent with how we currently handle regular subtype 4 UUIDs.

By that I mean, instead of separate types for each of the different UUIDs, can we change this to introduce separate `serde_helper` modules for each? This matches the current UUID subtype 4 API.

e.g.
```rust
#[derive(Serialize, Deserialize)]
struct MyUuids {
    // This is the existing UUID helper
    #[serde(with = "serde_helpers::uuid_as_binary")]
    regular: Uuid,

    // Below would be the new ones for each legacy type
    #[serde(with = "serde_helpers::uuid_as_java_legacy_binary")]
    java_legacy: Uuid,

    #[serde(with = "serde_helpers::uuid_as_python_legacy_binary")]
    python_legacy: Uuid,

    #[serde(with = "serde_helpers::uuid_as_c_sharp_legacy_binary")]
    c_sharp_legacy: Uuid,
}
```
Check out [`serde_helpers.rs`](https://github.com/mongodb/bson-rust/blob/master/src/serde_helpers.rs#L197) for an example implementation for the subtype 4 version.

For datetimes, we originally did go the route of introducing a new type (`bson::DateTime`) which wraps the regular `chrono::DateTime` and has special `Serialize` and `Deserialize` implementations, but users tended to find it confusing / hard to use / hard to discover. As a result, we opted to take the `serde_helper` approach for `Uuid` instead as it is a bit more explicit / configurable, so I think we should continue to use that pattern for these legacy variants too.

agolin95:
Hey @kenkoooo just wanted to drop a comment to make sure you saw Patrick's feedback, are these changes you would like to make or would you prefer we push up changes to this branch ourselves? Thanks for all the hard work!

kenkoooo:
@agolin95 Thank you for your confirmation. Let me try to implement as @patrickfreed suggested.

kenkoooo:
Hi, sorry for the very slow work. I update my PR as @patrickfreed suggested. Could you take a look into it? Thanks!

patrickfreed (commented):
These changes look great! I only have a few minor comments:
- Can we update the module names to include `_0_8_` after `uuid`? We're versioning these helpers in order to support multiple semver-breaking versions of `uuid` as they come out. See #255 for more info (you can leave the exising `uuid_as_binary` one alone, that'll be updated in that PR)
- Can we add top level re-exports of the `serialize` and `deserialize` functions similar to what we have for the other ones? Check the top of the file for the pattern there.

Thanks again for your contribution!

kenkoooo:
@patrickfreed Thanks for your review! I updated it.

patrickfreed:
Looks like rustfmt needs to be run, but other than that it looks good to me! (note: we use the nightly version of rustfmt).

tagging in @isabelatkinson for review as well.

isabelatkinson (approved):
LGTM once the formatting errors are fixed, thanks for your contribution!

kenkoooo:
@isabelatkinson @patrickfreed Thanks! I reformat by `cargo +nightly fmt`.

patrickfreed (approved):
LGTM! Thank you for your contribution @kenkoooo !

PR #242, opened 2021-03-21
minor: fix readme example using private function

bahildebrand:
The function `decode` is private for `Document`, so will cause a
compiler if following the example. Switch to the `from_reader` function.


agolin95:
Tracked in https://jira.mongodb.org/browse/RUST-712

saghm:
Thanks for the fix!

PR #247, opened 2021-04-08
RUST-283 Replace linked-hash-map with indexmap

patrickfreed:
RUST-283

This PR updates `Document` to wrap an `IndexMap` (from the `indexmap` crate) instead of `LinkedHashMap` from the now defunct `linked-hash-map` crate.

As part of that, I also updated the `Entry` API to match that of `IndexMap` / `std::HashMap`, since ours used to be slightly different.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ chrono = "0.4"
 rand = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order"] }
-linked-hash-map = "0.5.3"
+indexmap = "1.6.2"
 hex = "0.4.2"
 decimal = { version = "2.1.0", default_features = false, optional = true }
 base64 = "0.13.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ decimal128 = ["decimal"]
 name = "bson"
 
 [dependencies]
+ahash = "0.7.2"
 chrono = "0.4"
 rand = "0.7"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/document.rs
+++ b/src/document.rs
@@ -10,9 +10,7 @@ use std::{
 };
 
 use chrono::{DateTime, Utc};
-
-use linked_hash_map::{self, LinkedHashMap};
-
+use indexmap::IndexMap;
 use serde::de::{self, Error, MapAccess, Visitor};
 
 #[cfg(feature = "decimal128")]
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -39,11 +39,12 @@ decimal128 = ["decimal"]
 name = "bson"
 
 [dependencies]
+ahash = "0.7.2"
```
patrickfreed, line 42: The `ahash` crate provides a fast yet safe hashing algorithm. In fact, I think it's now the default one used by `std::HashMap`: https://github.com/rust-lang/hashbrown/blob/master/Cargo.toml. In 2.0, `indexmap` looks like they will be updating their default to use it, so we can remove this dependency then (https://github.com/bluss/indexmap/issues/135)

patrickfreed:
The clippy failures appear to be on master too, I'll fix them separately.

src/document.rs
```diff
@@ -575,37 +565,79 @@ impl Document {
     }
 }
 
-pub struct Entry<'a> {
-    inner: linked_hash_map::Entry<'a, String, Bson>,
+/// A view into a single entry in a map, which may either be vacant or occupied.
+///
+/// This enum is constructed from the entry method on HashMap.
+pub enum Entry<'a> {
```
patrickfreed, line 571: I updated this to match both `IndexMap` and `std::HashMap`'s `Entry` API. I updated just the parts that are breaking and left the remaining bits to future, additive work. I'll file a ticket once this is merged for us to remember to do the rest of it. It's basically just adding a few methods to `OccupiedEntry` and `VacantEntry`.

kmahar (commented):
couple questions, but generally LGTM!

src/document.rs
patrickfreed: Yep, `OccupiedEntry` will have a `fn get() -> &Bson` on it which will return the value.

In the old API (and the new one, for now), it looks like you had to use `or_insert` or `or_insert_with` to get the value out from an entry, inserting into the into it if it wasn't occupied. It was definitely incomplete and missing a lot of the useful parts of the `std::HashMap` one. 

The `Entry` API looks mostly useful for mutating / inspecting the map itself without having to compute the hash of an input multiple times. I remember we were doing that in the driver and a user actually submitted a PR to update it to use `entry` and be a little more efficient.
```rust
let mut map = HashMap::new();
// old
if map.get("key").is_none() {
    map.insert("key".to_string(), "value".to_string());
}

// new
if let Entry::Vacant(entry) = map.entry("key") {
    entry.insert("value".to_string());
}
```

kmahar: ahh I see. thanks for explanation! cool, seems like the new API should be a nice usability improvement then.

kmahar (approved):
lgtm!

PR #248, opened 2021-04-16
RUST-747 Add serde helper to deserialize hex string from ObjectId

moka491:
This PR tries to solve an issue I had in a project recently, where 
serializing a String as an ObjectId works fine using the already provided serde_helper function,
but (implicitly) deserializing said ObjectId back to a String would fail.

You can quickly test this by setting up a project with
```rust
[dependencies]
bson = "1.2.2"
serde = "1.0.125"
```

and 

```rust
use bson::serde_helpers::*;
use serde::{Deserialize, Serialize};

#[derive(Debug, Serialize, Deserialize)]
struct Test {
    #[serde(serialize_with = "serialize_hex_string_as_object_id")]
    id: String,
}

fn main() {
    let test_struct = Test {
        id: "6079529e00b3405c004ea780".to_string(),
    };

    let doc = bson::to_document(&test_struct).unwrap();
    println!("{:#?}", &doc);

    let deser_struct: Test = bson::from_document(doc).unwrap();
    println!("{:#?}", &deser_struct);
}
```

Above code will panic with a `DeserializationError { message: "invalid type: map, expected a string" }` at the second `unwrap()`.

In my project, I solved this using my own deserializer function,
which I then implemented similarly in serde_helpers:
```rust
pub fn deserialize_object_id_to_hex_string<'de, D>(deserializer: D) -> Result<String, D::Error>
where
    D: Deserializer<'de>,
{
    let object_id = ObjectId::deserialize(deserializer)?;
    Ok(object_id.to_hex())
}
```

Since this is my first Rust PR, I would like to ask for assistance in two things:
- the previously provided function `serialize_hex_string_as_object_id()` already uses the name I would like to use as the name for the exported function, to follow the naming convention of the rest. (This is also why the PR build will fail for now). However, renaming the existing function would be a breaking change. How would we deal with this?

- I marked the existing function as deprecated, I assume this would need an additional version argument. The function could also be removed altogether of course, since it was moved into the newly created module.

### DIFF
--- a/src/serde_helpers.rs
+++ b/src/serde_helpers.rs
@@ -18,6 +18,10 @@ pub use iso_string_as_bson_datetime::{
     deserialize as deserialize_iso_string_from_bson_datetime,
     serialize as serialize_iso_string_as_bson_datetime,
 };
+pub use hex_string_as_object_id::{
+    deserialize as deserialize_hex_string_from_object_id,
+    serialize as serialize_hex_string_as_object_id,
+};
 pub use timestamp_as_u32::{
     deserialize as deserialize_timestamp_from_u32,
     serialize as serialize_timestamp_as_u32,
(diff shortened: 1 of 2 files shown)


patrickfreed:
Hi @moka491, thanks for the PR! This seems like it would be a nice addition to the serde helpers.

Regarding the name conflict, since the signature of the one included in the new module is the same and is being re-exported under the same name, I think it's safe to just delete the old one without breaking compatibility. That being said, compatibility isn't a huge concern since this will be landing in 2.0 of the crate anyways.

Also, could you add a test for this behavior into `src/tests/serde.rs`? Thanks!

moka491:
Alright, this should be good now, I removed the duplicate function and added a few lines to the respective test 👍

patrickfreed:
Code changes look good to me, running a CI patch now.

patrickfreed:
The linter tests appear to have failed, could you make sure to run clippy and rustfmt first? You can use the scripts in the `.evergreen` directory of the repository to check if the code is formatter properly or not.

moka491:
It should work now. I fixed the linting errors I caused myself, but there's still some that are even on `master`, and I wasn't sure I should fix them here.

patrickfreed (approved):
LGTM! Thanks for your contribution!

And yeah, those are existing failures due to new features in clippy or something. We will fix those separately.

isabelatkinson (approved):
Thanks for your contribution, @moka491 !

PR #250, opened 2021-04-17
add deserialize_object_id_as_hex_string function to serde helpers

icub3d:
This function is the reverse of `serialize_hex_string_as_object_id`. It
helps with finding structs from a collection similar to how
`serialize_hex_string_as_object_id` helps with putting them into a
collection.


patrickfreed:
Another user actually opened a PR for the same change (#248) already, so I'll be closing this one as a duplicate of that. Thanks anyways for your contribution!

PR #252, opened 2021-04-26
RUST-738 Implement Clone on error types

patrickfreed:
RUST-738

This PR implements `Clone` on the various error types in the crate. It also updates a few of the error cases to include more information and removes redundant ones.

Once this PR merges, I'll open a companion one on the driver to remove the `Arc` wrappers around the BSON cases.

### DIFF
--- a/src/de/error.rs
+++ b/src/de/error.rs
@@ -1,15 +1,15 @@
-use std::{error, fmt, fmt::Display, io, string};
+use std::{error, fmt, fmt::Display, io, string, sync::Arc};
 
 use serde::de::{self, Unexpected};
 
 use crate::Bson;
 
 /// Possible errors that can arise during decoding.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 #[non_exhaustive]
 pub enum Error {
     /// A [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html) encountered while deserializing.
-    IoError(io::Error),
+    IoError(Arc<io::Error>),
 
     /// A [`std::string::FromUtf8Error`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html) encountered
     /// while decoding a UTF-8 String from the input data.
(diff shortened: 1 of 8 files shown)


src/de/error.rs
```diff
@@ -31,11 +31,14 @@ pub enum Error {
     /// The end of the BSON input was reached too soon.
     EndOfStream,
 
-    /// An invalid timestamp was encountered while decoding.
-    InvalidTimestamp(i64),
+    /// An invalid datetime was encountered while decoding.
+    InvalidDateTime {
```
patrickfreed, line 35: This error variant was a bit confusing, since it was referring to a datetime rather than BSON timestamp. I updated the name and included some extra information.

src/de/error.rs
```diff
@@ -31,11 +31,14 @@ pub enum Error {
...
+        /// The key at which an unexpected/unsupported datetime was encountered.
+        key: String,
 
-    /// An ambiguous timestamp was encountered while decoding.
-    AmbiguousTimestamp(i64),
```
patrickfreed, line 38: This case was never actually reachable as per [the docs for `chrono`](https://docs.rs/chrono/0.4.19/chrono/offset/trait.TimeZone.html#method.timestamp_opt).

patrickfreed:
Tagging in @kmahar as well for the error API changes.

kmahar (approved):

src/ser/error.rs
```diff
@@ -30,16 +30,16 @@ pub enum Error {
     ///
     /// To allow serialization of unsigned integers as signed integers, enable the "u2i" feature
     /// flag.
-    UnsupportedUnsignedType,
```
isabelatkinson, line 33: Can we update the comment on this variant as well to point to the serde helpers along with `u2i`? Something along the lines of what you added to the display implementation below

isabelatkinson (commented):
looks good, just one small documentation request

src/ser/error.rs
patrickfreed: Good idea, updated. I also updated the docstring and the error message to be a little more general in case users might want to serialize their unsigned integers as doubles or something.

isabelatkinson (approved):

PR #254, opened 2021-05-04
RUST-756 Replace compat::u2f with serde helpers

patrickfreed:
RUST-756

This replaces the `compat::u2f` module with a few `serde_helpers` modules instead to match the pattern of our other helpers.


src/serde_helpers.rs
```diff
@@ -21,13 +21,18 @@ pub use iso_string_as_bson_datetime::{
     serialize as serialize_iso_string_as_bson_datetime,
 };
 pub use timestamp_as_u32::{
-    deserialize as deserialize_timestamp_from_u32, serialize as serialize_timestamp_as_u32,
+    deserialize as deserialize_timestamp_from_u32,
+    serialize as serialize_timestamp_as_u32,
 };
+pub use u32_as_f64::{deserialize as deserialize_u32_from_f64, serialize as serialize_u32_as_f64};
```
patrickfreed, line 27: I opted to only include helpers for u32 and u64 since I doubt any of the smaller oners (`u8` or `u16`) are used very often. If someone asks for them we can add them back.

src/serde_helpers.rs
```diff
@@ -59,6 +64,87 @@ pub fn serialize_u64_as_i64<S: Serializer>(val: &u64, serializer: S) -> Result<S
...
+            Err(de::Error::custom(format!(
+                "cannot convert f64 (BSON double) {} to u64",
+                f
+            )))
+        }
+    }
+
+    /// Serializes a u64 as an f64 (BSON double). Errors if an exact conversion is not possible.
+    pub fn serialize<S: Serializer>(val: &u64, serializer: S) -> Result<S::Ok, S::Error> {
+        if val < &u64::MAX && *val == *val as f64 as u64 {
```
patrickfreed, line 137: The comparison to `u64::MAX` is used here since `val as f64 as u64` works for `u64::MAX` even though the `f64` created by it is not equivalent to `u64::MAX`:
```
u64::MAX: 18446744073709551615
u64::MAX as f64: 18446744073709552000
u64::MAX as f64 as u64: 18446744073709551615
```

src/serde_helpers.rs
isabelatkinson: sounds reasonable, fine by me

isabelatkinson (approved):

PR #255, opened 2021-05-10
RUST-765 Ensure API meets Rust API guidelines

patrickfreed:
RUST-765 / RUST-789 / RUST-788 / RUST-739

This PR makes a number of breaking changes to the API to ensure it meets the [Rust API Guidelines checklist](https://rust-lang.github.io/api-guidelines/checklist.html).

Notable changes:
- Restructured or eliminated public dependencies on unstable crates to be compatible with future major releases of those crates
  - `chrono`:
    -  `Bson::DateTime(chrono::DateTime<Utc>)` => `Bson::DateTime(crate::DateTime)`
    - removed `pub` from the wrapped `chrono::DateTime` in `bson::DateTime`
    - `serde_helpers::chrono_datetime_as_bson_datetime` => `serde_helpers::chrono_0_4_datetime_as_bson_datetime`
    - `ObjectId::timestamp` now returns a `crate::DateTime`
  - `uuid`:
    - `serde_helpers::uuid_as_binary` => `serde_helpers::uuid_0_8_as_binary`
  - `hex`: rewrote `oid::Error` to exclude it
- All `Document` getters now accept `impl AsRef<str>` instead of `&str`
- Updated `ObjectId`'s constructors to match the names of [`uuid::Uuid`](https://docs.rs/uuid/0.8.2/uuid/struct.Uuid.html#method.from_bytes)
- Attempting to serialize a `DateTime` with sub-millisecond precision to BSON data will result in an error
- Renamed error variants to be less redundant and more consistent
  - `de::Error`:
    - `Error:Io(...)` => `Error::Io(...)`
    - `Error::FromUtf8Error(...)` => `Error::InvalidUtf8String(...)`
    - `Error::SyntaxError` _removed_, replaced by `Error::DeserializationError`
  - `ser::Error`:
    - `Error::IoError(...)` => `Error::Io(...)`

### DIFF
--- a/examples/serialize.rs
+++ b/examples/serialize.rs
@@ -8,7 +8,7 @@ fn main() {
 
     let arr = vec![
         Bson::String("blah".to_string()),
-        Bson::DateTime(chrono::Utc::now()),
+        Bson::DateTime(chrono::Utc::now().into()),
         Bson::ObjectId(oid::ObjectId::from_bytes([
             1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12,
         ])),
--- a/examples/serialize.rs
+++ b/examples/serialize.rs
@@ -9,7 +9,7 @@ fn main() {
     let arr = vec![
         Bson::String("blah".to_string()),
         Bson::DateTime(chrono::Utc::now()),
-        Bson::ObjectId(oid::ObjectId::with_bytes([
+        Bson::ObjectId(oid::ObjectId::from_bytes([
             1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12,
         ])),
     ];
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -278,7 +278,7 @@ impl From<u64> for Bson {
 
 impl From<[u8; 12]> for Bson {
     fn from(a: [u8; 12]) -> Bson {
-        Bson::ObjectId(oid::ObjectId::with_bytes(a))
+        Bson::ObjectId(oid::ObjectId::from_bytes(a))
     }
 }
 
@@ -607,7 +607,7 @@ impl Bson {
         match keys.as_slice() {
             ["$oid"] => {
                 if let Ok(oid) = doc.get_str("$oid") {
-                    if let Ok(oid) = ObjectId::with_string(oid) {
+                    if let Ok(oid) = ObjectId::parse_str(oid) {
                         return Bson::ObjectId(oid);
                     }
                 }
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -21,10 +21,7 @@
 
 //! BSON definition
 
-use std::{
-    fmt::{self, Debug, Display},
-    ops::{Deref, DerefMut},
-};
+use std::fmt::{self, Debug, Display};
 
 use chrono::{Datelike, SecondsFormat, TimeZone, Utc};
 use serde_json::{json, Value};
@@ -68,7 +65,7 @@ pub enum Bson {
     /// [ObjectId](http://dochub.mongodb.org/core/objectids)
     ObjectId(oid::ObjectId),
     /// UTC datetime
-    DateTime(chrono::DateTime<Utc>),
+    DateTime(crate::DateTime),
     /// Symbol (Deprecated)
     Symbol(String),
     /// [128-bit decimal floating point](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst)
@@ -290,7 +287,7 @@ impl From<oid::ObjectId> for Bson {
 
 impl From<chrono::DateTime<Utc>> for Bson {
     fn from(a: chrono::DateTime<Utc>) -> Bson {
-        Bson::DateTime(a)
+        Bson::DateTime(DateTime(a))
     }
 }
 
@@ -373,15 +370,15 @@ impl Bson {
                 })
             }
             Bson::ObjectId(v) => json!({"$oid": v.to_hex()}),
-            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.year() <= 99999 => {
-                let seconds_format = if v.timestamp_subsec_millis() == 0 {
+            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.0.year() <= 99999 => {
+                let seconds_format = if v.0.timestamp_subsec_millis() == 0 {
                     SecondsFormat::Secs
                 } else {
                     SecondsFormat::Millis
                 };
 
                 json!({
-                    "$date": v.to_rfc3339_opts(seconds_format, true),
+                    "$date": v.0.to_rfc3339_opts(seconds_format, true),
                 })
             }
             Bson::DateTime(v) => json!({
@@ -539,15 +536,15 @@ impl Bson {
                     "$oid": v.to_string(),
                 }
             }
-            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.year() <= 99999 => {
-                let seconds_format = if v.timestamp_subsec_millis() == 0 {
+            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.0.year() <= 99999 => {
+                let seconds_format = if v.0.timestamp_subsec_millis() == 0 {
                     SecondsFormat::Secs
                 } else {
                     SecondsFormat::Millis
                 };
 
                 doc! {
-                    "$date": v.to_rfc3339_opts(seconds_format, true),
+                    "$date": v.0.to_rfc3339_opts(seconds_format, true),
                 }
             }
             Bson::DateTime(v) => doc! {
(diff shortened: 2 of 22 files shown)


src/bson.rs
```diff
@@ -973,76 +970,78 @@ impl Timestamp {
...
-        if num_millis < 0 {
-            num_secs -= 1;
-            num_millis += 1000;
-        };
+pub struct DateTime(chrono::DateTime<Utc>);
 
-        Utc.timestamp(num_secs, num_millis as u32 * 1_000_000)
-            .into()
+impl crate::DateTime {
+    pub(crate) fn from_millis(date: i64) -> Self {
```
patrickfreed, line 1014: I thought about making this public, but apparently too-large values here can cause panics (RUST-799), so I just left it private for now until we decide how / if we want to tackle this publicly.

src/bson.rs
```diff
@@ -973,76 +970,78 @@ impl Timestamp {
...
+impl From<crate::DateTime> for chrono::DateTime<Utc> {
     fn from(utc: DateTime) -> Self {
         utc.0
     }
 }
 
-impl From<chrono::DateTime<Utc>> for DateTime {
-    fn from(x: chrono::DateTime<Utc>) -> Self {
-        DateTime(x)
+impl<T: chrono::TimeZone> From<chrono::DateTime<T>> for crate::DateTime {
```
patrickfreed, line 1042: I updated this to be more general since it could be a little annoying to convert from `DateTime`s with other timezones otherwise.

src/serde_helpers.rs
```diff
@@ -289,20 +289,20 @@ pub mod hex_string_as_object_id {
...
+/// # use bson::serde_helpers::uuid_0_8_as_binary;
 /// #[derive(Serialize, Deserialize)]
 /// struct Item {
-///     #[serde(with = "uuid_as_binary")]
+///     #[serde(with = "uuid_0_8_as_binary")]
 ///     pub id: Uuid,
 /// }
 /// ```
-pub mod uuid_as_binary {
+pub mod uuid_0_8_as_binary {
```
patrickfreed, line 305: Note that with #239, each major version of `uuid` will come with 4 serde helper modules, which will pollute the namespace a bit. That doesn't seem too bad to me but we could consider gating it behind a feature flag if we want to avoid that.

examples/serialize.rs
```diff
@@ -8,7 +8,7 @@ fn main() {
 
     let arr = vec![
         Bson::String("blah".to_string()),
-        Bson::DateTime(chrono::Utc::now()),
+        Bson::DateTime(chrono::Utc::now().into()),
```
kmahar, line 11: maybe not technically possible, but figured I'd ask in case there's some way to be clever, with this change is there any way around users having to add the `into()` calls? (like maybe you can call that for them automatically in a constructor or something to avoid the clunkiness?)

on the other hand this sounds vaguely familiar to a recent decision we made.... somewhere... to avoid hiding allocations from the user, which I think we would be doing if we did this. so maybe it's good to make them know a conversion is happening explicitly

src/serde_helpers.rs
kmahar: I don't feel very strongly either way

src/bson.rs
```diff
@@ -973,76 +970,78 @@ impl Timestamp {
...
+/// let chrono_dt: chrono::DateTime<Utc> = "2014-11-28T12:00:09Z".parse()?;
+/// let bson_dt: bson::DateTime = chrono_dt.into();
+/// let back_to_chrono: chrono::DateTime<Utc> = bson_dt.into();
+/// # Ok(())
+/// # }
+/// ```
 ///
-/// Just a helper for convenience
+/// This type differs from [`chrono::DateTime`] in that it serializes to and deserializes from a
+/// BSON datetime rather than an ISO-8601 formatted string. This means that it will serialize to and
```
kmahar, line 989: what does encoding as an extended JSON object entail in a non-BSON, non-extended JSON format? using the extJSON key names?

kmahar (commented):
all seems pretty good, just some minor questions/comments

examples/serialize.rs
patrickfreed: We could introduce a `Bson::from_chrono_0_4_datetime` constructor method which would avoid the need for an `.into()` call, but that's probably less ergonomic than using the `From`/`Into` trait.

An alternative to that would be to use feature flags to dictate which chrono version we're using, and then depending on that flag just have the `chrono::DateTime` directly in the enum variant. I only mentioned this briefly in the doc though I probably should have included a more detailed example.
```rust
pub enum Bson {
     // other cases ...

     #[cfg(feature = "chrono_0_4")]
     DateTime(chrono_0_4::DateTime),

     #[cfg(feature = "chrono_0_5")]
     DateTime(chrono_0_5::DateTime),

    // rest of cases...
}
```

The problem here is we'll require that a user select a feature, which could be kind of annoying. We could default to the current implementation in this PR if no feature is specified, though then we'd be presenting two slightly different APIs that may make documentation or examples hard. A better default would probably be to enable the "chrono_0_4" feature by default and just add a new "chrono" feature if/when chrono does 1.0. Once that happens the `Cargo.toml` invocation for the driver would potentially be kind of annoying:

```toml
mongodb = { default-features = false, features = ["tokio-runtime", "bson/chrono"], version = "2" }
```
For async-std users this won't be much of a difference though.

src/bson.rs
patrickfreed: Yep, so the way the crate implemented `Serialize` / `Deserialize` for the various types was to do BSON in BSON and extended JSON (or its equivalent) for all other serialization formats (This is something I suggested we try in Swift a while back). Users have expressed a decent amount of frustration around this (particularly around DateTime), though the serde helpers have helped a lot. I added this comment in here to further help clarify the situation, though I agree that part is a bit confusing. Updated the wording to be a bit more specific.

Given that we're breaking API, we could consider changing this approach, though I think that would probably require a lot of work. We could additively support alternate default serialization formats in the future via feature flags too, so it's not completely necessary to do it now.

src/ser/error.rs
```diff
@@ -17,6 +17,9 @@ pub enum Error {
         key: Bson,
     },
 
+    /// Attempted to serialize a sub-millisecond precision datetime.
+    SubMillisecondPrecisionDateTime(crate::DateTime),
```
patrickfreed, line 18: Does this error name seem okay? it's a little verbose but I couldn't come up with anything much better.

src/serde_helpers.rs
isabelatkinson: Yeah I don't think that's a huge deal, I'm fine leaving as-is.

src/oid.rs
```diff
@@ -105,37 +116,52 @@ impl ObjectId {
         buf[COUNTER_OFFSET..(COUNTER_SIZE + COUNTER_OFFSET)]
             .clone_from_slice(&counter[..COUNTER_SIZE]);
 
-        ObjectId::with_bytes(buf)
+        ObjectId::from_bytes(buf)
     }
 
     /// Constructs a new ObjectId wrapper around the raw byte representation.
-    pub fn with_bytes(bytes: [u8; 12]) -> ObjectId {
+    pub const fn from_bytes(bytes: [u8; 12]) -> ObjectId {
```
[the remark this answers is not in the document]
patrickfreed: Nope, though there probably should be some mention of it there. I think it was quite limited for a while, but each release of Rust adds more things that can be done in `const`. In this case, I used it to match [`uuid::Uuid::from_bytes`](https://docs.rs/uuid/0.8.2/uuid/struct.Uuid.html#method.from_bytes).

src/ser/error.rs
kmahar: agree it's fine. might be worth expanding the comment a little to clarify _why_ that's an error though, e.g. adding "...which BSON does not support" or something like that?

examples/serialize.rs
kmahar: I think the initial decision we made (solution 1 over 2) still seems better from a stability perspective, since as you mention in the doc driver 2.0 may outlive chrono 0.4. given that, I think the minor ergonomic tradeoff of calling `From`/`Into` seems fine, and I agree a separate BSON constructor to avoid the call probably doesn't provide a ton of value

src/bson.rs
kmahar: ah I see. yeah, I don't think we should try to tackle this now given there are ways to work around it, but seems like something good to keep in mind for the future.

kmahar (approved):
couple minor comments (sorry I screwed up and forgot to batch...) but don't need to re-review, lgtm

kmahar:
oh and extremely small nit, in your PR title (which I think will become the commit message by default?): "guideliness" -> "guidelines"

isabelatkinson (approved):

src/bson.rs
patrickfreed: Filed RUST-801.

src/ser/error.rs
patrickfreed: Good idea, added.

PR #257, opened 2021-05-20
RUST-815 Truncate to millisecond precision in `bson::DateTime`'s `From<chrono::DateTime>`

vkill:
Hi, there

The auto remove microseconds and nanoseconds feature was removed from This [commit](https://github.com/mongodb/bson-rust/commit/8a92a568ba9de1192e8fb4217ae541c98b335cad#diff-d6adfba5b9892574146927e3f707649f3b601313211502ff5ae3c796f83dcee4L160). I think should add it in the `From<chrono::DateTime<Utc>>` trait, please review it. thank you.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -1019,12 +1019,6 @@ impl crate::DateTime {
     pub fn timestamp_millis(&self) -> i64 {
         self.0.timestamp_millis()
     }
-
-    /// If this DateTime is sub-millisecond precision. BSON only supports millisecond precision, so
-    /// this should be checked before serializing to BSON.
-    pub(crate) fn is_sub_millis_precision(&self) -> bool {
-        self.0.timestamp_subsec_micros() % 1000 != 0
-    }
 }
 
 impl Display for crate::DateTime {
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -23,7 +23,7 @@
 
 use std::fmt::{self, Debug, Display};
 
-use chrono::{Datelike, SecondsFormat, TimeZone, Utc};
+use chrono::{Datelike, NaiveDateTime, SecondsFormat, TimeZone, Utc};
 use serde_json::{json, Value};
 
 pub use crate::document::Document;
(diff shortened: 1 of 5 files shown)


vkill:
Tracked in https://jira.mongodb.org/browse/RUST-815

src/bson.rs
```diff
@@ -1041,7 +1041,12 @@ impl From<crate::DateTime> for chrono::DateTime<Utc> {
 
 impl<T: chrono::TimeZone> From<chrono::DateTime<T>> for crate::DateTime {
     fn from(x: chrono::DateTime<T>) -> Self {
-        DateTime(x.with_timezone(&Utc))
+        let dt = x.with_timezone(&Utc);
+
+        DateTime(chrono::DateTime::<Utc>::from_utc(
```
patrickfreed, line 1046: Now that we're ensuring all `bson::DateTime`s are millisecond precision, can we remove the `ser::Error::SubMillisecondPrecisionDateTime` error case and associated error checking logic during BSON serialization?

patrickfreed (commented):
Thanks for the PR! This does seem like a better way of handling these issues than the existing solution in 2.0.0-beta. I only have one suggestion to remove some of the now-redundant error handling.

It also looks like the rustfmt linter in our CI is failing, so make sure to use the nightly version of rustfmt when formatting the code. You can verify it will pass our linter by running the `.evergreen/check-rustfmt.sh` script.

src/bson.rs
vkill: Yes, I have removed it.

vkill:
Please review again. thank you.

max-frai:
Yes, please. It broke everything with mongodb serialization.

patrickfreed (approved):
LGTM! Thanks again for contributing this fix, and sorry for any breakages the original change caused!

Tagging in @isabelatkinson for review as well.

isabelatkinson (approved):
Thanks for your contribution!

PR #260, opened 2021-06-02
RUST-838 Improve bson::DateTime now() performance

pymongo:
link to jira: [RUST-838](https://jira.mongodb.org/browse/RUST-838)

## DateTime::now()

https://github.com/mongodb/bson-rust/blob/2e6bb8f4763889728cfe4339d63b817a5c99af5d/src/datetime.rs#L72-L74

because chrono::Utc::now() call SystemTime::now() and NaiveDateTime::from_timestamp

bson::DateTime::now() use SystemTime::now() can prevent **extra performance overhead** on `NaiveDateTime::from_timestamp`

```
  pub fn now() -> DateTime<Utc> {
      let now =
          SystemTime::now().duration_since(UNIX_EPOCH).expect("system time before Unix epoch");
      let naive = NaiveDateTime::from_timestamp(now.as_secs() as i64, now.subsec_nanos() as u32);
      DateTime::from_utc(naive, Utc)
  }
```

chrono source: https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/offset/utc.rs#L47-L51

## confuse on feature = "chrono-0_4"

since bson use chrono::DateTime::parse_from_rfc3339 in serde, so chrono must a required feature

libraray usally use feature to tell user this is a optional dependecy

I guess `feature = "chrono-0_4"` means bson's chrono is a optional dependency

But these code is just confuse me:

https://github.com/mongodb/bson-rust/blob/2e6bb8f4763889728cfe4339d63b817a5c99af5d/src/datetime.rs#L76-L86

So I think feature = "chrono-0_4" is useless because chrono is a required dependency

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -70,7 +70,7 @@ impl crate::DateTime {
 
     /// Returns a [`DateTime`] which corresponds to the current date and time.
     pub fn now() -> DateTime {
-        Self::from_chrono(Utc::now())
+        Self::from_system_time(SystemTime::now())
     }
 
     #[cfg(not(feature = "chrono-0_4"))]
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -163,7 +163,7 @@ impl ObjectId {
     }
 
     /// Returns the raw byte representation of an ObjectId.
-    pub fn bytes(&self) -> [u8; 12] {
+    pub const fn bytes(&self) -> [u8; 12] {
         self.id
     }
 
--- a/src/ser/error.rs
+++ b/src/ser/error.rs
@@ -14,7 +14,7 @@ pub enum Error {
     /// A key could not be serialized to a BSON string.
     InvalidDocumentKey(Bson),
 
-    /// A general error that ocurred during serialization.
+    /// A general error that occurred during serialization.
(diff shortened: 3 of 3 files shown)


patrickfreed (approved):
Code changes look good to me, thanks for your contribution! I did a quick benchmark and it looks to be almost twice as fast:
```
test tests::bench_chrono ... bench:          46 ns/iter (+/- 20)
test tests::bench_native ... bench:          28 ns/iter (+/- 1)
```

Regarding the `"chrono-0_4"` feature flag, the goal of it is to ensure a stable API rather than reduce dependencies. `bson` is a stable crate, but `chrono` is not, which means we can't safely include it in our default public API without a serious risk of needing to release another major version when they do.

Also, you're correct that this currently doesn't reduce the number of dependencies today, but that may not be the case in the future. For instance, if `chrono` releases an `0.5` version, the feature flag will allow us to only optionally depend on `0.4` or `0.5`. Right now it doesn't change the dependencies because `0.4` is the latest version of `chrono`. The same goes for `uuid`.

tagging in @isabelatkinson for review

BTW, I filed RUST-841 for the `ObjectId::bytes` -> `const` change.

isabelatkinson (approved):

patrickfreed:
Thanks for your contribution @pymongo!

PR #261, opened 2021-06-03
RUST-806 Allow conversion of `Option`s to `Bson` types

NBSquare:
Implemented `From<Option<T>>` for Bson where `T: Into<Bson>`

Second commit is to suppress an issue on Evergreen's end.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -304,6 +304,18 @@ impl From<DbPointer> for Bson {
     }
 }
 
+impl<T> From<Option<T>> for Bson
+where
+    T: Into<Bson>,
+{
+    fn from(a: Option<T>) -> Bson {
+        match a {
+            None => Bson::Null,
+            Some(t) => t.into(),
+        }
+    }
+}
+
 /// This will create the [relaxed Extended JSON v2](https://docs.mongodb.com/manual/reference/mongodb-extended-json/) representation of the provided [`Bson`](../enum.Bson.html).
 impl From<Bson> for Value {
     fn from(bson: Bson) -> Self {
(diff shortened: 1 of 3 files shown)


src/tests/modules/bson.rs
```diff
@@ -143,6 +143,15 @@ fn from_impls() {
...
     );
 
+    // Optionals
+    assert_eq!(Bson::from(Some(4)), Bson::Int32(4));
+    assert_eq!(
+        Bson::from(Some(String::from("data"))),
+        Bson::String(String::from("data"))
+    );
+    assert_eq!(Bson::from(None::<i32>), Bson::Null);
+    assert_eq!(Bson::from(None::<String>), Bson::Null);
```
isabelatkinson, line 153: Can we also add a test here to ensure that this works properly in the `doc!` macro? Something along the lines of the code in the ticket description.

isabelatkinson (commented):
Looks great! Just one small test request.

src/tests/modules/bson.rs
NBSquare: I added tests for the `Some` and `None` cases. Let me know what you think.

isabelatkinson (approved):
LGTM! Tagging in Patrick for review.

src/bson.rs
```diff
@@ -304,6 +304,18 @@ impl From<DbPointer> for Bson {
...
 }
 
+impl<T> From<Option<T>> for Bson
+where
+    T: Into<Bson>,
+{
+    fn from(a: Option<T>) -> Bson {
+        match a {
+            None => Bson::Null,
+            Some(t) => T::into(t),
```
patrickfreed, line 314: It's a little more idiomatic to just invoke `into` as a method on `t` here: `t.into()`

patrickfreed (commented):
Just one little nitpick, otherwise LGTM!

patrickfreed (approved):
LGTM!

PR #262, opened 2021-06-04
RUST-282 Add pretty-printed `Debug` implementation to `Document`

NBSquare:
It looks like the only relevant struct which doesn't derive `Debug` is `Document`, so I added support for pretty-print there.

I also added a test to the end of [bson.rs](https://github.com/mongodb/bson-rust/compare/master...NBSquare:RUST-282?expand=1#diff-d3ab75438f3cd06c982a058c8ee3fb9cc79c21af4d8424f6d7c44b3ca5ab5272R312) test.

### DIFF
--- a/src/tests/modules/bson.rs
+++ b/src/tests/modules/bson.rs
@@ -316,7 +316,7 @@ fn debug_print() {
         b: String,
     }
 
-    let doc = doc! {"a": 1, "b": "data"};
+    let doc = doc! { "a": 1, "b": "data" };
     let data = indexmap! {"a" => Bson::from(1), "b" => Bson::from("data")};
 
     assert_eq!(format!("{:?}", doc), format!("Document({:?})", data));
--- a/src/tests/modules/bson.rs
+++ b/src/tests/modules/bson.rs
@@ -319,11 +319,13 @@ fn debug_print() {
         ]),
         "doc": doc! { "a": 1, "b": "data"},
     };
-    let normal_print = "Document({\"oid\": ObjectId(000000000000000000000000), \"arr\": \
+    let normal_print = "Document({\"oid\": ObjectId(\"000000000000000000000000\"), \"arr\": \
                         Array([Null, Timestamp { time: 1, increment: 1 }]), \"doc\": \
                         Document({\"a\": Int32(1), \"b\": String(\"data\")})})";
     let pretty_print = "Document({
-    \"oid\": ObjectId(000000000000000000000000),
+    \"oid\": ObjectId(
+        \"000000000000000000000000\",
+    ),
     \"arr\": Array([
         Null,
         Timestamp {
--- a/src/tests/modules/bson.rs
+++ b/src/tests/modules/bson.rs
@@ -16,6 +16,7 @@ use crate::{
     Regex,
     Timestamp,
 };
+use indexmap::indexmap;
 use serde_json::{json, Value};
 
 #[test]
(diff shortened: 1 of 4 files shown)


src/tests/modules/bson.rs
```diff
@@ -306,3 +307,18 @@ fn system_time() {
...
+
+#[test]
+fn debug_print() {
+    #[derive(Debug)]
+    struct Example {
+        a: i32,
+        b: String,
+    }
+
+    let doc = doc! {"a": 1, "b": "data"};
```
isabelatkinson, line 319: tiny nitpick: we usually add a space after the opening bracket and a space before the closing bracket when invoking this macro, so this should be `doc! { "a": 1, "b": "data" };`

isabelatkinson (commented):
this looks great, just one small formatting request

isabelatkinson (approved):
LGTM! Tagging in Patrick for review.

src/tests/modules/bson.rs
```diff
@@ -306,3 +307,18 @@ fn system_time() {
...
+#[test]
+fn debug_print() {
+    #[derive(Debug)]
+    struct Example {
+        a: i32,
+        b: String,
+    }
+
+    let doc = doc! { "a": 1, "b": "data" };
+    let data = indexmap! {"a" => Bson::from(1), "b" => Bson::from("data")};
```
patrickfreed, line 320: Rather than testing against the internal representation here, we should define a String for what we want this to look like. That way, if we ever change the internal representation of `Document`, we can maintain a consistent debug format, and we also won't have to update this test.

src/tests/modules/bson.rs
```diff
@@ -306,3 +307,18 @@ fn system_time() {
...
+
+#[test]
+fn debug_print() {
+    #[derive(Debug)]
+    struct Example {
+        a: i32,
+        b: String,
+    }
+
+    let doc = doc! { "a": 1, "b": "data" };
```
patrickfreed, line 319: Can we include some BSON-specific types in here to ensure they get pretty-printed as well? (e.g. an ObjectID and a subdocument).

Also, I noticed while trying this out that we have an existing issue where such types have their type printed twice in the debug mode, e.g.:
```
{ a: 1, b: "data", c: { bool: true, date: Date("2021-06-07 20:20:26.905 UTC") } }

// gets debug printed as
Document({
    "a": Int32(
        1,
    ),
    "b": String(
        "data",
    ),
    "c": Document(
        Document({
            "bool": Boolean(
                true,
            ),
            "date": DateTime(
                DateTime(
                    2021-06-07T20:20:26.905Z,
                ),
            ),
        }),
    ),
})
```
Could we fix that up as part of this work too? I think it will require a custom `Debug` implementation for `Bson`, since the derived one is what's duplicating the inner type name.

patrickfreed (commented):
Looks good! I just have a test suggestion and request for an existing but related issue to be fixed.

src/tests/modules/bson.rs
NBSquare: Makes sense. Is there a possibility of the external representation changing? Or is that just not very likely/not possible?

src/tests/modules/bson.rs
NBSquare: Ahh, interesting, thanks for the catch! I'll get to work on this when I'm done with my current ticket.

src/tests/modules/bson.rs
patrickfreed: The likelihood is low, though we did do it before (we used to use `linked-hash-map` instead of `indexmap`). Another benefit of defining the expected result absolutely (i.e. as a string) rather than relative to `indexmap` is that it also protects us from the situation where `indexmap` changes its own `Debug` implementation to something that differs from the one we want / currently use. More abstractly put, by comparing to `indexmap`'s output, the test verifies only the implementation, but by comparing to a pre-defined string, it also tests the stability of the public interface.

NBSquare:
I modified the `Debug` implementation for `ObjectId` to reflect the fact that `ObjectId` is basically a tuple wrapping the ID slice, which meant modified a unit test or two. [Check that out here.](https://github.com/mongodb/bson-rust/pull/262/commits/faadd50ea0065bf1f3c78264d4d6bff603c4946a)

This means that instead of printing out `ObjectId`s as
```
ObjectId(000000000000000000000000)
```
it will print them out as
```
ObjectId("000000000000000000000000")
```
or
```
ObjectId(
    "000000000000000000000000",
)
```
depending on the format specifier. This is an easy fix if we want to follow the pattern `serde_json` uses for `Object`s and `Array`s.

patrickfreed (approved):
LGTM! Great job with this, especially with handling all the more ambiguous cases.

isabelatkinson (approved):
nice job!

PR #273, opened 2021-06-24
RUST-868 Add `serialize_object_id_as_hex_string` helper

isabelatkinson:
Adds a `serialize_object_id_as_hex_string` serde helper. Note that a deserialize helper is not needed as the implementation of `Deserialize` for `ObjectId` already does this.

### DIFF
--- a/src/serde_helpers.rs
+++ b/src/serde_helpers.rs
@@ -2,7 +2,9 @@
 
 use std::{convert::TryFrom, result::Result};
 
-use serde::{ser, Serializer};
+use serde::{ser, Serialize, Serializer};
+
+use crate::oid::ObjectId;
 
 pub use bson_datetime_as_rfc3339_string::{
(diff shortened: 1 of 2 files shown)


abr-egn (approved):

src/serde_helpers.rs
```diff
@@ -82,6 +84,14 @@ pub fn serialize_u64_as_i64<S: Serializer>(val: &u64, serializer: S) -> Result<S
     }
 }
 
+/// Serializes an [`ObjectId`] as a hex string.
+pub fn serialize_object_id_as_hex_string<S: Serializer>(
+    val: &ObjectId,
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    val.to_string().serialize(serializer)
```
patrickfreed, line 92: Can we use `to_hex` here instead? just in case we ever change the `Display` implementation.

patrickfreed (approved):

PR #275, opened 2021-06-30
RUST-665 Sync spec tests for dots and dollars in keys

abr-egn:


isabelatkinson (approved):

NBSquare:
Drive-by review 😮 
Do we care about syncing the DBRef spec test since we're not actually running it? I know there are a couple dollar-key-related tests that we haven't added.

abr-egn:
> Drive-by review 😮
> Do we care about syncing the DBRef spec test since we're not actually running it? I know there are a couple dollar-key-related tests that we haven't added.

I don't think a sync is needed here, since (as you say) the dbref test isn't relevant to the rust driver.  Whatever's needed for that will get resolved by RUST-762.

patrickfreed (approved):
As a general policy, we sync all the tests relevant to the DRIVERS ticket that the RUST ticket is descended from, and in the code of the test runner skip non-applicable tests. That way, in the future we can continue to simply sync the tests without having to manually verify whether a test should be included or not.

PR #277, opened 2021-07-07
RUST-891 draft: make chrono as a optional dependency

pymongo:
bson's datetime is a i64 timestamp in millisecond like javascript, it is quite different to chrono's Datetime

since bson-rust has chrono-0_4 feature to turn on/off chrono

but If bson-rust disable chrono-0_4 feature, still a amout of functions require chrono crate

---

This PR is my attempt to impl some functions without chrono, I think chrono-0_4 need more discussions.

Please don't merge it, it just a drat and still working in progress.

some test cases didn't pass!

```
Diff < left / right > :
 Object({
     "a": Object({
         "$date": String(
<            "1970-01-01T00:00:00+0000",
>            "1970-01-01T00:00:00Z",
         ),
     }),
 })

failures:
    tests::modules::macros::standard_format
    tests::serde::test_datetime_helpers
    tests::serde::test_ser_datetime
    tests::spec::corpus::run

test result: FAILED. 97 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ exclude = [
 # no features by default
 default = []
 # if enabled, include API for interfacing with chrono 0.4
-chrono-0_4 = []
+chrono-0_4 = ["chrono"]
 # if enabled, include API for interfacing with uuid 0.8
 uuid-0_8 = []
 # attempt to encode unsigned types in signed types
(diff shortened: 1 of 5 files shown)


patrickfreed:
Hi @pymongo, thanks for the PR! Unfortunately, implementing datetime parsing logic isn't something we think belongs in `bson` itself and would prefer to delegate that work to `chrono`. It's very tricky logic to get right and may involve a lot of low level / `unsafe` code which we'd rather not have to maintain when a suitable third-party crate already exists.

Note that the impact of having `chrono` as a dependency is minimal, as it is a relatively small crate with few required dependencies. Also, its very likely that some other crate in a user's application is already pulling in the same version of `chrono` as `bson`, meaning having it as a dependency is essentially cost free.

Thanks again for your contribution, but unfortunately for those reasons I'm going to have to close this PR. Let us know if you have any further questions!

PR #279, opened 2021-07-22
RUST-871 Support direct serialization to BSON bytes

patrickfreed:
RUST-871

This PR introduces the `to_writer` and `to_vec` methods, which can serialize a generic `T` directly to BSON bytes without going through `Document`. This should enable substantial performance improvements in the driver, as well as be generally useful in the BSON library.

As part of testing this feature, I enabled `decimal128` serde testing on evergreen (RUST-446). decimal128 tests were originally added to evergreen in #279.


src/bson.rs
```diff
@@ -698,6 +698,20 @@ impl Bson {
                 }
             }
 
+            ["$numberDecimalBytes"] => {
```
patrickfreed, line 701: Unintuitively, this function is used in the existing serializer to convert between the intermediate serialized form (i.e. and "extended document") and the final serialized form. Now that decimal128 can serialize this way, this needed to be updated to interpret it.

src/ser/raw/document_serializer.rs
```diff
@@ -0,0 +1,323 @@
...
+        value.serialize(&mut *self.root_serializer)
+    }
+
+    fn end(self) -> Result<Self::Ok> {
+        self.end_doc().map(|_| ())
+    }
+}
+
+/// Serializer used specifically for serializing document keys.
+struct KeySerializer<'a> {
```
patrickfreed, line 163: There's a whole lot of boilerplate involved in implementing this type, since unfortunately there isn't a serialization equivalent to the `forward_to_deserialize_any!` macro in deserialization.

All but a few of these are basically unimplemented, the few being `serialize_str` and `serialize_some`.

src/ser/raw/value_serializer.rs
```diff
@@ -0,0 +1,542 @@
...
+    oid::ObjectId,
+    ser::{write_binary, write_cstring, write_i32, write_i64, write_string, Error, Result},
+    spec::{BinarySubtype, ElementType},
+};
+
+use super::{document_serializer::DocumentSerializer, Serializer};
+
+/// A serializer used specifically for serializing the serde-data-model form of a BSON type (e.g.
+/// `Binary`) to raw bytes.
+pub(crate) struct ValueSerializer<'a> {
```
patrickfreed, line 20: This type is used to interpret any extended JSON formatted types in serde's data model and serialize them to actual BSON. It keeps track of where it's at via a state machine.

src/ser/raw/value_serializer.rs
```diff
@@ -0,0 +1,542 @@
...
+    ) -> Result<Self::SerializeStructVariant> {
+        Err(self.invalid_step("struct_variant"))
+    }
+}
+
+impl<'a, 'b> SerializeStruct for &'b mut ValueSerializer<'a> {
+    type Ok = ();
+    type Error = Error;
+
+    fn serialize_field<T: ?Sized>(&mut self, key: &'static str, value: &T) -> Result<()>
```
patrickfreed, line 407: this method handles all the state transitions.

src/ser/serde.rs
```diff
@@ -53,22 +56,54 @@ impl Serialize for Bson {
     where
         S: ser::Serializer,
     {
-        match *self {
```
patrickfreed, line 56: this was updated to just call through to the underlying value's serialize method. Previously, it was the other way around. This lets us use named structs in the data model instead of nameless maps.

src/ser/serde.rs
```diff
@@ -551,8 +607,12 @@ impl Serialize for Decimal128 {
     where
         S: ser::Serializer,
     {
-        let value = Bson::Decimal128(self.clone());
-        value.serialize(serializer)
+        let mut state = serializer.serialize_struct("$numberDecimal", 1)?;
```
patrickfreed, line 610: When the `decimal128` flag was enabled, this used to produce actual extended JSON format (i.e. `{ "$numberDecimal": <decimal as a string> }`), but per the bug we discussed in triage, the current implementation's `to_string` is lossy. For that reason, it seems safer to produce this unique format that does round trip properly until we have a complete decimal128 implementation.

For users relying on the output of this, their code will break, but that's probably better than them silently losing data.
Also note that we classify the `decimal128` feature flag as "experimental" and subject to breaking changes, so this is within our contract. Code that omits the flag will not break.

src/tests/spec/corpus.rs
```diff
@@ -61,63 +61,95 @@ fn run_test(test: TestFile) {
 
         let canonical_bson = hex::decode(&valid.canonical_bson).expect(&description);
 
-        let bson_to_native_cb =
```
patrickfreed, line 64: I went ahead and updated the naming scheme of the different things here. Let me know if this makes sense. I know the old one got the point of being unreadable to me at least, so hopefully this is better.

src/ser/mod.rs
```diff
@@ -204,3 +216,27 @@ where
         }),
     }
 }
+
+/// Serialize the given `T` as BSON bytes into the provided writer.
+#[inline]
+pub fn to_writer<T, W>(value: &T, mut writer: W) -> Result<()>
```
abr-egn, line 222: This is equivalent to `writer.write_all(to_vec(value)?.as_slice())`, yes?  If so, I'm not sure it needs to be in the public API.

abr-egn (approved):

isabelatkinson (approved):

PR #281, opened 2021-07-22
RUST-882 Fix lossy `From<u32>` Bson impl, prevent overflow in `From<u64>` (1.2.x)

patrickfreed:
RUST-882

This is the `1.2.x` backport fix of RUST-882.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -22,6 +22,7 @@
 //! BSON definition
 
 use std::{
+    convert::{TryFrom, TryInto},
     fmt::{self, Debug, Display},
     ops::{Deref, DerefMut},
 };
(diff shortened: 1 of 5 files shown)


src/bson.rs
```diff
@@ -266,13 +263,21 @@ impl From<i64> for Bson {
...
 }
 
 impl From<u64> for Bson {
+    /// This conversion is lossy if the provided `u64` is greater than `i64::MAX`, which it will be converted to.
+    /// Using this `From` implementation is highly discouraged and it will be removed in the next major version.
+    ///
+    /// Note: due to https://github.com/rust-lang/rust/issues/39935 we cannot deprecate this implementation.
     fn from(a: u64) -> Bson {
-        Bson::Int64(a as i64)
+        Bson::Int64(a.try_into().unwrap_or(i64::MAX))
```
patrickfreed, line 280: for lack of a better option, I tried to make this "less bad", but short of panicking I'm not sure we have any other choice. I'm hesitant to panic given that we can't even deprecate this impl though.

abr-egn: Agreed that this is the less-bad option.

abr-egn (approved):

isabelatkinson (approved):

PR #285, opened 2021-08-02
RUST-880 Fix crash when deserializing/serializing `Document` that contains decimal128

patrickfreed:
RUST-880

This PR fixes a crash that occurs when deserializing from or serializing to a `Document` that contains a decimal128 when the `decimal128` feature flag is not enabled.

This PR also fixes the clippy failures that started since the latest Rust release.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -122,7 +122,7 @@ impl Display for Bson {
             }) => write!(fmt, "/{}/{}", pattern, options),
             Bson::JavaScriptCode(ref code)
             | Bson::JavaScriptCodeWithScope(JavaScriptCodeWithScope { ref code, .. }) => {
-                fmt.write_str(&code)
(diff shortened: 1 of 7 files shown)


src/bson.rs
```diff
@@ -574,6 +574,15 @@ impl Bson {
                     "$numberDecimal": (v.to_string())
                 }
             }
+            #[cfg(not(feature = "decimal128"))]
+            Bson::Decimal128(ref v) => {
+                doc! {
+                    "$numberDecimalBytes": Bson::Binary(Binary {
```
patrickfreed, line 580: This format matches what we went for on master.

abr-egn (approved):

isabelatkinson (approved):

NBSquare (approved):

Issue #282, opened 2021-07-29
Decimal128 feature not storing the right value to MongoDB


ale-rinaldi:
Hello,

I'm having a fairly serious issue in using the decimal128 feature, with bson 2.0.0-beta.2.

I created a really simple, compilable project that describes the issue: https://github.com/ale-rinaldi/rust-decimal-test

The important part is here:
```rust
/* uses */

#[derive(Serialize)]
struct TestData {
    pub(crate) my_date: bson::DateTime,
    pub(crate) my_decimal: bson::Decimal128,
    pub(crate) my_string: String,
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    /* Code for connection */

    let data = TestData {
        my_date: bson::DateTime::from_millis(1278720000000),
        my_decimal: bson::Decimal128::from_u32(42),
        my_string: "foobar".to_string()
    };
    let doc = bson::to_document(&data).unwrap();
    collection.insert_one(doc, None).await.unwrap();
    Ok(())
}
```

So I'm creating a simple test struct with a Date, a Decimal128 and a String, and saving it to Mongo (I'm on MongoDB 4.4.7 but I tried this with various versions, even back to 4.1.13, and the issue is the same).
What happens here is that, if I look at the created document using MongoDB Compass, the date and the string fields have the value that I except (`2010-07-10T00:00:00.000+00:00` and `foobar`), while the decimal has no value 42 but `6.6E-1819`.

The strange thing is that, if I read the document back from Rust, the value of the returned Decimal128 is now correct (42).

Am I totally missing the point here, or is it an actual problem with the `bson` serializer?

Thanks

Bedotech:
The same issue there is in the master branch.

patrickfreed:
So the `decimal128` feature flag enables an "experimental" and incomplete decimal128 implementation that we inherited from when this library was maintained by the community. Unfortunately it currently has little testing and is not recommended for use in production.

Looking into your issue in particular, I've noticed that the `decimal` crate, which we use for our `Decimal128` type when the `decimal128` feature flag is enabled, implements a different coefficient encoding than what is used by MongoDB and its other drivers. This means that the Rust driver / BSON library will be able to correctly deserialize `Decimal128` values that _it_ inserted into the database, but other tools and the database itself will read garbage data (as you have observed). Likewise, the Rust driver will not be able to properly deserialize decimal128 values that _other_ drivers and tools have inserted.

[...]

PR #288, opened 2021-08-04
RUST-960 Deprecate `decimal128` feature flag (1.2.x)

patrickfreed:
RUST-960

This PR deprecates the `decimal128` feature flag due to the reasons highlighted in https://github.com/mongodb/bson-rust/issues/282.

semver requires that deprecations happen in minor rather than patch versions, but `decimal128` being "unstable and experimental" means it isn't really bound by semver, so I _think_ we're okay to do this in a patch release. Thoughts?

### DIFF
--- a/src/decimal128.rs
+++ b/src/decimal128.rs
@@ -10,8 +10,12 @@ use std::fmt;
 /// deserialized from existing documents that contain BSON decimal128s.
 ///
 /// Experimental functionality can be enabled through the usage of the `"decimal128"`
-/// feature flag. Note that the API and behavior of such functionality are unstable and
-/// subject to change.
+/// feature flag. The flag is not recommended for use however, as it causes `Decimal128` values to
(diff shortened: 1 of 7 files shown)


src/decimal128.rs
```diff
@@ -23,6 +27,8 @@ pub struct Decimal128 {
 }
 
 #[cfg(feature = "decimal128")]
+#[deprecated = "The feature-gated decimal128 implementation serializes to BSON incorrectly and \
+                should not be used. It will be removed completely in 2.0.0"]
 impl Decimal128 {
```
patrickfreed, line 32: To only deprecate the feature-flag and not the `Decimal128` struct itself, I only added the attribute to this impl block.

isabelatkinson (approved):
releasing in a patch version SGTM

abr-egn (approved):
Patch release sounds fine to me as well.

NBSquare (approved):
Patch version seems fine.

PR #302, opened 2021-09-09
RUST-908 Validate that C strings do not contain null bytes

isabelatkinson:
Validates that any cstrings in BSON (as defined by the [BSON spec](https://bsonspec.org/spec.html)) do not contain null bytes. Rather than enforce this for all of our Rust BSON types (e.g. `Document`, `Regex`) which would be difficult/impossible in some cases, I opted to enforce it when the BSON is encoded into bytes. The entry points for this in our API are `Document::to_reader` and `to_vec`.

### DIFF
--- a/src/ser/error.rs
+++ b/src/ser/error.rs
@@ -41,7 +41,7 @@ impl fmt::Display for Error {
             Error::Io(ref inner) => inner.fmt(fmt),
             Error::InvalidDocumentKey(ref key) => write!(fmt, "Invalid map key type: {}", key),
             Error::InvalidCString(ref string) => {
-                write!(fmt, "cstrings cannot contain null bytes: {}", string)
+                write!(fmt, "cstrings cannot contain null bytes: {:?}", string)
             }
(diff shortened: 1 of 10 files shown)


src/ser/error.rs
```diff
@@ -37,6 +40,9 @@ impl fmt::Display for Error {
         match *self {
             Error::Io(ref inner) => inner.fmt(fmt),
             Error::InvalidDocumentKey(ref key) => write!(fmt, "Invalid map key type: {}", key),
+            Error::InvalidCString(ref string) => {
+                write!(fmt, "cstrings cannot contain null bytes: {}", string)
```
abr-egn, line 44: Minor: might want to use `{:?}` instead to make it easier to see the escaped contents of the string.

abr-egn (approved):
LGTM!

patrickfreed (approved):

src/ser/error.rs
isabelatkinson: good idea, updated

PR #305, opened 2021-09-13
RUST-787 impl Display for various Bson types

awitten1:
Implements display for Bson types.  Wraps Bson enum display.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -984,6 +984,13 @@ pub struct Timestamp {
     pub increment: u32,
 }
 
+impl Display for Timestamp {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let x: Bson = self.into();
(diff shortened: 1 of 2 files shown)


abr-egn (approved):

src/bson.rs
```diff
@@ -1035,6 +1056,13 @@ pub struct Binary {
     pub bytes: Vec<u8>,
 }
 
+impl Display for Binary {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        let x: Bson = self.into();
```
patrickfreed, line 1061: These `.into()` calls will make copies, which can potentially be costly. Instead, can we move the meat of the `Display` implementation here and call into it from `Bson`? This will let us avoid a copy. Ditto for the other types.

patrickfreed (approved):
LGTM!

PR #306, opened 2021-09-13
RUST-908 Validate that C strings do not contain null bytes (1.2.x backport)

isabelatkinson:


src/tests/spec/corpus.rs
```diff
@@ -283,7 +288,11 @@ fn run_test(test: TestFile) {
         let json: serde_json::Value =
             serde_json::from_str(parse_error.string.as_str()).expect(&parse_error.description);
 
-        Bson::try_from(json).expect_err(&parse_error.description);
+        if let Ok(bson) = Bson::try_from(json.clone()) {
+            // if converting to bson succeeds, assert that translating that bson to bytes fails
+            let mut vec = Vec::new();
+            assert!(bson.as_document().unwrap().to_writer(&mut vec).is_err());
```
isabelatkinson, line 294: `to_vec` was introduced in 2.0 so we need to use `Document::to_writer` here. All of the tests that make it to this logic use documents so the unwrap is safe.

abr-egn (approved):

src/ser/error.rs
```diff
@@ -17,8 +17,12 @@ pub enum Error {
...
     },
 
-    /// A general error that ocurred during serialization.
-    /// See: https://docs.rs/serde/1.0.110/serde/ser/trait.Error.html#tymethod.custom
+    /// An invalid string was specified.
+    InvalidCString(String),
+
+    /// A general error that occurred during serialization.
+    /// See: <https://docs.rs/serde/1.0.110/serde/ser/trait.Error.html#tymethod.custom>
+    #[non_exhaustive]
```
patrickfreed, line 25: Adding `non_exhaustive` here would actually be a breaking change, so we'll have to omit it.

src/macros.rs
```diff
@@ -175,7 +175,7 @@ macro_rules! bson {
     };
 
     ({$($tt:tt)+}) => {
-        $crate::Bson::Document($crate::doc!{$($tt)+});
+        $crate::Bson::Document($crate::doc!{$($tt)+})
```
patrickfreed, line 178: What does this change do / why was it necessary?

patrickfreed (approved):

PR #308, opened 2021-09-15
RUST-819 Add methods for converting between Binary and Uuid

awitten1:
Refactor Binary serde implementation.  Add Binary constructors consistent with [specs](https://github.com/mongodb/specifications/blob/master/source/uuid.rst).

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -1056,6 +1056,110 @@ impl Binary {
 
 #[cfg(feature = "uuid-0_8")]
 #[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+pub enum UuidRepresentation {
+    Standard,
+    CSharpLegacy,
+    JavaLegacy,
+    PythonLegacy,
+}
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+impl Binary {
+    pub fn from_uuid(uuid: &uuid::Uuid) -> Self {
+        Binary::from(*uuid)
+    }
+
+    pub fn from_uuid_with_representation(uuid: &uuid::Uuid, rep: UuidRepresentation) -> Self {
+        match rep {
+            UuidRepresentation::Standard => Binary::from_uuid(uuid),
+            UuidRepresentation::CSharpLegacy => {
+                let mut bytes = uuid.as_bytes().to_vec();
+                bytes[0..4].reverse();
+                bytes[4..6].reverse();
+                bytes[6..8].reverse();
+                Binary {
+                    subtype: BinarySubtype::UuidOld,
+                    bytes,
+                }
+            }
+            UuidRepresentation::PythonLegacy => Binary {
+                subtype: BinarySubtype::UuidOld,
+                bytes: uuid.as_bytes().to_vec(),
+            },
+            UuidRepresentation::JavaLegacy => {
+                let mut bytes = uuid.as_bytes().to_vec();
+                bytes[0..8].reverse();
+                bytes[8..16].reverse();
+                Binary {
+                    subtype: BinarySubtype::UuidOld,
+                    bytes,
+                }
+            }
+        }
+    }
+
+    pub fn to_uuid_with_representation(&self, rep: UuidRepresentation) -> Result<uuid::Uuid, &str> {
+        match rep {
+            UuidRepresentation::Standard => self.to_uuid(),
+            UuidRepresentation::CSharpLegacy => {
+                if self.subtype != BinarySubtype::UuidOld {
+                    Err("expecting BinarySubtype::UuidOld")
+                } else if self.bytes.len() != 16 {
+                    Err("expecting 16 bytes")
+                } else {
+                    let mut buf = [0u8; 16];
+                    buf.copy_from_slice(&self.bytes);
+                    buf[0..4].reverse();
+                    buf[4..6].reverse();
+                    buf[6..8].reverse();
+                    Ok(uuid::Uuid::from_bytes(buf))
+                }
+            }
+            UuidRepresentation::PythonLegacy => {
+                if self.subtype != BinarySubtype::UuidOld {
+                    Err("expecting BinarySubtype::UuidOld")
+                } else if self.bytes.len() != 16 {
+                    Err("expecting 16 bytes")
+                } else {
+                    let mut buf = [0u8; 16];
+                    buf.copy_from_slice(&self.bytes);
+                    Ok(uuid::Uuid::from_bytes(buf))
+                }
+            }
+            UuidRepresentation::JavaLegacy => {
+                if self.subtype != BinarySubtype::UuidOld {
+                    Err("expecting BinarySubtype::UuidOld")
+                } else if self.bytes.len() != 16 {
+                    Err("expecting 16 bytes")
+                } else {
+                    let mut buf = [0u8; 16];
+                    buf.copy_from_slice(&self.bytes);
+                    buf[0..8].reverse();
+                    buf[8..16].reverse();
+                    Ok(uuid::Uuid::from_bytes(buf))
+                }
+            }
+        }
+    }
+
+    pub fn to_uuid(&self) -> Result<uuid::Uuid, &str> {
+        if self.subtype == BinarySubtype::Uuid {
+            if self.bytes.len() == 16 {
+                let mut bytes = [0u8; 16];
+                bytes.copy_from_slice(&self.bytes);
+                Ok(uuid::Uuid::from_bytes(bytes))
+            } else {
+                Err("cannot convert Binary to Uuid: incorrect bytes length")
+            }
+        } else {
+            Err("cannot convert Binary to Uuid: incorrect binary subtype")
+        }
+    }
+}
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
 impl From<uuid::Uuid> for Binary {
     fn from(uuid: uuid::Uuid) -> Self {
         Binary {
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -1084,7 +1084,8 @@ pub enum UuidRepresentation {
     CSharpLegacy,
     /// The legacy representation of UUIDs in BSON used by the Java driver (binary subtype 0x03)
     JavaLegacy,
-    /// The legacy representation of UUIDs in BSON used by the Python driver, which is the same format as STANDARD, but has binary subtype 0x03
+    /// The legacy representation of UUIDs in BSON used by the Python driver, which is the same
+    /// format as STANDARD, but has binary subtype 0x03
     PythonLegacy,
 }
 
@@ -1126,26 +1127,34 @@ impl Binary {
             }
         }
     }
-    /// Deserializes a BSON binary type into a UUID, takes the representation with which the Binary was serialized.
+    /// Deserializes a BSON binary type into a UUID, takes the representation with which the Binary
+    /// was serialized.
     pub fn to_uuid_with_representation(
         &self,
         rep: UuidRepresentation,
     ) -> Result<uuid::Uuid, crate::de::Error> {
         // If representation is non-standard, then its subtype must be UuidOld
         if rep != UuidRepresentation::Standard && self.subtype != BinarySubtype::UuidOld {
-            return Err(Error::custom(
-                format!("expected binary subtype 3 when converting to UUID with a non-standard representation, instead got {:#04x}", u8::from(self.subtype)),
-            ));
+            return Err(Error::custom(format!(
+                "expected binary subtype 3 when converting to UUID with a non-standard \
+                 representation, instead got {:#04x}",
+                u8::from(self.subtype)
+            )));
         }
         // If representation is standard, then its subtype must be Uuid
         if rep == UuidRepresentation::Standard && self.subtype != BinarySubtype::Uuid {
-            return Err(Error::custom(
-                format!("expected binary subtype 4 when converting to UUID with the standard representation, instead got {:#04x}", u8::from(self.subtype))
-            ));
+            return Err(Error::custom(format!(
+                "expected binary subtype 4 when converting to UUID with the standard \
+                 representation, instead got {:#04x}",
+                u8::from(self.subtype)
+            )));
         }
         // Must be 16 bytes long
         if self.bytes.len() != 16 {
-            return Err(Error::custom(format!("expected UUID to contain 16 bytes, instead got {}", self.bytes.len())));
+            return Err(Error::custom(format!(
+                "expected UUID to contain 16 bytes, instead got {}",
+                self.bytes.len()
+            )));
         }
         let mut buf = [0u8; 16];
         buf.copy_from_slice(&self.bytes);
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -36,6 +36,9 @@ use crate::{
     Decimal128,
 };
 
+#[cfg(feature = "uuid-0_8")]
+use serde::de::Error;
+
 /// Possible BSON value types.
 #[derive(Clone, PartialEq)]
 pub enum Bson {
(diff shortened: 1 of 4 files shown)


src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
...
+                let mut bytes = [0u8; 16];
+                bytes.copy_from_slice(&self.bytes);
+                Ok(uuid::Uuid::from_bytes(bytes))
+            } else {
+                Err("cannot convert Binary to Uuid: incorrect bytes length")
+            }
+        } else {
+            Err("cannot convert Binary to Uuid: incorrect binary subtype")
+        }
+    }
```
abr-egn, line 1147: I try to make the happy path more visible (and reduce rightward drift) by making the error cases early returns, i.e.

```suggestion
        if self.subtype != BinarySubtype::Uuid {
            return Err("cannot convert Binary to Uuid: incorrect binary subtype");
        }
        if self.bytes.len() != 16 {
            return Err("cannot convert Binary to Uuid: incorrect bytes length");
        }
        let mut bytes = [0u8; 16];
        bytes.copy_from_slice(&self.bytes);
        Ok(uuid::Uuid::from_bytes(bytes))
    }
```

src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
...
+                bytes[8..16].reverse();
+                Binary {
+                    subtype: BinarySubtype::UuidOld,
+                    bytes,
+                }
+            }
+        }
+    }
+
+    pub fn to_uuid_with_representation(&self, rep: UuidRepresentation) -> Result<uuid::Uuid, &str> {
```
abr-egn, line 1103: This (and others) should return an error enum rather than a &str.

patrickfreed, line 1103: We typically define error types rather than using strings for errors; an example can see an example [here](https://github.com/mongodb/bson-rust/blob/master/src/de/error.rs#L10). Though for this, we could just add a case to `de::Error` and use that here. In the interest of trying to keep the error API simple and reduce the number of disparate types, I think I lean towards just reusing `de::Error` here, but I'm not 100% certain. Any opinions @abr-egn ?

src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
...
+                buf.copy_from_slice(&self.bytes);
+                buf[0..8].reverse();
+                buf[8..16].reverse();
+                Ok(uuid::Uuid::from_bytes(buf))
+            }
+        }
+    }
+
+    pub fn to_uuid(&self) -> Result<uuid::Uuid, &str> {
+        if self.subtype == BinarySubtype::Uuid {
```
patrickfreed, line 1136: to consolidate this logic / error checking to  a single method, I think it would be slightly better to move this code to `to_uuid_with_representation` and then just call through to that method here.

src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
     }
 }
 
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+#[derive(PartialEq)]
+pub enum UuidRepresentation {
```
patrickfreed, line 1060: we generally strive to add docstrings for all public types, so that includes this enum itself, each variant, and all its methods. You can use the spec and documentation from other drivers (e.g. [Java](https://mongodb.github.io/mongo-java-driver/4.3/apidocs/bson/org/bson/UuidRepresentation.html)) as inspiration for what the docstrings might contain.

patrickfreed, line 1060: just to be safe, I think we'll need to tag this with `#[non_exhaustive]`. That way it won't be considered a breaking change to add a new case in the future.

src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
     }
 }
 
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+#[derive(PartialEq)]
```
patrickfreed, line 1059: can we derive `Debug`, `Copy,` and `Clone` on this type?

patrickfreed (commented):
Great job on this so far, it looks to be off to a good start!

While this PR does involve refactoring the UUID serde interop, the main user-facing change is the introduction of new `Uuid` conversions to the `Binary` type. In light of that, could we update the title of the PR to highlight the new conversions instead? That way in the release notes / commit log it'll be a bit more clear what the functional change was.

src/bson.rs
```diff
@@ -1054,6 +1054,99 @@ impl Binary {
...
+                Binary {
+                    subtype: BinarySubtype::UuidOld,
+                    bytes,
+                }
+            }
+        }
+    }
+
+    pub fn to_uuid_with_representation(&self, rep: UuidRepresentation) -> Result<uuid::Uuid, &str> {
+        if self.subtype != BinarySubtype::UuidOld && UuidRepresentation::Standard != rep {
```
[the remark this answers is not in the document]
awitten1: This check is saying that if the representation is non-standard, then the subtype should be type 3 (`BinarySubtype::UuidOld`)..  So it should permit Standard representation with subtype 4 (`BinarySubtype::Uuid`).  Should `BinarySubtype::UuidOld` be accepted with the Standard representation?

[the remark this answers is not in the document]
awitten1: Yup, I was already doing that.  Definitely more clear that way.

src/bson.rs
abr-egn: Reusing `de::Error` sounds good :)

src/serde_helpers.rs
```diff
@@ -437,16 +423,12 @@ pub mod uuid_as_java_legacy_binary {
         D: Deserializer<'de>,
     {
         let binary = Binary::deserialize(deserializer)?;
-        if binary.subtype != BinarySubtype::UuidOld {
-            Err(de::Error::custom("expecting BinarySubtype::UuidOld"))
-        } else if binary.bytes.len() != 16 {
-            Err(de::Error::custom("expecting 16 bytes"))
+        let res = binary.to_uuid_with_representation(UuidRepresentation::JavaLegacy);
```
awitten1, line 426: I'm not thrilled with this unpacking and then repacking of the result of `to_uuid` but returning the result directly doesn't type check.  I'm not sure if it's possible to constrain `D::Error` to be of type `crate::de::Error` but it doesn't compile when I directly attempt the constraint.

src/serde_helpers.rs
```diff
@@ -374,20 +374,13 @@ pub mod uuid_as_binary {
...
-            if binary.bytes.len() == 16 {
-                let mut bytes = [0u8; 16];
-                bytes.copy_from_slice(&binary.bytes);
-                Ok(Uuid::from_bytes(bytes))
-            } else {
-                Err(de::Error::custom(
-                    "cannot convert Binary to Uuid: incorrect bytes length",
-                ))
-            }
+        let res = binary.to_uuid();
```
awitten1, line 377: I'm not thrilled with this unpacking and then repacking of the result of to_uuid but returning the result directly doesn't type check. I'm not sure if it's possible to constrain `D::Error` to be of type `crate::de::Error` but it doesn't compile when I directly attempt the constraint.

src/bson.rs
```diff
@@ -1054,6 +1057,118 @@ impl Binary {
...
+    }
+
+    pub fn to_uuid_with_representation(
+        &self,
+        rep: UuidRepresentation,
+    ) -> Result<uuid::Uuid, crate::de::Error> {
+        // If representation is non-standard, then its subtype must be UuidOld
+        if rep != UuidRepresentation::Standard && self.subtype != BinarySubtype::UuidOld {
+            return Err(Error::custom(
+                "expecting binary subtype 3 for non-standard representations",
```
patrickfreed, line 1135: can we add some more context to this error? It'll make it easier for users to debug.

e.g.
```rust
format!("expected binary subtype 3 when converting to UUID, instead got {:#04x}", u8::from(self.subtype))
```

src/bson.rs
```diff
@@ -1054,6 +1057,118 @@ impl Binary {
...
+            ));
+        }
+        // If representation is standard, then its subtype must be Uuid
+        if rep == UuidRepresentation::Standard && self.subtype != BinarySubtype::Uuid {
+            return Err(Error::custom(
+                "expecting binary subtype 4 for standard representation",
```
patrickfreed, line 1141: ditto add subtype information here

src/bson.rs
```diff
@@ -1054,6 +1057,118 @@ impl Binary {
...
+            ));
+        }
+        // Must be 16 bytes long
+        if self.bytes.len() != 16 {
+            return Err(Error::custom("expecting 16 bytes"));
```
patrickfreed, line 1146: here too we could add a little context:
```rust
format!("expected UUID to contain 16 bytes, instead got {}", self.bytes.len())
```

src/serde_helpers.rs
patrickfreed: All `Error` types implement `Display` (see trait requirement [here](https://doc.rust-lang.org/std/error/trait.Error.html)), so we can avoid destructuring and just do the following:
```rust
binary.to_uuid().map_err(de::Error::custom)?
```
And this works because [`serde::de::Error::custom`](https://docs.serde.rs/serde/de/trait.Error.html#tymethod.custom) accepts any `T: Display`. This will probably clone the error, but the error path isn't the "hot path" so it's not really a concern.

patrickfreed (commented):
Sorry to nitpick the title of the PR again, but we want to make sure that it conveys the change on its own without necessarily having to refer to another resource (e.g. the specifications), and we also want to make sure it's clear that this change has to do with UUID support. With that in mind, could we update it to something like "Add methods for converting between `Binary` and `Uuid`"

src/bson.rs
```diff
@@ -1054,6 +1057,131 @@ impl Binary {
     }
 }
 
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
```
patrickfreed, line 1061: typically we group all the attributes together, so this would be better below with the `enum` declaration. Also note that the doctest won't get run unless these are met, even if they come after it.

src/bson.rs
```diff
@@ -1054,6 +1057,131 @@ impl Binary {
...
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+impl Binary {
+    /// Serializes a UUID into BSON binary type
+    pub fn from_uuid(uuid: uuid::Uuid) -> Self {
+        Binary::from(uuid)
+    }
+    /// Serializes a UUID into BSON binary type and takes the desired representation as a parameter.
+    /// Binary::from_uuid_with_representation(uuid, UuidRepresentation::Standard) is equivalent
```
patrickfreed, line 1102: These docstrings are markdown, so you can enclose the code portions in \`\` for easier reading. Also, in rust docs if you also enclose them in square brackets `[]`, it'll automatically link to the thing. e.g. [\`Binary::from_uuid\`].

src/bson.rs
```diff
@@ -1054,6 +1057,131 @@ impl Binary {
...
+                bytes[8..16].reverse();
+                Binary {
+                    subtype: BinarySubtype::UuidOld,
+                    bytes,
+                }
+            }
+        }
+    }
+    /// Deserializes a BSON binary type into a UUID, takes the representation with which the Binary
+    /// was serialized.
```
patrickfreed, line 1133:
```suggestion
    /// was serialized.
    ///
    /// See the documentation for [`UuidRepresentation`] for more information on the possible representations.
```

patrickfreed: Can we add this line to `from_uuid_with_representation` too?

src/tests/modules/bson.rs
```diff
@@ -345,3 +345,177 @@ fn debug_print() {
...
+#[cfg(feature = "uuid-0_8")]
+#[test]
+fn test_binary_constructors() {
+    use crate::bson::UuidRepresentation;
+    use uuid::Uuid;
+
+    let uuid = Uuid::parse_str("00112233445566778899AABBCCDDEEFF").unwrap();
+    let bin = Binary::from_uuid(uuid);
+    assert_eq!(bin.bytes, uuid.as_bytes());
+    assert_eq!(bin.subtype, BinarySubtype::Uuid);
```
patrickfreed, line 358: can we also add a test here that explicit specifies the standard representation?

src/tests/modules/bson.rs
```diff
@@ -345,3 +345,177 @@ fn debug_print() {
     assert_eq!(format!("{:?}", doc), normal_print);
     assert_eq!(format!("{:#?}", doc), pretty_print);
 }
+
+#[cfg(feature = "uuid-0_8")]
+#[test]
+fn test_binary_constructors() {
+    use crate::bson::UuidRepresentation;
```
patrickfreed, line 352: rather than repeating these `use` statements in each test, we can just add them at the beginning of the module.

src/tests/modules/bson.rs
```diff
@@ -345,3 +345,177 @@ fn debug_print() {
...
+#[test]
+fn test_binary_to_uuid() {
+    use crate::{bson::UuidRepresentation, de::Error};
+    use uuid::Uuid;
+
+    let uuid = Uuid::parse_str("00112233445566778899AABBCCDDEEFF").unwrap();
+    let bin = Binary::from_uuid(uuid);
+    // Error message for attempting to deserialize binary to a UUID using a non-standard
+    // representation when the representation is standard.
+    let error_msg = "expected binary subtype 3 when converting to UUID with a non-standard \
```
patrickfreed, line 398: these error messages can change from time to time, so it's best if we don't encode them in the tests. Otherwise we'll need to update this test each time we want to change the message. Instead, we can just verify that some error is returned. Ditto for the other tests.

src/lib.rs
```diff
@@ -305,7 +305,7 @@
 
 #[doc(inline)]
 pub use self::{
-    bson::{Array, Binary, Bson, DbPointer, Document, JavaScriptCodeWithScope, Regex, Timestamp},
+    bson::{Array, Binary, Bson, DbPointer, Document, JavaScriptCodeWithScope, Regex, Timestamp, UuidRepresentation},
```
patrickfreed, line 308: the `use` for `UuidRepresentation` needs to be gated behind the `uuid-0_8` feature flag, otherwise regular builds will fail (e.g. see [here](https://evergreen.mongodb.com/task_log_raw/rust_bson_compile_only__extra_rust_versions~min_compile_only_patch_fb18ed299db7a7bdf75569d5908ab582621667c6_6144bbe12a60ed25d50fdbcf_21_09_17_16_18_03/0?type=T#L169) for the failure in the evg patch).

patrickfreed (commented):
Looking good! I mostly just have documentation and testing suggestions

src/bson.rs
```diff
@@ -1054,6 +1057,137 @@ impl Binary {
     }
 }
 
+/// Enum of the possible representations to use when converting between Uuid and Binary.
```
patrickfreed, line 1060: can we linkify `Uuid` and `Binary` here? You might have to do a regular markdown link for `Uuid`. You can try the shorthand one, and if it doesn't work, `check-rustdoc.sh` will let you know.

Ditto for the other methods. I think its fine to leave the references to "UUID" alone in the docstrings for the enum variants, since those are more generally referring to UUIDs rather than the specific Rust `Uuid` type.

The "link all the things" approach is another one of the Rust API guidelines: https://rust-lang.github.io/api-guidelines/documentation.html#c-link

src/bson.rs
```diff
@@ -1054,6 +1057,137 @@ impl Binary {
...
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+impl Binary {
+    /// Serializes a UUID into BSON binary type
+    pub fn from_uuid(uuid: uuid::Uuid) -> Self {
+        Binary::from(uuid)
+    }
+    /// Serializes a UUID into BSON binary type and takes the desired representation as a parameter.
+    /// ``Binary::from_uuid_with_representation(uuid, UuidRepresentation::Standard)`` is equivalent
```
patrickfreed, line 1102: inline code blocks (e.g. `this`) only need single backticks. The double ones are reserved for whole sections of code
```
like this
```

patrickfreed (commented):
Looking good! Only have a couple more documentation suggestions

patrickfreed (approved):
LGTM! Great job with this!

abr-egn (approved):
LGTM!

PR #314, opened 2021-10-19
RUST-465 Introduce a `bson::Uuid` wrapper type

patrickfreed:
RUST-465

This PR introduces a `Uuid` type which can be used to correctly serialize to / from BSON, regardless of whether `bson::to_bson` or `bson::to_vec` is used. This type requires no feature flags to be used, though the existing flag can be enabled to allow for easy conversion between it and `uuid::Uuid` (similar to `bson::DateTime`).

As part of this, I updated the UUID spec implementation to operate on the new `bson::Uuid` type rather than `uuid::Uuid`, and as a result, was able to move that out from behind a feature flag too. Since these methods haven't been included in a release yet, this isn't a breaking change.

When serialized/deserializing from other formats, I opted to have `bson::Uuid` use `uuid::Uuid`'s implementation, rather than sticking with extended JSON (unlike `bson::DateTime`). Based on past experience, I think users will prefer this, but let me know if this doesn't seem like the right choice.


src/bson.rs
```diff
@@ -1076,148 +1073,6 @@ impl Binary {
     }
 }
 
-/// Enum of the possible representations to use when converting between [Uuid](https://docs.rs/uuid/0.8.2/uuid/) and [`Binary`].
```
patrickfreed, line 1079: this was all moved to the `uuid` module. The main difference is that these methods now work with `bson::Uuid` instead of `uuid::Uuid`, and that they also can be used without a feature flag.

src/ser/serde.rs
```diff
@@ -248,13 +251,28 @@ impl ser::Serializer for Serializer {
...
     fn serialize_newtype_struct<T: ?Sized>(
         self,
-        _name: &'static str,
+        name: &'static str,
         value: &T,
     ) -> crate::ser::Result<Bson>
     where
         T: Serialize,
     {
-        value.serialize(self)
```
patrickfreed, line 257: serialization works by serializing a newtype with a name that this serializer can recognize (`UUID_NEWTYPE_NAME`). When it sees this type, it uses the output of `uuid::Uuid`'s `Serialize` implementation (for the human readable serializer,  a string) and turns it into a subtype 4 Binary.

The same thing is done for the raw serializer, except that it expects `uuid::Uuid` to produce bytes.

src/tests/modules/bson.rs
```diff
@@ -396,146 +391,3 @@ fn debug_print() {
     assert_eq!(format!("{:?}", doc), normal_print);
     assert_eq!(format!("{:#?}", doc), pretty_print);
 }
-
-#[cfg(feature = "uuid-0_8")]
```
patrickfreed, line 400: these were moved to `src/uuid/test.rs`

src/uuid/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+impl Default for Uuid {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+impl Uuid {
+    /// Convert this [`Uuid`] to a [`uuid::Uuid`] from the [`uuid`](https://docs.rs/uuid/latest) crate.
```
patrickfreed: I figured it would be confusing to call this `from_uuid`, so I just appended the version number to the end to disambiguate. This will also be useful for when `uuid` releases another minor version.

src/uuid/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+        serializer.serialize_newtype_struct(UUID_NEWTYPE_NAME, &self.uuid)
+    }
+}
+
+impl<'de> Deserialize<'de> for Uuid {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        match deserializer.deserialize_newtype_struct(UUID_NEWTYPE_NAME, BsonVisitor)? {
```
patrickfreed, line 201: deserialization works by signaling to the deserializer, again through the newtype name, that we're looking for a UUID. The BSON deserializers ensure that a binary subtype 4 is present in the data and then pass it on. Non-BSON deserializers will just attempt to deserialize a `Bson` normally, and from there we deserialize a `uuid::Uuid` manually.

src/uuid/test.rs
```diff
@@ -0,0 +1,255 @@
...
+            subtype: BinarySubtype::UserDefined(100)
+        }
+    };
+    crate::from_document::<U>(other.clone()).unwrap_err();
+    let other_bytes = crate::to_vec(&other).unwrap();
+    crate::from_slice::<U>(&other_bytes).unwrap_err();
+}
+
+#[test]
+fn test_binary_constructors() {
```
patrickfreed, line 105: this test and below were moved, the above ones are new.

src/uuid/mod.rs
```diff
@@ -0,0 +1,475 @@
...
+//! [`crate::Uuid`]'s `serde` implementation is the same as `uuid::Uuid`'s
+//! for non-BSON formats such as JSON:
+//!
+//! ``` rust
+//! # #[cfg(feature = "uuid-0_8")]
+//! # {
+//! # use serde::{Serialize, Deserialize};
+//! # #[derive(Serialize, Deserialize)]
+//! # struct Foo {
+//! #   /// serializes as a String or subtype 0 BSON binary, depending
```
abr-egn, line 83: These comments are somewhat confusing to include in this example.

abr-egn (approved):
LGTM!  Minor comment nit only.

patrickfreed (commented):
One issue I noticed with the current approach to deserialization here is that our custom deserialization logic will be skipped if the `Uuid` is part of a `#[serde(flatten)]` chain (see https://github.com/serde-rs/serde/issues/2106). This means that `Uuid`s will erroneously allow deserialization from Binary values with subtype 0 in those cases. The driver currently deserializes all responses using `#[serde(flatten)]`, so in practice this means this behavior will affect most users. 

We could fix this issue by using the extended JSON format like we do for ObjectId and DateTime instead of the newtype pattern, but I think using `uuid::Uuid`'s serde implementation in non-BSON contexts is a valuable enough of a feature to warrant living with slightly too permissive deserialization in BSON. Note that this issue doesn't affect serialization, only deserialization. Thoughts?

src/uuid/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+impl Default for Uuid {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(feature = "uuid-0_8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
+impl Uuid {
+    /// Create a [`Uuid`] from a [`uuid::Uuid`](https://docs.rs/uuid/latest/uuid/struct.Uuid.html) from
```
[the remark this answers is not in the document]
patrickfreed: While they are technically redundant, I think they're useful for visibility. This is also consitent with our `to_chrono` and `from_chrono` methods on `bson::DateTime`.

src/uuid/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    CSharpLegacy,
+    /// The legacy representation of UUIDs in BSON used by the Java driver (binary subtype 0x03)
+    JavaLegacy,
+    /// The legacy representation of UUIDs in BSON used by the Python driver, which is the same
+    /// format as STANDARD, but has binary subtype 0x03
+    PythonLegacy,
+}
+
+impl Binary {
+    /// Serializes a [Uuid](https://docs.rs/uuid/0.8.2/uuid/) into BSON [`Binary`] type
```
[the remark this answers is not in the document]
patrickfreed: Good catch, fixed.

isabelatkinson (approved):

PR #316, opened 2021-10-27
deps: Remove `time` transitive dep from chrono

RoccoDev:
Currently, [`chrono`](https://lib.rs/chrono) depends on a really old version of [`time`](https://lib.rs/time), which is affected by CVE-2020-26235.  
This dependency is withheld by the `oldtime` feature in `chrono`, which is enabled by default.

Unfortunately, Cargo doesn't support disabling specific default features just yet, so the only way to get rid of the default feature is to re-import them all.

Code-wise, all tests seem to pass and I couldn't find any re-exports of the `time` crate (or even `chrono`'s replacements, for that matter), so this shouldn't break the API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ name = "bson"
 
 [dependencies]
 ahash = "0.7.2"
-chrono = { version = "0.4.15", features = ["clock", "std"], default-features = false }
+chrono = { version = "0.4.15", features = ["std"], default-features = false }
 rand = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order"] }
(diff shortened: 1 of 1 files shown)


patrickfreed:
Hi @RoccoDev, thanks for the PR! It looks like `chrono` itself is also vulnerable to this issue when it has the `clock` feature flag enabled. Could you update this to disable that feature as well? I don't believe we need it.

RoccoDev:
Disabling the feature worked, though it's still needed for tests as they use `Utc::now()`.

patrickfreed (approved):
LGTM, thanks for your contribution! Tagging in the other team members for review.

abr-egn (approved):
LGTM!

isabelatkinson (approved):
LGTM, can we update the PR title to begin with "minor:" before merging?

PR #317, opened 2021-10-28
RUST-977 Support RFC 3339 to `bson::DateTime` conversion without chrono feature flag

karmenliang:
[RUST-977](https://jira.mongodb.org/browse/RUST-977)

Add a method to convert a RFC 3339 formatting string to `bson::DateTime` even without the chrono feature flag.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -172,12 +172,13 @@ impl crate::DateTime {
             .to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true)
     }
 
+    /// Convert the given RFC 3339 format string to a [`DateTime`].
     pub fn from_rfc3339(s: &str) -> Result<Self, crate::de::Error> {
         match chrono::DateTime::<chrono::FixedOffset>::parse_from_rfc3339(s) {
-            Ok(d)  => Ok(Self::from_chrono(d)),
+            Ok(d) => Ok(Self::from_chrono(d)),
             Err(_) => Err(crate::de::Error::InvalidTimestamp {
-                message: format!("cannot convert {} to chrono::Datetime", s)
-            })
+                message: format!("cannot convert {} to chrono::Datetime", s),
+            }),
         }
     }
 }
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -167,13 +167,14 @@ impl crate::DateTime {
         self.0
     }
 
-    pub(crate) fn to_rfc3339(self) -> String {
+    pub(crate) fn to_rfc3339_string(self) -> String {
         self.to_chrono()
             .to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true)
     }
 
-    /// Convert the given RFC 3339 format string to a [`DateTime`].
-    pub fn from_rfc3339(s: impl AsRef<str>) -> Result<Self, crate::de::Error> {
+    /// Convert the given RFC 3339 format string to a [`DateTime`], truncating it to millisecond
+    /// precision.
+    pub fn parse_rfc3339_str(s: impl AsRef<str>) -> Result<Self, crate::de::Error> {
         match chrono::DateTime::<chrono::FixedOffset>::parse_from_rfc3339(s.as_ref()) {
             Ok(d) => Ok(Self::from_chrono(d)),
             Err(e) => Err(crate::de::Error::InvalidTimestamp {
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -167,12 +167,13 @@ impl crate::DateTime {
         self.0
     }
 
+    /// Convert this [`DateTime`] to an RFC 3339 formatted string.
     pub fn to_rfc3339_string(self) -> String {
         self.to_chrono()
             .to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true)
     }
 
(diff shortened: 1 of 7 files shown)


src/de/error.rs
```diff
@@ -15,6 +15,12 @@ pub enum Error {
     /// while decoding a UTF-8 String from the input data.
     InvalidUtf8String(string::FromUtf8Error),
 
+    /// An error encountered during the conversion of one datetime format to another.
+    InvalidTimestamp {
```
patrickfreed, line 19: rather than reusing the deserialization error here, I wonder if it would be better to introduce another error type for datetimes like we have for ObjectIds (`oid::Error`, returned from `ObjectId::parse_str`) and UUIDs (`uuid::Error`, returned from `Uuid::parse_str`). It's unfortunate that this crate has so many error types, but I think it's something we probably have to commit to at this point.

cc @abr-egn @isabelatkinson

src/datetime.rs
```diff
@@ -171,6 +171,16 @@ impl crate::DateTime {
         self.to_chrono()
             .to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true)
     }
+
+    /// Convert the given RFC 3339 format string to a [`DateTime`].
```
patrickfreed, line 175: can we add a note about truncating to millisecond precision, similar to the one above in `from_chrono`?

src/datetime.rs
```diff
@@ -171,6 +171,16 @@ impl crate::DateTime {
...
+    pub fn from_rfc3339(s: impl AsRef<str>) -> Result<Self, crate::de::Error> {
```
patrickfreed, line 176: Can we name this `parse_rfc3339_str`? That way it'll match `ObjectId::parse_str` and `Uuid::parse_str`. It's also somewhat closer to the familiar `chrono` API.

src/datetime.rs
```diff
@@ -171,6 +171,16 @@ impl crate::DateTime {
         self.to_chrono()
```
patrickfreed, line 171: Could we make `to_rfc3339` public as part of this work? I know it was omitted from the ticket description, but I think that was accidental. As part of that I guess we'll want to rename this to `to_rfc3339_string` to match the `parse_rfc3339_str`.

src/datetime.rs
karmenliang: Added in https://github.com/mongodb/bson-rust/pull/317/commits/9496a4f83481f037935d44bbcbeeedbb882e383a.

src/datetime.rs
karmenliang: Yup, done in https://github.com/mongodb/bson-rust/pull/317/commits/9496a4f83481f037935d44bbcbeeedbb882e383a.

src/de/error.rs
isabelatkinson: Yeah that seems fine to me for the sake of consistency

src/tests/datetime.rs
```diff
@@ -0,0 +1,21 @@
+use std::str::FromStr;
+
+#[test]
+fn rfc3339_to_datetime() {
```
isabelatkinson, line 4: We'll want to acquire the test lock in both of these tests:
`let _guard = LOCK.run_concurrently();`

src/datetime.rs
```diff
@@ -167,10 +167,21 @@ impl crate::DateTime {
         self.0
     }
 
-    pub(crate) fn to_rfc3339(self) -> String {
+    pub fn to_rfc3339_string(self) -> String {
```
patrickfreed, line 170: can we add a docstring to this method? As a general rule we try to document all new public facing types/methods. Right now we don't have an automated lint enabled to verify this because there are a decent number of older things left to be documented, but the eventual plan is to enable one.

src/tests/datetime.rs
```diff
@@ -0,0 +1,21 @@
...
+    let rfc = "2020-06-09T10:58:07.095Z";
+    let date = chrono::DateTime::<chrono::Utc>::from_str(rfc).unwrap();
+    assert_eq!(
+        crate::DateTime::parse_rfc3339_str(rfc).unwrap(),
+        crate::DateTime::from_chrono(date)
+    );
```
patrickfreed, line 10: can we add an assertion that converting the datetime back to a string matches the input?

src/de/error.rs
karmenliang: Added an error enum and switched over to using that instead in https://github.com/mongodb/bson-rust/pull/317/commits/0bd1b009c5d1a4bad0760a9a68bdc66936f77845.

src/datetime.rs
```diff
@@ -220,3 +233,27 @@ impl<T: chrono::TimeZone> From<chrono::DateTime<T>> for crate::DateTime {
         Self::from_chrono(x)
     }
 }
+
+/// Errors that can occur during [`DateTime`] construction and generation.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum Error {
```
patrickfreed, line 240: To allow users to refer to this type, we'll need to make the `datetime` module `pub` in `lib.rs`. Then users can say `bson::datetime::Result`.

patrickfreed (commented):
One last suggestion but otherwise LGTM!

isabelatkinson (approved):
lgtm pending patrick's last comment!

PR #318, opened 2021-11-01
RUST-1082 Implement `Serialize` and `Deserialize` for raw BSON types

patrickfreed:
RUST-1082

This PR adds `Serialize` and `Deserialize` implementations for the raw BSON types.

e.g. the following is now possible:

``` rust
#[derive(Debug, Deserialize, Serialize)]
struct MyData<'a> {
    name: &'a str,

    #[serde(borrow)]
    binary: RawBinary<'a>,

    #[serde(borrow)]
    doc: &'a RawDocument,

    #[serde(borrow)]
    some_field: RawBson<'a>,
}
```

The implementations behave as follows:
- `&RawDocument`:
  - for human readable formats: iterate over document and serialize each kvp
  - for binary formats: serialize the document as a newtype around a slice of bytes.
      - The BSON serializers handle this by interpreting the name of the newtype
  - Open Question: should this use something like `{ "$rawBsonDocument": <bytes> }` for binary, non-BSON formats? This will allow `RawBson` to round trip, but that's about it.
- `RawDocumentBuf`: same as `RawDocument`. In a future PR it will be able to deserialize from arbitrary maps, but that requires the ability to append.
- `RawBson`:
  - Defer to the wrapped type
   - The types that map directly to the serde data model (e.g. integers) are deserialized as-is, similar to `Bson`.
    - For everything else, the extended JSON serde data model equivalent is used, also similar to `Bson`
        - for many types, the model needed to be changed a little bit to ensure to extra allocations are done
- `&RawArray`
  - Same as `&RawDocument`, but uses a different newtype name to signal an array.
- `RawJavaScriptCodeWithScope`
  - serialize to/from: `{ "$code": <borrowed string>, "$scope": <borrowed string> }`
- `RawDbPointer`
  - serialize to/from: `{ "$ref": <borrowed string>, "$id": <objectid> }`
- `ObjectId`
  - Same as before, but it can now also deserialize from `{ "$oid": <12 byte slice> }`
- `RawBinary`
  - On human readable platforms, serializes to ext JSON format in serde data model
  - On binary formats, serializes to `{ "$binary": { "bytes": <borrowed bytes>, subType: <u8> } }` in the serde data model
  - In both cases, the corresponding serializer handles it and correctly encodes as BSON
  - In order for `RawBinary` to round trip on non-BSON formats, the separate bytes model (rather than base64 string model) was required. It is also much faster.
  - Unfortunately, we can't also update `Binary` to exhibit this behavior, as it would technically be a breaking change.
- `RawRegex`:
  - serialize to/from extended JSON format but uses borrowed instead of owned strings
- `DateTime`: updated to also deserialize from `{ "$date": <i64> }` to avoid a string allocation during deserialization.

### DIFF
--- a/src/de/raw.rs
+++ b/src/de/raw.rs
@@ -643,6 +643,10 @@ impl<'d, 'de> serde::de::Deserializer<'de> for DocumentKeyDeserializer<'d, 'de> 
         }
     }
 
+    fn is_human_readable(&self) -> bool {
+        false
+    }
+
     forward_to_deserialize_any! {
         bool char str bytes byte_buf option unit unit_struct string
         identifier newtype_struct seq tuple tuple_struct struct map enum
@@ -663,6 +667,10 @@ impl<'de> serde::de::Deserializer<'de> for FieldDeserializer {
         V: serde::de::Visitor<'de>,
     {
         visitor.visit_borrowed_str(self.field_name)
+    }
+
+    fn is_human_readable(&self) -> bool {
+        false
     }
 
     serde::forward_to_deserialize_any! {
@@ -750,6 +758,10 @@ impl<'de> serde::de::Deserializer<'de> for RawDocumentDeserializer<'de> {
         visitor.visit_borrowed_bytes(self.raw_doc.as_bytes())
     }
 
+    fn is_human_readable(&self) -> bool {
+        false
+    }
+
     serde::forward_to_deserialize_any! {
         bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string seq
         bytes byte_buf map struct option unit newtype_struct
@@ -818,6 +830,10 @@ impl<'de> serde::de::Deserializer<'de> for ObjectIdDeserializer {
         }
     }
 
+    fn is_human_readable(&self) -> bool {
+        false
+    }
+
     serde::forward_to_deserialize_any! {
         bool u8 u16 u32 u64 i8 i16 i32 i64 f32 f64 char str string seq
         bytes byte_buf map struct option unit newtype_struct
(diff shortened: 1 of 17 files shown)


src/raw/bson.rs
```diff
@@ -239,6 +245,307 @@ impl<'a> RawBson<'a> {
...
+        E: serde::de::Error,
+    {
+        crate::de::convert_unsigned_to_signed_raw(value.into())
+    }
+
+    fn visit_u64<E>(self, value: u64) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        crate::de::convert_unsigned_to_signed_raw(value)
```
abr-egn, line 328: Nit: might be worth just `use`ing `convert_unsigned_to_signed_raw`?

abr-egn (approved):
LGTM!  Extremely minor comments only.

On a larger scope, it's difficult for me to keep all the cogs in my head to get a sense of the overall machine operation - a walkthrough would be very welcome at some point.

patrickfreed (commented):
Yeah I agree, this code is pretty hard to read due to the sheer amount of boilerplate involved, especially across multiple serializer/deserializer implementations. I filed RUST-1098 to cover the work for writing up a doc that explains it all at a high level.

src/raw/bson.rs
patrickfreed: fixed

src/de/raw.rs
```diff
@@ -527,6 +674,99 @@ impl<'de> serde::de::Deserializer<'de> for FieldDeserializer {
...
+
+/// A `MapAccess` used to deserialize entire documents as chunks of bytes without deserializing
+/// the individual key/value pairs.
+struct RawDocumentAccess<'d> {
+    deserializer: RawDocumentDeserializer<'d>,
+
+    /// Whether the first key has been deserialized yet or not.
+    deserialized_first: bool,
+
+    /// Whether or not this document being deserialized is for anarray or not.
```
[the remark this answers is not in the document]
patrickfreed: fixed

isabelatkinson (approved):
lgtm with one typo fix

abr-egn (approved):

PR #321, opened 2021-11-09
RUST-1022 Support configuring `Serializer` and `Deserializer` to be non-human-readable

patrickfreed:
RUST-1022

This PR introduces two new options structs that enable users to configure `Serializer` and `Deserializer` (used in `to_bson` and `from_bson`) to be considered not-human-readable, just as the raw serializer/deserializer is (used in `to_vec` and `from_slice`). As part of that, several `to_x_with_options` functions were introduced.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -1062,19 +1062,30 @@ impl Display for Binary {
 
 impl Binary {
     pub(crate) fn from_extended_doc(doc: &Document) -> Option<Self> {
-        let binary = doc.get_document("$binary").ok()?;
-        let bytes = binary.get_str("base64").ok()?;
-        let bytes = base64::decode(bytes).ok()?;
-        let subtype = binary.get_str("subType").ok()?;
-        let subtype = hex::decode(subtype).ok()?;
+        let binary_doc = doc.get_document("$binary").ok()?;
 
-        if subtype.len() == 1 {
-            Some(Self {
-                bytes,
-                subtype: subtype[0].into(),
-            })
+        if let Some(bytes) = binary_doc.get_str("base64").ok() {
+            let bytes = base64::decode(bytes).ok()?;
+            let subtype = binary_doc.get_str("subType").ok()?;
+            let subtype = hex::decode(subtype).ok()?;
+            if subtype.len() == 1 {
+                Some(Self {
+                    bytes,
+                    subtype: subtype[0].into(),
+                })
+            } else {
+                None
+            }
         } else {
-            None
+            // in non-human-readable mode, RawBinary will serialize as
+            // { "$binary": { "bytes": <bytes>, "subType": <i32> } };
+            let binary = binary_doc.get_binary_generic("bytes").ok()?;
+            let subtype = binary_doc.get_i32("subType").ok()?;
+
+            Some(Self {
+                bytes: binary.clone(),
+                subtype: u8::try_from(subtype).ok()?.into(),
+            })
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/bson.rs
```diff
@@ -1062,19 +1062,30 @@ impl Display for Binary {
 
 impl Binary {
     pub(crate) fn from_extended_doc(doc: &Document) -> Option<Self> {
```
patrickfreed, line 1064: `RawBinary` uses `is_human_readable` to determine if it should serialize to extjson or a more compact form. Since the `Bson` based serializer/deserializer can now present itself as not-human-readable, we need to add some special handling to them for when BSON types use a different serialization output.

src/de/serde.rs
```diff
@@ -562,11 +562,58 @@ where
...
+    /// Create a builder struct used to construct a [`DeserializerOptions`].
+    pub fn builder() -> DeserializerOptionsBuilder {
+        DeserializerOptionsBuilder {
+            options: Default::default(),
+        }
+    }
+}
+
+/// Builder used to construct a [`DeserializerOptions`].
+pub struct DeserializerOptionsBuilder {
```
patrickfreed, line 588: Since it was just a single option, I opted to create the builder manually rather than bringing in `typed-builder`.

src/raw/bson.rs
```diff
@@ -610,20 +610,11 @@ impl<'a> TryFrom<RawBson<'a>> for Bson {
 /// A BSON binary value referencing raw bytes stored elsewhere.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct RawBinary<'a> {
```
patrickfreed, line 612: I updated this type to match the regular `Binary`, which has all `pub` fields. Note that this isn't `non_exhaustive` (neither is `Binary`).

src/de/serde.rs
```diff
@@ -659,11 +713,25 @@ impl<'de> de::Deserializer<'de> for Deserializer {
...
         }
     }
 
+    #[inline]
+    fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        match self.value {
+            Some(Bson::ObjectId(oid)) if !self.is_human_readable() => {
```
abr-egn, line 728: This special-casing of `ObjectId` seems like it might fit better as a case in `deserialize_any`?

src/ser/serde.rs
```diff
@@ -258,21 +303,40 @@ impl ser::Serializer for Serializer {
...
-                Bson::String(s) => {
-                    // the serializer reports itself as human readable, so `Uuid` will
-                    // serialize itself as a string.
-                    let uuid = crate::Uuid::parse_str(s).map_err(Error::custom)?;
-                    Ok(Bson::Binary(uuid.into()))
+        match name {
+            UUID_NEWTYPE_NAME => {
+                match value.serialize(self)? {
+                    Bson::String(s) => {
+                        // the serializer reports itself as human readable, so `Uuid` will
```
abr-egn, line 310: This is no longer an invariant, no?

patrickfreed: good catch, forgot about this. Fixed and added a test case for this.

src/de/serde.rs
patrickfreed: The reason this is here is because in the deserialize implementation of `ObjectId`, it checks to see if the format is human readable and, if not, uses `deserialize_bytes` instead of `deserialize_any`. The serializer then uses that as a hint that `ObjectId` is being deserialized. We can't do this in `deserialize_any` because there's no way to hint to the deserializer that an `ObjectId` is being deserialized as opposed to say `Bson`, which needs the `{ "$oid": <hex string> }` format to know its deserializing an ObjectId.

abr-egn (approved):

isabelatkinson (approved):

PR #324, opened 2021-11-12
minor: hide work-in-progress raw BSON API

patrickfreed:
This PR hides the not yet complete raw BSON API from the public API in preparation for the 2.1 release. The raw BSON stuff will land in 2.2 instead.

I've gone ahead and made a `2.1.x` branch which this will merge into, since master won't need to have these changes in it.

evg patch: https://evergreen.mongodb.com/version/618ed47757e85a14b56c83e1

### DIFF
--- a/serde-tests/test.rs
+++ b/serde-tests/test.rs
@@ -29,14 +29,6 @@ use bson::{
     DeserializerOptions,
     Document,
     JavaScriptCodeWithScope,
-    RawArray,
-    RawBinary,
-    RawBson,
-    RawDbPointer,
-    RawDocument,
-    RawDocumentBuf,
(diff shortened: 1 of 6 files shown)


serde-tests/test.rs
```diff
@@ -29,14 +29,6 @@ use bson::{
     DeserializerOptions,
     Document,
     JavaScriptCodeWithScope,
-    RawArray,
```
patrickfreed, line 32: these tests had to be removed because they're implemented in a separate crate and thus only have access to the public API of `bson`

abr-egn (approved):

isabelatkinson (approved):

PR #326, opened 2021-11-18
RUST-1045 Support appending to `RawDocumentBuf`

patrickfreed:
RUST-1045

This PR adds a `RawDocumentBuf::append` method which enables `RawDocumentBuf`s to be created and mutated without having to go through `Document`.

To achieve this, a number of public API additions/changes were needed:
- An `OwnedRawBson` type was added
  - This is essentially the same as `Bson`, except the cases that hold documents/arrays use `RawDocumentBuf` and `RawArrayBuf` (also new type)
  - A `OwnedRawJavaScriptCodeWithScope` was also added, since the existing `JavaScriptCodeWithScope` stored its scope as a `Document`
- A `RawDocumentBuf` can now de deserialized from arbitrary maps
  - Ditto for `RawArrayBuf` and arbitrary sequences
- `RawDocumentBuf::new` was renamed to `RawDocumentBuf::from_bytes`
  - This allowed `RawDocumentBuf::new()` to be added for constructing brand new documents
  - `RawDocument::new` was also renamed to be consistent with this
- `RawDocumentBuf` and `RawArrayBuf` now implement `FromIterator`
- `RawBson` and `OwnedRawBson` both implement a number of `From` traits


serde-tests/json.rs
```diff
@@ -0,0 +1,132 @@
...
+use serde_json::json;
+
+use super::AllTypes;
+
+use bson::{doc, Bson, JavaScriptCodeWithScope, OwnedRawBson, RawArrayBuf, RawDocumentBuf};
+
+use serde::{Deserialize, Serialize};
+
+#[test]
+fn all_types_json() {
```
patrickfreed, line 11: this test was moved without modification from `test.rs`. The one after it is new though.

src/de/raw.rs
```diff
@@ -648,13 +648,20 @@ impl<'d, 'de> serde::de::Deserializer<'de> for DocumentKeyDeserializer<'d, 'de>
         }
     }
 
+    fn deserialize_newtype_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value>
```
patrickfreed, line 651: This was needed to support deserializing the `Cow` wrappers used in deserialization now. It just forwards deserialization to the wrapped type.

src/raw/bson.rs
```diff
@@ -244,245 +253,44 @@ impl<'a> RawBson<'a> {
             _ => None,
         }
     }
-}
-
-/// A visitor used to deserialize types backed by raw BSON.
-pub(crate) struct RawBsonVisitor;
```
patrickfreed, line 250: This was moved to `src/raw/serde.rs` and slightly modified so that it could be shared with `OwnedRawBson`.

src/raw/document_buf.rs
```diff
@@ -136,19 +159,169 @@ impl RawDocumentBuf {
...
+    /// let expected = doc! {
+    ///     "a string": "some string",
+    ///     "an integer": 12_i32,
+    ///     "a document": { "a key": true },
+    /// };
+    ///
+    /// assert_eq!(doc.to_document()?, expected);
+    /// # Ok::<(), Error>(())
+    /// ```
+    pub fn append(&mut self, key: impl Into<String>, value: impl Into<OwnedRawBson>) {
```
patrickfreed, line 195: Since this only appends to the end, we can't call it `insert` (which would imply key lookup). I opted for something different than `push` though which seems to be coupled with stacks/lists rather than key-value stores. Let me know if these seems right.

src/raw/owned_bson.rs
```diff
@@ -0,0 +1,504 @@
...
+
+use super::{
+    serde::{OwnedOrBorrowedRawBson, OwnedOrBorrowedRawBsonVisitor},
+    Error,
+    Result,
+};
+
+/// A BSON value backed by owned raw BSON bytes.
+#[derive(Debug, Clone, PartialEq)]
+pub enum OwnedRawBson {
```
patrickfreed, line 34: This type was needed to have something to append to documents. I originally tried using a `RawBson<'a>` for this, but it was clunky / awkward, and it would potentially make the `rawdoc!` macro impossible.

Regarding the name, I originally opted for `RawBson` being the owned one and `RawBsonRef` being the borrowed one, but I ended up going with `OwnedRawBson` and `RawBson` to match how we added the suffix to the owned variants of `RawDocument` and `RawArray`. I'm still not certain this is the correct naming scheme for `RawBson` though, would love to hear your all's thoughts on this.

src/raw/owned_bson.rs
```diff
@@ -0,0 +1,504 @@
...
+            OwnedRawBson::JavaScriptCodeWithScope(s) => Some(RawJavaScriptCodeWithScope {
+                code: s.code.as_str(),
+                scope: &s.scope,
+            }),
+            _ => None,
+        }
+    }
+
+    /// Gets a [`RawBson`] value referencing this owned raw BSON value.
+    pub fn as_raw_bson(&self) -> RawBson<'_> {
```
patrickfreed, line 272: Unfortunately, we can't implement `Deref`, `AsRef`, `Borrow`, or `ToOwned` between these two types, the first three because the trait requires a reference to be returned and the last one due to the blanket implementation for `Clone` types. See https://github.com/rust-lang/rust/issues/44950 for more info.

src/raw/serde.rs
```diff
@@ -0,0 +1,527 @@
...
+            o => Err(serde::de::Error::custom(format!(
+                "expected raw array, instead got {:?}",
+                o
+            ))),
+        }
+    }
+}
+
+/// A visitor used to deserialize types backed by raw BSON.
+pub(crate) struct OwnedOrBorrowedRawBsonVisitor;
```
patrickfreed, line 180: This is essentially the same as the old `RawBsonVisitor`, but it now can handle owned stuff (e.g. `visit_string`, `visit_byte_buf`) and arbitrary maps and sequences.

src/raw/owned_bson.rs
```diff
@@ -0,0 +1,504 @@
...
+    pub fn as_str(&self) -> Option<&'_ str> {
+        match self {
+            OwnedRawBson::String(s) => Some(s),
+            _ => None,
+        }
+    }
+
+    /// Gets a reference to the [`RawArrayBuf`] that's wrapped or returns `None` if the wrapped
+    /// value isn't a BSON array.
+    pub fn as_array(&self) -> Option<&'_ RawArray> {
```
abr-egn, line 127: Since there are now mutating methods for those, I think it'd be good to have `as_array_mut -> Option<&mut RawArrayBuf>` and similar for `as_document_mut`.

src/raw/array_buf.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use crate::{RawArray, RawBson, RawDocumentBuf};
+
+use super::{owned_bson::OwnedRawBson, serde::OwnedOrBorrowedRawArray, RawArrayIter};
+
+/// An owned BSON array value (akin to [`std::path::PathBuf`]), backed by a buffer of raw BSON
+/// bytes. This type can be used to construct owned array values, which can be used to append to
+/// [`RawDocumentBuf`] or as a field in a `Deserialize` struct.
+///
+/// Iterating over a [`RawArrayBuf`] yields either an error or a key-value pair that borrows from
```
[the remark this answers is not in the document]
patrickfreed: fixed

src/raw/owned_bson.rs
patrickfreed: good idea, added

src/raw/owned_bson.rs
patrickfreed: I think I agree. I originally thought it would be confusing to have `RawDocument` typically be borrowed and `RawBson` be owned, but `OwnedRawBson` does seem odd enough to warrant the inconsistency. Also, `RawDocument` _technically_ isn't a borrowed type anyways, though it's always accessed as a wide-pointer (`&RawDocument`).

abr-egn (approved):
LGTM modulo naming questions.

src/raw/owned_bson.rs
patrickfreed: Updated to be `RawBsonRef` and `RawBson`. I also updated `RawBinary<'a>` et al to be `RawBinaryRef<'a>` and what not. `OwnedRawJavaScriptCodeWithScope` dropped the `Owned` prefix. I also renamed `RawBsonRef::to_owned_raw_bson` to just `RawBsonRef::to_raw_bson`.

abr-egn (approved):

isabelatkinson (approved):

PR #329, opened 2021-12-01
RUST-1109 Implement `rawbson!` and `rawdoc!` macros

patrickfreed:
RUST-1109

This PR implements the `rawbson!` and `rawdoc!` macros to allow easier creation of raw BSON types. They are largely copy/pasted from the existing `bson!` and `doc!` macros with just a few modifications to build raw BSON. This PR also updates a number of examples and tests to use `rawdoc!` instead of building `RawDocumentBuf` values manually or from `Document`.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -212,3 +212,212 @@ macro_rules! doc {
         object
     }};
 }
+
+/// Construct a [`crate::RawBson`] value from a literal.
+///
+/// ```rust
+/// use bson::rawbson;
+///
+/// let value = rawbson!({
+///     "code": 200,
+///     "success": true,
+///     "payload": {
+///       "some": [
+///           "pay",
+///           "loads",
+///       ]
+///     }
+/// });
+/// ```
+#[macro_export]
+macro_rules! rawbson {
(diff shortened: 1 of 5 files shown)


src/macros.rs
```diff
@@ -212,3 +212,212 @@ macro_rules! doc {
...
+    //////////////////////////////////////////////////////////////////////////
+    // TT muncher for parsing the inside of an array [...]. Produces a
+    // RawArrayBuf containing the elements.
+    //
+    // Must be invoked as: bson!(@array [] $($tt)*)
+    //////////////////////////////////////////////////////////////////////////
+
+    // Finished with trailing comma.
+    (@array [$($elems:expr,)*]) => {
+        <$crate::RawArrayBuf as std::iter::FromIterator::<$crate::RawBson>>::from_iter(vec![$($elems,)*])
```
patrickfreed, line 243: This isn't as efficient as it could be, but it was a lot easier then rewriting this to build up the `RawArrayBuf` like we do for `RawDocumentBuf`. I field RUST-1118 to cover the work for doing this without the `Vec` allocation.

isabelatkinson (approved):

PR #330, opened 2021-12-02
RUST-1119 Add `bson::to_raw_document_buf` function

patrickfreed:
RUST-1119

This PR adds the `bson::to_raw_document_buf` convenience function. I originally set out to also add `bson::to_raw_bson` as well (RUST-1110), but it looked like it was going to be a lot of effort for functionality that might not be all that useful. For now, I've split off just this function (which definitely will be useful) and am putting the `to_raw_bson` work on the backburner. Let me know if that sounds okay.

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -289,6 +289,7 @@ where
 ///
 /// ```rust
 /// use serde::Serialize;
+/// use bson::rawdoc;
 ///
 /// #[derive(Serialize)]
 /// struct Cat {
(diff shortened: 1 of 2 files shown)


src/ser/mod.rs
```diff
@@ -283,3 +284,27 @@ where
...
+///
+/// #[derive(Serialize)]
+/// struct Cat {
+///     name: String,
+///     age: i32
+/// }
+///
+/// let cat = Cat { name: "Garfield".to_string(), age: 43 };
+/// let doc = bson::to_raw_document_buf(&cat)?;
+/// // assert_eq!(doc, rawdoc! { "name": "Garfield", "age": 43 })
```
patrickfreed, line 301: once #329 is merged, I'll rebase and uncomment this line.

abr-egn (approved):

isabelatkinson (approved):

PR #343, opened 2022-01-26
RUST-939 Run BSON fuzzer in Evergreen

patrickfreed:
RUST-939

This PR begins running the existing fuzzer in evergreen, and it also runs some new fuzzing targets intended for the new raw BSON API. As part of that, it found some crashes which have also been fixed in this PR.


abr-egn (approved):
LGTM!

isabelatkinson (approved):

PR #344, opened 2022-02-16
RUST-1175 Introduce owned iterator for `RawArrayBuf`, `"unstable"` feature flag

patrickfreed:
RUST-1175

This PR introduces an iterator type that owns the underlying `RawArrayBuf` for use in cursors in the driver. The main advantage of this type over the existing, borrowed iterators is that by owning the underlying data, it can be paused and resumed in between polls of the cursor.

My first idea was simply to implement `IntoIterator` on `RawArrayBuf` and have it return this type, but that seemed like possibly a bad idea, since this iterator needs to copy out each element, despite owning the underlying data. This would make it really easy for users to do something like:

``` rust
for i in doc {
    println!("{:?}", i);
}
```
and not realizing they're making needless copies. To that end, I introduced an `into_copying_iter()` method on `RawArrayBuf`, which does the same thing but prevents users from accidentally making extra copies. Since this didn't seem like generally useful API, I gated it behind an "unstable" feature flag, with the warning that its API is not stable. A slight wrinkle in that is we need it in the driver, and since Cargo unions all the feature flags for a given dependency, all users of the driver will have the `"unstable"` flag enabled for them, which sort of defeats it's purpose.

I don't think we have any great options here for introducing private BSON functionality. IMO, the `IntoIterator` stuff bleeds into the actual user's interface a bit too much, hence why I went this route, but I also see the flaws in this approach too. Not exactly sure what's the best path here, so would appreciate any opinions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,11 @@ uuid-0_8 = []
 # all optional dependencies.
 # serde_with
 
-# # private feature flag containing unstable functionality for special use in the
-# # mongodb crate. This feature flag should not be enabled by other users.
-# __private = []
+# feature flag containing unstable functionality for special use in the
+# mongodb crate. Any API exposed by using this feature flag is subject
+# to change at any time and should not be relied upon in stable
+# applications.
+unstable = []
 
 [lib]
 name = "bson"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ uuid-0_8 = []
 # all optional dependencies.
 # serde_with
 
+# # private feature flag containing unstable functionality for special use in the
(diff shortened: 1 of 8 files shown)


abr-egn (approved):
I agree that the combination of unstable feature (even though it'll be selected for anyone using this through the driver crate) and undocumented seem like the better way to go; that makes it not accidentally discoverable and clear it shouldn't be relied on.

patrickfreed:
I benchmarked a few cases using this type vs a `VecDeque<RawDocumentBuf>`, and as it turns out the performance is basically the same in one case and faster in the `VecDeque` case (not sure why at the moment). Given these results, I'm going to close this PR since it seems to not be necessary.

PR #348, opened 2022-03-28
RUST-1243 Handle enum keys when deserializing a map from binary

abr-egn:
RUST-1243

Prior to this PR, all keys were treated as strings, so the deserialization would fail with "invalid type: string \"Key\", expected enum Enum".  This bug doesn't occur when deserializing from the parsed Bson types because that one uses the main deserializer to parse the key (which knows how to handle enum values) rather than a specialized one.

### DIFF
--- a/serde-tests/test.rs
+++ b/serde-tests/test.rs
@@ -410,7 +410,7 @@ fn hashmap_enum_key() {
     };
     let doc = doc! {
         "map": {
-            "Baz": "2",  
+            "Baz": "2",
         },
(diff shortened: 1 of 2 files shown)


patrickfreed (approved):
great job hunting this down so quick!

patrickfreed:
Looks like lint needs to be fixed up but besides that LGTM holds (no need for another round of review from me)

isabelatkinson (approved):

PR #352, opened 2022-04-05
RUST-1129 Make the dependency on `chrono` optional and provide equivalent `time` interop

abr-egn:
RUST-1129

This PR swaps out the internal operations that used `chrono` for equivalent ones using `time`, makes the `chrono` dependency optional, and adds optional interop methods for `time` structs.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -181,9 +181,9 @@ axes:
   - id: "extra-rust-versions"
     values:
       - id: "min"
-        display_name: "1.51 (minimum supported version)"
+        display_name: "1.53 (minimum supported version)"
         variables:
-          RUST_VERSION: "1.51.0"
+          RUST_VERSION: "1.53.0"
       - id: "nightly"
         display_name: "nightly"
         variables:
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -181,9 +181,9 @@ axes:
   - id: "extra-rust-versions"
     values:
       - id: "min"
-        display_name: "1.48 (minimum supported version)"
+        display_name: "1.51 (minimum supported version)"
         variables:
-          RUST_VERSION: "1.48.0"
+          RUST_VERSION: "1.51.0"
       - id: "nightly"
         display_name: "nightly"
         variables:
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -437,9 +437,10 @@ impl Bson {
                 })
             }
             Bson::ObjectId(v) => json!({"$oid": v.to_hex()}),
-            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.to_time().year() <= 99999 => {
+            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.to_time().year() <= 9999 => {
                 json!({
-                    "$date": v.to_rfc3339_string(),
+                    // Unwrap safety: timestamps in the guarded range can always be formatted.
+                    "$date": v.try_to_rfc3339_string().unwrap(),
                 })
             }
             Bson::DateTime(v) => json!({
@@ -601,7 +602,8 @@ impl Bson {
             },
             Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.to_time().year() <= 9999 => {
                 doc! {
-                    "$date": v.to_rfc3339_string(),
+                    // Unwrap safety: timestamps in the guarded range can always be formatted.
+                    "$date": v.try_to_rfc3339_string().unwrap(),
                 }
             }
             Bson::DateTime(v) => doc! {
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -33,7 +33,7 @@ use crate::{
     oid::{self, ObjectId},
     spec::{BinarySubtype, ElementType},
     Decimal128,
-    RawBinaryRef,
+    RawBinaryRef, DateTime,
 };
 
 /// Possible BSON value types.
@@ -371,9 +371,6 @@ impl From<Bson> for Value {
     }
 }
 
-// 9999-12-31 23:59:59.999
-const MAX_RFC3339_TIMESTAMP: i64 = 253402300799999;
-
 impl Bson {
     /// Converts the Bson value into its [relaxed extended JSON representation](https://docs.mongodb.com/manual/reference/mongodb-extended-json/).
     ///
@@ -440,8 +437,7 @@ impl Bson {
                 })
             }
             Bson::ObjectId(v) => json!({"$oid": v.to_hex()}),
-            Bson::DateTime(v)
-                if v.timestamp_millis() >= 0 && v.timestamp_millis() <= MAX_RFC3339_TIMESTAMP =>
+            Bson::DateTime(v) if dt_use_rfc3339(&v) =>
             {
                 json!({
                     "$date": v.to_rfc3339_string(),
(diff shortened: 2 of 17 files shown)


src/bson.rs
```diff
@@ -364,6 +371,9 @@ impl From<Bson> for Value {
     }
 }
 
+// 9999-12-31 23:59:59.999
+const MAX_RFC3339_TIMESTAMP: i64 = 253402300799999;
```
abr-egn, line 375: Because `to_time` clamps rather than erroring (following the behavior of `to_chrono`), and `time::OffsetDateTime` maxes out at the year 9999, the conditionals checking `year` were always passing, leading to a fun off-by-one-millisecond failure in the corpus tests.  Checking against actual timestamp avoids conversion issues.

src/datetime.rs
```diff
@@ -210,26 +278,30 @@ impl crate::DateTime {
 
     /// Convert this [`DateTime`] to an RFC 3339 formatted string.
     pub fn to_rfc3339_string(self) -> String {
-        self.to_chrono()
-            .to_rfc3339_opts(chrono::SecondsFormat::AutoSi, true)
+        use std::result::Result;
+        match self.to_time().format(&Rfc3339) {
+            Result::Ok(s) => s,
+            Result::Err(e) => e.to_string(),
```
abr-egn, line 284: I'm not thrilled about this but I couldn't think of a better option that wouldn't be a breaking change.

src/extjson/models.rs
```diff
@@ -256,16 +255,13 @@ impl DateTime {
...
-                    chrono::DateTime::parse_from_rfc3339(date.as_str())
-                        .map_err(|_| {
-                            extjson::de::Error::invalid_value(
-                                Unexpected::Str(date.as_str()),
-                                &"rfc3339 formatted utc datetime",
-                            )
-                        })?
-                        .into();
-                Ok(crate::DateTime::from_chrono(datetime))
+                let datetime = crate::DateTime::parse_rfc3339_str(date.as_str()).map_err(|_| {
```
abr-egn, line 258: Here and elsewhere I changed it from the "parse into external datetime, convert to bson datetime" pattern to just directly using the methods `bson::DateTime` provides.

src/bson.rs
```diff
@@ -430,7 +437,7 @@ impl Bson {
                 })
             }
             Bson::ObjectId(v) => json!({"$oid": v.to_hex()}),
-            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.to_chrono().year() <= 99999 => {
+            Bson::DateTime(v) if v.timestamp_millis() >= 0 && v.to_time().year() <= 99999 => {
```
abr-egn, line 440: I'm not sure if the year constraint here is correct - it's different than the one below.

kmahar (commented):
looking good! just a few questions/thoughts

src/datetime.rs
abr-egn: From the source, the `time` [method](https://docs.rs/time/0.3.9/src/time/formatting/formattable.rs.html#158-236) fails if the year is outside [0,9999]; the chrono [equivalent](https://docs.rs/chrono/0.4.19/src/chrono/datetime.rs.html#436-476) doesn't have that check.

The `time` method has some other potential failure cases as well since it's a more general method, but for the specific purpose of formatting an `OffsetDateTime` I don't think they can happen.

.evergreen/config.yml
```diff
@@ -181,9 +181,9 @@ axes:
   - id: "extra-rust-versions"
     values:
       - id: "min"
-        display_name: "1.48 (minimum supported version)"
+        display_name: "1.53 (minimum supported version)"
         variables:
-          RUST_VERSION: "1.48.0"
+          RUST_VERSION: "1.53.0"
```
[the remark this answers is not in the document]
abr-egn: Filed RUST-1263.

src/datetime.rs
patrickfreed: The `time` crate actually has a `large-dates` flag which expands this to +-99,999, which we may want to consider enabling here, as MongoDB technically supports any date that can be represented as an `i64` from the Unix epoch. Should we enable that flag, we'll want to let the maintainer know of our use case [here](https://github.com/time-rs/time/discussions/306) so that the feature flag stays around in the future.

Also, rather than erroring, it may be more appropriate to panic here. Could we test some inputs that error in `time` and see if they panic in `chrono`? If they do it might be okay for us to panic here as well. Another option would be to cap the date to `time`'s maximum value and use that for the date here. I think we do that in a few places already with `chrono`'s maximum date. If none of those work, then I agree we probably should deprecate this method and introduce a new one, since I think including error information as a String could be pretty cumbersome or dangerous.

src/serde_helpers.rs
```diff
@@ -185,6 +191,49 @@ pub mod u64_as_f64 {
...
+/// #[derive(Serialize, Deserialize)]
+/// struct Event {
+///     #[serde(with = "time_offsetdatetime_as_bson_datetime")]
+///     pub date: time::OffsetDateTime,
+/// }
+/// # }
+/// ```
+#[cfg(feature = "time-0_3")]
+#[cfg_attr(docsrs, doc(cfg(feature = "time-0_3")))]
+pub mod time_offsetdatetime_as_bson_datetime {
```
patrickfreed, line 211: I know this matches the existing pattern, but now that I'm looking at this again, I think we'll want to include the versioning in the names here as well. That way we wont run into clashes if need a `time-0_4` or just `time` feature.

abr-egn: Done.

src/datetime.rs
abr-egn: The `large-dates` flag is already used, actually (it's needed for the y10k comparisons to be meaningful), but it doesn't affect the range of dates supported by the rfc3339 format.  I went with the panic and deprecate option because outputting truncated values seems even more likely to cause bad downstream behavior than error strings.

kmahar: that solution SGTM. since we use the flag already it probably seems worth chiming in on that discussion linked above to let them know why

kmahar (approved):

src/datetime.rs
patrickfreed: And as a last check, are we okay with the risk of breaking user code with this? It's technically possible that a user has a super large date in their DB somewhere that used to format okay but will panic now via this method or fail to serialize via the serde helper.

abr-egn: I think the risk is okay here - we're being loud about this change with the deprecation warning, and I think we should also call it out in the patch notes.  It's also worth considering that [the RFC itself](https://datatracker.ietf.org/doc/html/rfc3339) is explicit about only applying to years in the 0000-9999 range and tha the year field is fixed at four digits, so the chrono implementation is producing output that's probably not going to be accepted elsewhere.

patrickfreed: Oh nice, I didn't know that about the RFC, that definitely makes this change seem fine to me then.

kmahar: agree it seems fine 👍

isabelatkinson (approved):
lgtm mod patrick's renaming comment!

src/datetime.rs
```diff
@@ -140,30 +160,78 @@ impl crate::DateTime {
...
+    /// Convert the given `time::OffsetDateTime` into a `bson::DateTime`, truncating it to
+    /// millisecond precision.
     ///
-    /// Note: Not every BSON datetime can be represented as a [`chrono::DateTime`]. For such dates,
-    /// [`chrono::MIN_DATETIME`] or [`chrono::MAX_DATETIME`] will be returned, whichever is closer.
+    /// If the provided time is too far in the future or too far in the past to be represented
+    /// by a BSON datetime, either [`DateTime::MAX`] or [`DateTime::MIN`] will be
+    /// returned, whichever is closer.
+    #[cfg(feature = "time-0_3")]
+    pub fn from_time(dt: time::OffsetDateTime) -> Self {
```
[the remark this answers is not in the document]
abr-egn: Done.

patrickfreed (approved):

PR #358, opened 2022-06-03
Updated uuid version, fixed clippy lints

aftix:
I was needing to use this crate for a WASM target, but the version of `uuid`
in use does not have the `js` feature, while later ones do. I updated the dependency.
In the process, I had to fix clippy default lints to get my CI to work, so I've also
included that in this PR. I have run the test suite for all features and every test
still passes. The changes to the code are just adding `Eq` to derive macros and
in two places removing unused lifetime parameters from an `impl` block.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ indexmap = "1.6.2"
 hex = "0.4.2"
 base64 = "0.13.0"
 lazy_static = "1.4.0"
-uuid = { version = "0.8.1", features = ["serde", "v4"] }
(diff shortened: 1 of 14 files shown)


isabelatkinson:
Hi @aftix, thank you for creating this pull request! We've had a lot of interest in bumping our `uuid` dependency and plan to do this in the near future; however, due to the way current API interacts with previous versions of `uuid`, we have some design considerations to discuss as a team before making the necessary changes. You can follow RUST-1305 for updates on when this work will be completed.

PR #361, opened 2022-06-16
RUST-1198 Support creating bson::DateTime from a given year/month/date/hour/minute/second/millisecond

sanav33:
Added function to create a specific bson::DateTime type from from a given year/month/date/hour/minute/second/millisecond without using chrono.


src/datetime.rs
```diff
@@ -131,6 +131,28 @@ impl crate::DateTime {
...
+    /// `bson::DateTime`, truncating it to millisecond precision.
+    pub fn from_specific_datetime(
+        year: i32,
+        month: u8,
+        day: u8,
+        hour: u8,
+        min: u8,
+        sec: u8,
+        millis: u16,
+    ) -> Result<Self> {
```
isabelatkinson, line 144: I think the intent of this ticket was to have a separate method for each time increment listed here but is that what you were thinking @patrickfreed? Sorry if this was unclear!

src/datetime.rs
```diff
@@ -105,6 +105,78 @@ use serde_with::{DeserializeAs, SerializeAs};
...
+        self.second = Some(second);
+        self
+    }
+
+    pub fn millisecond(&mut self, millisecond: u16) -> &mut Self {
+        self.millisecond = Some(millisecond);
+        self
+    }
+
+    pub fn builder(&mut self) -> Result<DateTime> {
```
isabelatkinson, line 162: nit: can we call this method `build`? that's the naming convention for the builders we have in the driver

src/tests/modules/bson.rs
```diff
@@ -265,6 +267,34 @@ fn from_external_datetime() {
         assert_millisecond_precision(from_chrono);
     }
 
+    {
```
isabelatkinson, line 270: can we put these in their own separate test (something like `from_datetime_builder`)? it's a little easier to debug failures that way

isabelatkinson (commented):
looks good! just a few small comments

src/tests/modules/bson.rs
sanav33: Fixed!

src/datetime.rs
```diff
@@ -1,12 +1,63 @@
...
+///
+/// fn main() -> Result<(), Box<dyn Error>> {
+///     let dt = RawDateTime::builder().year(1998).month(2).day(12).minute(1).millisecond(23).build().to_bson()?;
+///     let expected = DateTime::parse_rfc3339_str("1998-02-12T00:01:00.023Z")?;
+///     assert_eq!(dt, expected);
+///     Ok(())
+/// }
+/// ```
+#[derive(TypedBuilder)]
+pub struct RawDateTime {
```
patrickfreed, line 28: Oh hmm, this is a bit of an issue here. I don't think we necessarily want to be introducing a new date time type just for this method, but `TypedBuilder` won't let us customize the `build` method to produce something other than what we're deriving the builder from (see https://github.com/idanarye/rust-typed-builder/issues/52).

Do you still have the manual implementation laying around? We may want to use that actually :grimacing: (sorry!)

What do you guys think @isabelatkinson @abr-egn? Any ideas on how we could leverage `TypedBuilder` here without making a new datetime type?

sanav33: I was thinking about this as well. Having to call a `to_bson()` function after calling `build()` didn't seem like the best idea. I was able to recover the manual implementation, so let me know how that looks!

abr-egn: It looks like this is simply something `TypedBuilder` doesn't support yet; I'm fine with using the manual implementation, I don't think it'll matter to users one way or the other as long as how it works is consistent.

src/datetime.rs
```diff
@@ -1,12 +1,147 @@
...
     fmt::{self, Display},
     result,
     time::{Duration, SystemTime},
 };
 
-use time::format_description::well_known::Rfc3339;
+use builder::*;
+use time::{format_description::well_known::Rfc3339, Date};
+
+mod builder {
```
patrickfreed, line 12: we generally try (with the exception of tests sometimes) to have a single module per file. So with that, could we split this out to a `datetime/builder.rs` file?

src/datetime/builder.rs
```diff
@@ -0,0 +1,159 @@
...
+            hour,
+            minute,
+            second,
+            millisecond,
+        }
+    }
+}
+
+impl<Y, D> DateTimeBuilder<Y, NoMonth, D> {
+    /// Sets the month for the builder instance.
```
patrickfreed, line 65: These docstrings are good, but they are a little straightforward. I think it would be good to include some extra information that user might not necessarily be able to infer themselves, e.g. supported values (helps the user know if january is 1 or 0) or what happens when an invalid value is provided (the final `build` will return an error).

Ditto for others. For instance, in the `day` builder, it could be good to clarify what are accepted values to help answer questions like: what if we supply a day that's between 1 and 31 but isn't valid for that month year combo? (e.g. feb 29th on a non-leap year or the 31st of april.

Also, it's probably worth mentioning in the docstrings which ones are required and which ones are optional, and if one is optional what the default value will be.

patrickfreed (commented):
Overall looks good! I just have some suggestions for expanding the documentation a little bit. Also looks like the lint task is failing, so might need to run format or clippy.

patrickfreed:
Ah okay, that might just be from the recent Rust release. No need to fix that there if it's unrelated code.

patrickfreed (approved):
LGTM! great job with this

abr-egn (approved):
LGTM!

isabelatkinson (approved):
lgtm!

PR #362, opened 2022-06-22
RUST-1305 Add support for version 1.x of the `uuid` crate to the `bson` crate

abr-egn:
The approach taken here is to alias the imports rather than duplicating the various code blocks; it turns out that there wasn't any breakage in the API we use for uuid between 0.8 and 1.x.


Cargo.toml
```diff
@@ -57,7 +59,8 @@ indexmap = "1.6.2"
 hex = "0.4.2"
 base64 = "0.13.0"
 lazy_static = "1.4.0"
-uuid = { version = "0.8.1", features = ["serde", "v4"] }
+uuid-0_8 = { package = "uuid", version = "0.8.1", features = ["serde", "v4"], optional = true }
+uuid_1_lib = { package = "uuid", version = "1.1.2", features = ["serde", "v4"] }
```
abr-egn, line 63: Features and deps are in the same namespace, even for non-optional deps, so this couldn't be called just `uuid_1`.

src/serde_helpers.rs
```diff
@@ -4,6 +4,11 @@ use std::{convert::TryFrom, result::Result};
 
 use serde::{ser, Serialize, Serializer};
 
+#[cfg(feature = "uuid-0_8")]
```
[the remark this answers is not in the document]
abr-egn: I thought we'd decided in discussion that it was okay for these to be mutually exclusive; that's certainly what was recorded in RUST-1305.  Is that not accurate?

I don't see how the dependency union is an issue here; the user of this library will have a single version of `uuid` that they're working with and they can select that as the one the `bson` interface accepts; if other dependencies happen to use a different version internally that won't matter.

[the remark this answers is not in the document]
patrickfreed: I don't recall that but it's possible there was a miscommunication. The reason it's an issue is that Cargo treats `0.x` and `1.x` as essentially different crates, so libraries can depend on either or both versions of `uuid`. This means you could have a situation where dependency A (either direct or transitive) depends on `bson` with `uuid-1` flag enabled and dependency B (again direct or transitive) depends on `bson` with `uuid-0_8` flag on, and this would fail to compile due to the current mutually exclusive implementation. Since this covers the entire dependency graph and is basically out of an individual user's control, I think it's important that both features are purely additive in nature. The rust book has some [good info](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) on this problem.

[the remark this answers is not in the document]
abr-egn: Blarg, you're right, I hadn't considered the case where `bson` was the crate pulled in multiple times.  Updated - reduced some of the repetition with macros but there's still a _lot_ of boilerplate now.  We should consider dropping uuid 0.8 support for bson 3.0.

.evergreen/feature-sets.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+BASE_FEATURES="chrono-0_4,time-0_3,serde_with"
+FEATURE_SETS=("${BASE_FEATURES},uuid-0_8" "${BASE_FEATURES},uuid-1")
```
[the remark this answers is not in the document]
abr-egn: Since they're no longer exclusive, put this back to just `--all-features`.

Cargo.toml
```diff
@@ -35,7 +35,9 @@ default = []
 # if enabled, include API for interfacing with chrono 0.4
 chrono-0_4 = ["chrono"]
 # if enabled, include API for interfacing with uuid 0.8
-uuid-0_8 = []
```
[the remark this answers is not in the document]
abr-egn: Added a comment.  And yup, Cargo errors:
```
error: failed to parse manifest at `/Users/abraham.egnor/Source/bson-rust/Cargo.toml`

Caused by:
  optional dependency `uuid-0_8` is not included in any feature
  Make sure that `dep:uuid-0_8` is included in one of features in the [features] table.
```

We could avoid that by using the `dep:uuid-0_8` syntax, but that's not available in our MSRV.

src/serde_helpers.rs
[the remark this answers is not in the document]
patrickfreed: Yep, definitely. I filed RUST-1380 to track that.

Cargo.toml
```diff
@@ -35,7 +35,11 @@ default = []
...
+# This is commented out because Cargo implicitly adds this feature since
+# uuid-0_8 is also an optional dependency.
+# uuid-0_8 = []
+# if enabled, include API for interfacing with uuid 1.x
+uuid-1 = []
```
patrickfreed, line 42: Since uuid is stable and the flags aren't mutually exclusive, we could just include this support out of the box without a feature flag. We may have had a discussion about this already though that I can't remember the conclusions of. Given that we export a `bson::Uuid` type maybe its not necessary.

src/serde_helpers.rs
```diff
@@ -388,189 +412,273 @@ pub mod hex_string_as_object_id {
...
-    #[cfg_attr(docsrs, doc(cfg(feature = "uuid-0_8")))]
-    pub fn deserialize<'de, D>(deserializer: D) -> Result<Uuid, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let bson_uuid = crate::uuid::Uuid::deserialize(deserializer)?;
-        Ok(bson_uuid.into())
-    }
+#[allow(unused_macros)]
+macro_rules! as_legacy_binary_mod {
```
patrickfreed, line 487: these macros are very nice, great idea

src/uuid/mod.rs
```diff
@@ -436,40 +474,46 @@ impl Binary {
...
-impl<'de> serde_with::DeserializeAs<'de, uuid::Uuid> for crate::Uuid {
-    fn deserialize_as<D>(deserializer: D) -> std::result::Result<uuid::Uuid, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let uuid = Uuid::deserialize(deserializer)?;
-        Ok(uuid.into())
-    }
-}
+        #[$feat]
```
patrickfreed, line 490: these ones need to also only be implemented when `"serde_with"` is enabled

patrickfreed (commented):
looks good! one question and one small comment

src/uuid/mod.rs
abr-egn: Good catch, fixed.

Cargo.toml
abr-egn: It's not precisely equivalent, but the stumbling blocks we've run into in the driver with things that would have been better as optional features from the start has me inclined to keep this as optional.

patrickfreed (approved):
LGTM!

Cargo.toml
kmahar, line 42: we list our other features in the README, should we add this there too?

src/uuid/mod.rs
```diff
@@ -34,41 +35,46 @@
...
 //!
-//! // this also requires the "uuid-0_8" feature flag.
-//! let uuid = bson::Uuid::new().to_uuid_0_8();
+//! // this also requires the "uuid-1" feature flag.
+//! let uuid = bson::Uuid::new().to_uuid_1();
 //! # };
 //! ```
 //!
+//! For backwards compatibility, a `uuid-0_8` feature flag can be enabled, which provides the same
+//! API for interoperation with version 0.8 of the `uuid` crate.  Note that at most one of `uuid-1`
```
kmahar, line 75: in a comment above @patrickfreed mentioned that the flags are not mutually exclusive, but this seems to imply they are -- which is accurate?

kmahar (commented):
couple documentation related comments, LGTM otherwise!

src/uuid/mod.rs
abr-egn: Whoops, this is left over from when they were exclusive.  Fixed.

kmahar (approved):

PR #364, opened 2022-06-29
RUST-1374 Pin all dependencies when checking MSRV compilation

abr-egn:


kmahar (approved):
LGTM; since this is happening more and more it seems logical to just pin everything rather than having to make one-off PRs every time a dependency bumps its MSRV.

patrickfreed (approved):

PR #365, opened 2022-07-05
RUST-1225 Add base64 string to Binary constructor

sanav33:
Added a Binary constructor that accepts a base64 string instead of the bytes themselves.


src/bson.rs
```diff
@@ -1102,6 +1102,18 @@ impl Display for Binary {
 }
 
 impl Binary {
+    /// Creates a [`Binary`] from a base64 string and optional [`BinarySubtype`]. If the
+    /// `subtype` argument is `None`, the [`Binary`] constructed will default to
+    /// [`BinarySubtype::Generic`].
+    pub fn from_base64(input: &str, subtype: Option<BinarySubtype>) -> Option<Self> {
```
isabelatkinson, line 1108: nit: can we make the argument for `subtype` `impl<Into<Option<BinarySubtype>>`?

src/bson.rs
```diff
@@ -1102,6 +1103,21 @@ impl Display for Binary {
 }
 
 impl Binary {
+    /// Creates a [`Binary`] from a base64 string and optional [`BinarySubtype`]. If the
+    /// `subtype` argument is `None`, the [`Binary`] constructed will default to
+    /// [`BinarySubtype::Generic`].
+    pub fn from_base64(
+        input: &str,
+        subtype: impl Into<Option<BinarySubtype>>,
+    ) -> Result<Self, DecodeError> {
```
isabelatkinson, line 1112: We tend to avoid including types from other crates in our public API as they don't always have the same stability guarantees that we do. Under semantic versioning, any crate with a version that begins with 0 (e.g. 0.13.0, which is the version of the `base64` crate) can make breaking changes at any time -- as an example, `base64` could rename `DecodeError` to `DecodingError`, which would cause compilation failures for any code that relies upon the name being `DecodeError`. Because of this we'll want to avoid returning this type directly.

I looked over the various `Error` types we have defined in the BSON library and I'm not sure any of them are quite what we need for this situation. The closest would be the `SerializationError` under `ser::Error`, but that's more so used in the context of serializing to `Bson`. Alternatively, we could introduce a new `Error` type for this case.

Thoughts on this @patrickfreed @abr-egn @kmahar?

abr-egn: Agreed with Patrick on both the immediate thing to do and what we'll want to queue up for 3.0.

isabelatkinson: That sounds good to me. @sanav33 we can walk through what the new error type should look like, I think we'll want to add a `binary` module for it to live in to be consistent with our other error types. Also filed RUST-1406 to clean this up when we do a 3.0 release.

isabelatkinson (commented):
just a few small comments! looks like there are also some lint failures

isabelatkinson (commented):
looks like there are still some lint errors

sanav33:
> looks like there are still some lint errors

It looks like those failures are from an unrelated PR (https://github.com/mongodb/bson-rust/pull/258). Should I go ahead and fix them anyway? @isabelatkinson

isabelatkinson:
hmm you probably need to rebase this branch on main, it looks like #366 went in after this PR was created

sanav33:
I might be wrong but I think I rebased before the last push to this PR and it looks like all the changes from #366 are reflected in this PR.

isabelatkinson:
ohh I see what the issue is, `chrono` must've just released a new version that deprecated some constants. I needed to run `cargo update` to see the lint errors locally. don't worry about those in this PR, I'll make a separate one to fix.

isabelatkinson (approved):

abr-egn (approved):
LGTM!

src/binary.rs
```diff
@@ -0,0 +1,98 @@
...
+            base64::encode(&self.bytes)
+        )
+    }
+}
+
+impl Binary {
+    /// Creates a [`Binary`] from a base64 string and optional [`BinarySubtype`]. If the
+    /// `subtype` argument is `None`, the [`Binary`] constructed will default to
+    /// [`BinarySubtype::Generic`].
+    pub fn from_base64(input: &str, subtype: impl Into<Option<BinarySubtype>>) -> Result<Self> {
```
patrickfreed, line 33: this `&str` can be an `impl AsRef<str>`

src/binary.rs
```diff
@@ -0,0 +1,98 @@
...
+            u8::from(self.subtype),
+            base64::encode(&self.bytes)
+        )
+    }
+}
+
+impl Binary {
+    /// Creates a [`Binary`] from a base64 string and optional [`BinarySubtype`]. If the
+    /// `subtype` argument is `None`, the [`Binary`] constructed will default to
+    /// [`BinarySubtype::Generic`].
```
patrickfreed, line 32: I think it would be good to have a doc example here. I think the test we added for this would actually be pretty good.

patrickfreed (commented):
A few minor comments, sorry for not getting them in before Sana left. Do you want to take this over @isabelatkinson?

isabelatkinson:
there's an unrelated clippy failure -- going to fix in a separate PR

patrickfreed (approved):

PR #370, opened 2022-08-12
RUST-1385 Add binary/raw conversions

abr-egn:
RUST-1385

This makes public the existing `Binary::as_raw_binary`, and adds the reciprocal `RawBinaryRef::to_binary`.  Tagging RUST-1385 because I noticed these didn't exist during implementation of that :)

### DIFF
--- a/src/raw/bson_ref.rs
+++ b/src/raw/bson_ref.rs
@@ -460,6 +460,14 @@ pub struct RawBinaryRef<'a> {
 }
 
 impl<'a> RawBinaryRef<'a> {
+    /// Copy the contents into a `Binary`.
+    pub fn to_binary(&self) -> Binary {
+        Binary {
+            subtype: self.subtype,
+            bytes: self.bytes.to_owned(),
+        }
+    }
+
     pub(crate) fn len(&self) -> i32 {
         match self.subtype {
             BinarySubtype::BinaryOld => self.bytes.len() as i32 + 4,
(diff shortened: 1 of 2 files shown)


kmahar (approved):

isabelatkinson (approved):

src/raw/bson_ref.rs
```diff
@@ -460,6 +460,14 @@ pub struct RawBinaryRef<'a> {
 }
 
 impl<'a> RawBinaryRef<'a> {
+    /// Copy the contents into a `Binary`.
+    pub fn to_binary(&self) -> Binary {
+        Binary {
+            subtype: self.subtype,
+            bytes: self.bytes.to_owned(),
```
patrickfreed, line 467: WDYT about also implementing the [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) / [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) relationship for these two types similar to what we do for `RawDocument`/`RawDocumentBuf`?

abr-egn: AFAICT that won't work because `Borrow` requires that the signature be `fn borrow(&self) -> &Borrowed`, and because `RawBsonRef` is a wrapper rather than an unsized type that won't work.

patrickfreed (approved):

PR #378, opened 2022-10-11
#376 Implemented TryFrom<&str> for ObjectId

### LINKED ISSUE #376: Implement conversion traits between strings and ObjectId
I'd like to be able to do things like

```rust
fn query<K: TryInto<ObjectId>>(key: K) {}

fn main() {
    let key = "000000000000000000000000";
    query(key);

    let key = String::from("000000000000000000000000");
    query(key);

    let key = ObjectId::new();
    query(key);
}
```

but it doesn't seem like the necessary conversion traits have been implemented. I'm not a Rust expert and so I cannot speak to what the solution is, but it would be really convenient to be able to both pass ObjectIds and types that can be turned into ObjectIds.

clarkmcc:
Fixes #376 and allows for things like this
```rust
fn handle<T>(key: T) where T: TryInto<ObjectId> {
    let id = key.try_into();
}

handle("000000000000000000000000");
handle(ObjectId::new());
```

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -1,7 +1,7 @@
 //! ObjectId
 
 use std::{
-    convert::TryInto,
+    convert::{TryFrom, TryInto},
     error,
     fmt,
     result,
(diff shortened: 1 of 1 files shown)


abr-egn:
It looks like there's an unused import in the test code - fix that and this'll be good to merge!

> What I'd like to be able to do is create functions that accept both strings or `ObjectId`s and do the necessary conversions under-the-hood. Is there a better way to accomplish this?

Using `TryFrom` is definitely the right way to do that specific thing - what I typically see is a more distinct separation of input parsing from other logic.

abr-egn:
It looks like it's still failing the lint check - please make sure to use rustfmt to format the files.

clarkmcc:
Sorry for the delay. Per your advice, I've found probably a more idiomatic way to handle this problem. If you'd like to close this without merging, that is no problem with me.

abr-egn:
No worries!  Glad you found a way forward.

PR #384, opened 2022-11-14
RUST-1062 More efficiently serialize array indexes

patrickfreed:
RUST-1062

This PR updates the array serialization logic in the raw document serializer to avoid unnecessary heap allocations, resulting in an ~8% improvement in performance in a simple benchmark I'm using.

The solution used here is a bit different than the originally proposed one to just push bytes directly onto the buffer, but it was necessary in order to preserve the expected byte order (pushing them would result in a backwards string). As a result, the performance improvements were not as significant as originally envisioned unfortunately.

### DIFF
--- a/src/ser/raw/document_serializer.rs
+++ b/src/ser/raw/document_serializer.rs
@@ -30,17 +30,29 @@ impl<'a> DocumentSerializer<'a> {
         })
     }
 
+    /// Serialize a document key using the provided closure.
+    fn serialize_doc_key_custom<F: FnOnce(&mut Serializer) -> Result<()>>(
+        &mut self,
+        f: F,
+    ) -> Result<()> {
+        // push a dummy element type for now, will update this once we serialize the value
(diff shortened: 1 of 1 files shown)


abr-egn (approved):
LGTM!

src/ser/raw/document_serializer.rs
```diff
@@ -225,7 +226,10 @@ impl<'a> serde::Serializer for KeySerializer<'a> {
 
     #[inline]
     fn serialize_u64(self, v: u64) -> Result<Self::Ok> {
-        Err(Self::invalid_key(v))
+        use std::io::Write;
```
kmahar, line 229: should we be concerned at all about users accidentally encountering this and being surprised by the behavior, e.g. if they are working with a HashMap<i64, String> and try to serialize that to BSON?

kmahar (approved):

isabelatkinson (approved):

PR #386, opened 2022-11-23
Double canonical extended json fraction fails round trip on Deserialize from_slice

mitchrmiller:
I believe this was a bug as try_from has this functionality--while from_slice had functionality that failed on fractions.  Either this is the proper decode fix, or the encoding needs fixed where .fract() adds ".0" to string.


abr-egn:
Thanks for catching this and sending a fix!  I've merged it and it'll go out with the next release.

PR #388, opened 2022-12-12
Fix links in the Feature Flags table.

judy2k:
The links in the feature flags table in the README weren't absolute URLs, so they went to a GitHub 404 page.


isabelatkinson:
Thanks for this fix!

PR #389, opened 2022-12-16
RUST-782 Add hyperlinks for all types in rustdoc

abr-egn:


src/uuid/mod.rs
```diff
@@ -13,8 +13,8 @@
 //! binary values. Instead, when serialized with `bson::to_bson`, it produces as a string, and when
 //! serialized with `bson::to_vec`, it produces a binary value with subtype _0_ rather than 4.
 //! Because of this, it is highly recommended to use the [`crate::Uuid`] type when working with BSON
-//! instead of the `uuid` crate's `Uuid`, since [`crate::Uuid`] correctly produces subtype 4 binary
-//! values via either serialization function.
+//! instead of the `uuid` crate's [`Uuid`], since [`crate::Uuid`] correctly produces subtype 4
+//! binary values via either serialization function.
```
kmahar, line 17: here and elsewhere in these docs is it worth linking to the corresponding `BinarySubtype` cases rather than just mentioning subtypes numbers, e.g. `BinarySubtype::Generic` and `BinarySubtype::Uuid`?

kmahar (commented):
one suggestion but LGTM otherwise, will tag in the others

src/uuid/mod.rs
abr-egn: I agree that's a reasonable idea, it looks like there's a bunch of places that could use that update so I filed RUST-1561 for that.

kmahar (approved):

isabelatkinson (approved):

Issue #381, opened 2022-10-31
RUST-1540 Add Ability to deserialize String to ObjectId and i64 to DateTime 


borngraced:
For some reasons I see `Bson` doesn't deserialize `Objected` and `DateTime` to `String` and `i64` respectively, so I think it'd be very helpful if there's an inbuilt helper fn to use with serde's `deserialize_with` macro attribute to automatically deserialize `ObjectId` to `String` and `DateTime` to `i64`..

# Example
```rust
fn des_objectid<'de, D>(deserializer: D) -> Result<String, D::Error>
where
    D: de::Deserializer<'de>,
{
    let s: ObjectId = de::Deserialize::deserialize(deserializer)?;

    Ok(s.to_string())
}
```

then we'll be able to do something like this 
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct User {
    #[serde(deserialize_with = "des_objectid")]
    pub _id: String,
    }
 ```

This is very useful for returning `ObjectId` as String and DateTime as i64 

```json
{ "id": "3895893bf983nc9..." }
```

instead of something like this
```json
{ "id": { "$oid" : "3895893bf983nc9..."} }
```

isabelatkinson:
Hi @borngraced, thanks for filling this issue! We have a [`serde_helpers`](https://docs.rs/bson/latest/bson/serde_helpers/index.html) module that contains functions you can use with the `serialize_with` and `deserialize_with` attributes. Included is a [`hex_string_as_object_id`](https://docs.rs/bson/latest/bson/serde_helpers/hex_string_as_object_id/index.html) that allows conversion between `ObjectId`s and strings. We don't, however, have helpers for converting between `DateTime`s and `i64`s, so I've filed RUST-1540 to add them. If you're interesting in contributing to the BSON library, feel free to make a pull request implementing this functionality.

PR #393, opened 2023-01-09
RUST-1540 Rename `i64_as_datetime` to `i64_as_bson_datetime`

patrickfreed:
This is a follow on PR to #381 which renames the new `i64_as_datetime` helper to `i64_as_bson_datetime`, which more closely matches our existing helpers. The "bson" portion helps differentiate between different datetime types that we have support for, such as those from `chrono`.


abr-egn (approved):
I don't think this needs full review :)

Issue #392, opened 2023-01-04
Improve Docs


elibroftw:
The docs don't indicate how the serde serialization will be performed on ObjectID or Datetime. I suggest improving the documentation for objectID and Datetime to show how access the fields in a frontend that expects json.

I also ask to improve the datetime documentation to include how to show the date in different formats.

patrickfreed:
Thanks for filing this issue, I've gone ahead and updated the docs for `DateTime` and `ObjectId` to be clearer about how they interact with serde. Let us know if you have any further feedback.

PR #401, opened 2023-01-23
minor: improve serde documentation of `DateTime` and `ObjectId`

patrickfreed:
Addresses #392 

To see what the new docs look like, you can run `cargo doc --open`. Some of the `chrono` hyperlinks are broken when generating the docs that way, but I think that's not the case when docs.rs actually generates them.


abr-egn (approved):
LGTM!

isabelatkinson (approved):

kmahar (approved):

PR #402, opened 2023-01-24
feat: add ability to generate ObjectId from SystemTime

Denys-Bushulyak:
Hi, dear folks!

This PR adds the ability to generate an ObjectId from SystemTime. I found it useful because sometimes there is a need to create records in MongoDB from the "past" and use the nice ability of MongoDB to sort by ObjectId.

In the official documentation [page](https://www.mongodb.com/docs/manual/reference/method/ObjectId/#mongodb-method-ObjectId), there is a description of the ability to create an ObjectId from a hex and unix epoch timestamp, but I didn't find the ability to create one from a timestamp in the Rust Crate. However, since we are already there, I have decided to implement it using SystemTime as a first step. If you find it useful, I will be glad, because I think it would be better to extend the official crate rather than creating a separate one for this ability.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -98,6 +98,21 @@ impl From<[u8; 12]> for ObjectId {
     }
 }
 
+impl From<SystemTime> for ObjectId {
+    fn from(datetime: SystemTime) -> Self {
+        let mut object_id = ObjectId::new();
+        let timestamp: u32 = datetime
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("system clock is before 1970")
+            .as_secs()
+            .try_into()
+            .unwrap(); // will succeed until 2106 since timestamp is unsigned
+        object_id.id[..TIMESTAMP_SIZE].copy_from_slice(&timestamp.to_be_bytes()[..TIMESTAMP_SIZE]);
+
+        object_id
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Denys-Bushulyak:
I have tried to follow the contribution guide and all tests and builds have passed on my local machine. It's strange to see that all checks have failed.

kmahar:
Hi @Denys-Bushulyak, GitHub was displaying that because patches from external contributors have to be manually authorized by a team member, and for whatever reason it considers the check to have "failed" when really the check just hasn't run yet. I just authorized the patch.

Denys-Bushulyak:
If there's anything I can assist with, please don't hesitate to reach out and I'll be happy to help or to discuss.

isabelatkinson:
Hi @Denys-Bushulyak, thanks for creating this PR and apologies for the delayed response. After discussion amongst the team we've decided not to expose this functionality, as we consider the usage of system time for the creation of object IDs to be an implementation detail of the type rather than a feature. If you'd like to create an `ObjectId` from bytes (including those generated from the system time), I recommend using [`ObjectId::from_bytes`](https://docs.rs/bson/latest/bson/oid/struct.ObjectId.html#method.from_bytes). Please feel free to open an issue or create another pull request if you have additional questions or requests!

PR #403, opened 2023-02-05
add decimal128 to string function

Wenyi1995:


abr-egn:
Hi!

Thank you for the contribution!  Coincidentally, we're already working on support for reading and writing decimal128 values as human-readable strings, so we have an implementation of this logic already - this will go out with the next release.

PR #407, opened 2023-03-13
RUST-1613 Update feature flags table

isabelatkinson:
This table was out of date/had broken links -- would like to get the fix in prior to releasing 2.6.


abr-egn (approved):
LGTM!

PR #414, opened 2023-05-09
update ahash from 0.7.6 to 0.8.3

Liyixin95:


abr-egn:
Thanks for contributing!  I've kicked off the evergreen run and assuming that passes I'll merge this in.

abr-egn:
Sorry for the delay!

It looks like (for reasons I don't currently understand) versions 0.8.1 and above of `ahash` don't work with our MSRV, so I've updated your PR only bump the dependency to 0.8.0, and added a lockfile needed for our MSRV testing.  Does that work for you?

Liyixin95:
> Does that work for you?

of course.

> It looks like (for reasons I don't currently understand) versions 0.8.1 and above of ahash don't work with our MSRV.

the msrv of `ahash` version 0.8.3 is `1.6.0`, which they do not indicate in their project. 

The specific reason is [namespaced-features](https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html)

In their project looks like:

`atomic-polyfill = [ "dep:atomic-polyfill", "once_cell/atomic-polyfill"]`

abr-egn (approved):

PR #417, opened 2023-06-08
RUST-1677 Add duration_since methods to `DateTime`

isabelatkinson:
Adds `checked_duration_since` and `saturating_duration_since` methods to `crate::DateTime`. These are based off of the methods of the same name in `std::time::Instant`. I chose not to add any `Sub` implementation because we'd need to decide what to do when the result is negative (i.e. panic or return 0); I'd rather let users make this decision explicitly.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -389,6 +389,19 @@ impl crate::DateTime {
         })?;
         Ok(Self::from_time_0_3(odt))
     }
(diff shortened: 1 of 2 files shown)


isabelatkinson:
I can't repro the MSRV failure locally, but it seems to be unrelated to these changes

PR #431, opened 2023-09-07
RUST-1753: Add a utility HumanReadable wrapper

abr-egn:
RUST-1753

I propose that instead of adding an explicit option to `Collection` to force on human-readable [de]serialization, we provide a type wrapper that does the same thing.  This should provide equivalent functionality (instead of setting `opts.human_readable_serialization = true` for a given `Collection<Foo>`, specify the type as `Collection<HumanReadable<Foo>>`), but means that we don't have to support checking the explicit option in every relevant code path for every method.  Subjectively, I also find this cleaner :)

I put this in the `bson::serde_helpers` module because it's arguably independent of the driver; it's even arguably independent of bson, but I don't think we want to maintain another crate.

If we do decide to use this, the path forward would be:
1. Put out a new release of the `bson` crate with these helpers.
2. Update the driver to remove the option and consistently use binary [de]serialization; AFAICT this is mostly done already as part of the earlier work, it just needs the fix in https://github.com/mongodb/mongo-rust-driver/pull/948.
3. Put out a `beta.1` release of the driver with those changes.

### DIFF
--- a/src/serde_helpers.rs
+++ b/src/serde_helpers.rs
@@ -2,7 +2,7 @@
 
 use std::{convert::TryFrom, result::Result};
 
-use serde::{ser, Serialize, Serializer};
+use serde::{ser, Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::oid::ObjectId;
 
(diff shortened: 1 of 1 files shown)


isabelatkinson (approved):
lgtm! i think we should add some docs/examples of how to use this in the driver but that can be part of step 2 :)

abr-egn:
Unfortunately, writing tests for this shows that it doesn't work :(  When serializing nested fields, a new serializer is constructed, and that doesn't have (and has no way of knowing about) the wrapper of the top-level serializer.

Issue #437, opened 2023-10-31
use simd to optimize uft8 validation.


Liyixin95:
# benchmark env

## benchmark suit
[benchmark suit](https://github.com/mongodb/mongo-rust-driver/tree/main/benchmarks) 

id: 10, 12, 14

## my computer

windows11 
13th Gen Intel(R) Core(TM) i7-13700H
32.0 GB 3200 MHz

# benchmark result

## original
```shell
Running tests nonverbosely...

Running Flat BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Flat BSON Decoding -- Score: 367.726 MB/s, Median Iteration Time: 0.205s

Running Deep BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Deep BSON Decoding -- Score: 67.747 MB/s, Median Iteration Time: 0.290s

Running Full BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Full BSON Decoding -- Score: 274.048 MB/s, Median Iteration Time: 0.209s

BSONBench Score = 236.507 MB/s
```

## bstr
```shell
Running tests nonverbosely...

Running Flat BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Flat BSON Decoding -- Score: 394.848 MB/s, Median Iteration Time: 0.191s

Running Deep BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Deep BSON Decoding -- Score: 73.264 MB/s, Median Iteration Time: 0.268s

Running Full BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Full BSON Decoding -- Score: 277.837 MB/s, Median Iteration Time: 0.206s

BSONBench Score = 248.650 MB/s
```

**[bstr](https://github.com/BurntSushi/bstr) need to be pinned to `<1.7.0` to satisfy the msrv requirment.**

## simdutf8
```shell
Running tests nonverbosely...

Running Flat BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Flat BSON Decoding -- Score: 404.186 MB/s, Median Iteration Time: 0.186s

Running Deep BSON Decoding...
  [00:01:01] [########################################] 100/100 (0s)
TEST: Deep BSON Decoding -- Score: 74.206 MB/s, Median Iteration Time: 0.265s

Running Full BSON Decoding...
[...]

PR #440, opened 2023-11-08
RUST-1798 use simd to optimize utf8 validation

Liyixin95:
#437

### DIFF
--- a/src/de/error.rs
+++ b/src/de/error.rs
@@ -1,6 +1,7 @@
-use std::{error, fmt, fmt::Display, io, string, sync::Arc};
+use std::{error, fmt, fmt::Display, io, sync::Arc};
 
 use serde::de::{self, Unexpected};
+use simdutf8::basic::Utf8Error;
 
 use crate::Bson;
 
@@ -13,7 +14,7 @@ pub enum Error {
 
     /// A [`std::string::FromUtf8Error`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html) encountered
     /// while decoding a UTF-8 String from the input data.
-    InvalidUtf8String(string::FromUtf8Error),
+    InvalidUtf8String(Utf8Error),
 
     /// While decoding a [`Document`](crate::Document) from bytes, an unexpected or unsupported
     /// element type was encountered.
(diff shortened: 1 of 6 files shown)


src/de/error.rs
```diff
@@ -13,7 +14,7 @@ pub enum Error {
 
     /// A [`std::string::FromUtf8Error`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html) encountered
     /// while decoding a UTF-8 String from the input data.
-    InvalidUtf8String(string::FromUtf8Error),
+    InvalidUtf8String(Utf8Error),
```
isabelatkinson, line 17: Changing the type contained in this error variant is a breaking change and would require a major version release, which we are not planning to do in the near term. Can a `simdutf8::simple::Utf8Error` be mapped to a `string::FromUtf8Error` to avoid this?

src/de/mod.rs
```diff
@@ -152,7 +149,13 @@ fn read_cstring<R: Read + ?Sized>(reader: &mut R) -> Result<String> {
...
     }
 
-    Ok(String::from_utf8(v)?)
+    to_string(v)
+}
+
+fn to_string(v: Vec<u8>) -> Result<String> {
+    let _ = simdutf8::basic::from_utf8(&v)?;
+    // Safety: `v` is a valid UTF-8 string.
+    unsafe { Ok(String::from_utf8_unchecked(v)) }
```
isabelatkinson, line 158: Any reason the result of the first line of this function cannot be converted to a `String` and returned directly?

```
fn to_string(v: Vec<u8>) -> Result<String> {
    let str = simdutf8::basic::from_utf8(&v)?;
    Ok(str.to_string())
}
```
I'd like to avoid unsafe code if we do not need it here.

isabelatkinson (changes requested):
A few suggestions, it looks like there are also some formatting errors. Can you please run the following on the files you've changed?
```
rustfmt +nightly --unstable-features
```

src/de/mod.rs
Liyixin95: `str.to_string() ` will make a copy of original bytes. It may be even slower then the std's `String::from_utf8`.

Also, you can see the issue [here](https://github.com/rusticstuff/simdutf8/issues/73).

src/de/error.rs
Liyixin95: The std's `String::FromUtf8Error` does not expose any constructor. So may be this can not be avoided ...

isabelatkinson: Got it. Unfortunately we'll need to hold off on this until we can change our error types. I've added the 3.0 label to RUST-1798 to pick up this work again when we do our next major version release.

PR #448, opened 2023-12-22
minor: bump indexmap to `2.1.0`

froydnj:
This change aligns the required `indexmap` version with that of `serde_json` so there's less crate duplication in the dependency graph.


froydnj:
(It would be nice if this was landed on the 2.8.x branch too.)

isabelatkinson:
Hey @froydnj, thanks for making this PR! I ran CI and the minimum supported Rust version (MSRV) compilation task failed due to `indexmap` 2.1.0 requiring Rust 1.64, which is higher than our current MSRV (1.61). Bumping our MSRV to 1.64 should be fine for the sake of these changes -- would you mind pushing a commit here doing so? The easiest way to make this change is to grep `bson-rust` for `1.61` and update to `1.64` wherever necessary.

> (It would be nice if this was landed on the 2.8.x branch too.)

Per our [MSRV policy](https://github.com/mongodb/bson-rust/#minimum-supported-rust-version-msrv) we don't update the MSRV in patch versions, so we will need to wait to release this until the next minor version of this crate is published.

froydnj:
@isabelatkinson Thanks for the CI run and the additional context!  I pushed a MSRV bump.

isabelatkinson:
Thanks for the changes @froydnj! I just pushed an update to our CI config and a fix for a clippy failure that came with the MSRV bump. I'm going to tag in another team member for review and then this should be ready to merge.

abr-egn (approved):
LGTM, thank you!

PR #449, opened 2024-01-01
provide lazy iterator implementation

tychoish:
Introduces a `RawElement` type which will can be used to lazily resolve bson values during iteration. 

This does not change the functionality of any existing interfaces for compatibility concerns, (though the `iter()` functions do return different types) and the `IntoIter` implementations changed, but semantically it's all the same. 

Happy to make changes if ya'll are interested. in this :)


abr-egn:
Hi!  This is definitely interesting, I'll be looking over the code in detail in the next few days.  Just a clarification, when you said
> This does change the functionality of any existing interfaces for compatibility concerns

that was a typo for
> This does _not_ change the functionality of any existing interfaces for compatibility concerns

correct?  We're strict about maintaining backwards compatibility.

tychoish:
Yes, sorry that's correct. 

The only publicly visible change is the return type of the `RawDocument`'s `iter()` method is now a wrapper around the new implementation (which has the previous type's name): but the returned iterator functions identically.

I'm not opposed to changing the name of the new (this) iterator and putting back a shim implementation with this name and the legacy functionality, though I don't know what things should be named (my inclination would be to call this `RawIter`, but I defer to you all, of course.

> We're strict about maintaining backwards compatibility.

Oh, I know ;)

src/raw/iter.rs
```diff
@@ -49,56 +52,224 @@ impl<'a> Iter<'a> {
...
+        self.value()?.to_raw_bson().try_into()
+    }
+}
+
+impl<'a> RawElement<'a> {
+    pub fn len(&self) -> usize {
+        self.size
+    }
+
+    pub fn key(&self) -> String {
```
abr-egn, line 124: Is there any particular reason to return the string as copy here rather than the ref?

src/raw/array.rs
```diff
@@ -275,7 +275,10 @@ impl<'a> Iterator for RawArrayIter<'a> {
 
     fn next(&mut self) -> Option<Result<RawBsonRef<'a>>> {
         match self.inner.next() {
-            Some(Ok((_, v))) => Some(Ok(v)),
+            Some(Ok(elem)) => match elem.value() {
```
abr-egn, line 278: Can this just be `Some(Ok(elem)) => Some(elem.value())` ?

src/raw/document.rs
```diff
@@ -577,10 +594,10 @@ impl TryFrom<&RawDocument> for crate::Document {
 }
 
 impl<'a> IntoIterator for &'a RawDocument {
-    type IntoIter = Iter<'a>;
+    type IntoIter = Box<dyn Iterator<Item = Self::Item> + 'a>;
```
abr-egn, line 597: Unfortunately, this is a breaking change.  I think this can be solved by what you suggested in the conversation - keep the existing functionality named `Iter` and call the new one `RawIter`.

abr-egn, line 597: Unfortunately, this is a breaking change.  I think this can be solved by what you suggested in the conversation - keep the existing functionality named `Iter` and call the new one `RawIter`.

abr-egn, line 597: Unfortunately, this is a breaking change.  I think this can be solved by what you suggested in the conversation - keep the existing functionality named `Iter` and call the new one `RawIter`.

abr-egn:
It looks like this fails the fuzz test; this program reproduces the failure:
```
use bson::RawDocument;

fn main() {
    let doc = RawDocument::from_bytes(&[25, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 5, 0, 254, 255, 255, 127, 0, 5, 0, 255, 255, 255, 9, 0]).unwrap();
    for _ in doc {
        println!("_");
    }
}
```

tychoish:
Nice! I'll look at this this weekend or early next week.

src/raw/array.rs
tychoish: yup!

src/raw/document.rs
tychoish: alas, easy enough done

tychoish (commented):
thanks for the review; I made the changes and fixed the fuzz test.

abr-egn:
This fails the fuzzer with different input (`[23, 0, 0, 0, 5, 0, 255, 255, 255, 255, 255, 255, 255, 3, 0, 0, 0, 247, 4, 247, 245, 0, 0]`) and the corpus test (`tests::spec::corpus::run`) also fails with what is probably a related overflow error.

tychoish:
> This fails the fuzzer with different input (`[23, 0, 0, 0, 5, 0, 255, 255, 255, 255, 255, 255, 255, 3, 0, 0, 0, 247, 4, 247, 245, 0, 0]`) and the corpus test (`tests::spec::corpus::run`) also fails with what is probably a related overflow error.

Ok, _this_ time I think I have.

abr-egn:
Still failing, I'm afraid (`[17, 0, 0, 0, 5, 5, 0, 255, 255, 255, 127, 255, 255, 255, 11, 13, 0]`).  You can run the fuzzer test yourself (assuming you've installed `cargo-fuzz` and a nightly toolchain) with `cargo +nightly fuzz run iterate -- -rss_limit_mb=4096 -max_total_time=60`; probably a good idea to do that before pushing.

tychoish:
I fixed it again, ran the fuzz test on my laptop for a while, nothing popped, and everything else passed (though my laptop is linux, it's not un-weird so I'm not sure what makes sense to validate.)

abr-egn (approved):
Looks good!  Tagging in Isabel for review.

isabelatkinson (approved):
lgtm, thank you for the contribution!

tychoish:
Thanks @isabelatkinson @abr-egn! Do you schedule merges for things like this? Is there a regular release cadence for these libraries?

Cheers,
sam

abr-egn:
We don't have a specific release schedule, but I anticipate releasing 2.9.0 with this included in a week or so.

PR #452, opened 2024-01-19
Implement From<&mut T> for Bson

tyilo:
To match the `From<&T>` implementation.


abr-egn:
Thanks for the contribution!  This will be included in the 2.9.0 release next week.

PR #453, opened 2024-01-21
Fix feature name for serde_with Uuid implementation

jonasbb:
The feature is named `serde_with-3`. The `_` makes it such that the `SerializeAs` implementation is not working. The `DeserializeAs` implementation above uses the correct feature name and shows up in the documentation.


abr-egn:
Thanks for the fix!  This will go out with the 2.9.0 release this week.

PR #461, opened 2024-02-26
fix: export RawElement type

tychoish:
Missed that exports had to be explicitly re-exported in #449.

### DIFF
--- a/src/raw/iter.rs
+++ b/src/raw/iter.rs
@@ -145,6 +145,10 @@ impl<'a> TryInto<Bson> for RawElement<'a> {
 }
 
 impl<'a> RawElement<'a> {
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
     pub fn len(&self) -> usize {
         self.size
     }
--- a/src/raw/iter.rs
+++ b/src/raw/iter.rs
@@ -145,11 +145,7 @@ impl<'a> TryInto<Bson> for RawElement<'a> {
 }
 
 impl<'a> RawElement<'a> {
-    pub fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
-
-    pub fn len(&self) -> usize {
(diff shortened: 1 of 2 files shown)


tychoish:
@abr-egn, I found this when trying to use this in addition to #462. I missed that the exports of symbols here needed to be explicit, so while you can use this, if you want to pass the iterator around explicitly, you can't because the symbol isn't exported.

abr-egn:
Yup :)  Isabel's taking a look at this.

src/raw/iter.rs
```diff
@@ -145,6 +145,10 @@ impl<'a> TryInto<Bson> for RawElement<'a> {
 }
 
 impl<'a> RawElement<'a> {
+    pub fn is_empty(&self) -> bool {
```
isabelatkinson, line 148: Thanks for catching this clippy error! I think it would make more sense to call the `len` method `size` since it's exposing the number of bytes in the element. That would remove the need to have this `is_empty` method, which likely won't have any utility for the users of this type.

isabelatkinson (changes requested):
Just one minor request, going to tag in @abr-egn as well. On an unrelated note, it doesn't seem like our documentation lint task is catching that there's missing documentation for this type and its methods. I filed RUST-1868 to fix this.

src/raw/iter.rs
isabelatkinson: Oh that's a great point, in that case I think we'll want to stick to `len` for the name and change it with the next major release (RUST-1869). I still don't think it makes that much sense to have an `is_empty` method, so I suggest silencing the clippy error with `#[allow(clippy::len_without_is_empty)]`. Sorry for all of the back-and-forth here!

tychoish: done!

isabelatkinson (approved):

abr-egn (approved):
LGTM!

PR #463, opened 2024-03-18
RUST-1512 Provide a method to append a borrowed value

abr-egn:
RUST-1512

This adds `RawDocumentBuf::append_ref`, providing a method equivalent to `append` for borrowed data.  This can be useful for avoiding the need for intermediate buffers when the source data is borrowed.

### DIFF
--- a/src/raw/array_buf.rs
+++ b/src/raw/array_buf.rs
@@ -96,10 +96,6 @@ impl RawArrayBuf {
         self.inner.append(self.len.to_string(), value);
         self.len += 1;
     }
-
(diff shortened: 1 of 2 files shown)


isabelatkinson (approved):
nice! could we combine `append` and `append_ref` somehow in a future major version release?

abr-egn:
> nice! could we combine `append` and `append_ref` somehow in a future major version release?

I'm not sure, actually - what we want is basically anything that can convert into either owned or borrowed bson, and I don't know if that can be expressed reasonable in the Rust type system.  Filed RUST-1887 to track that.

PR #465, opened 2024-03-25
RUST-1892 Fix timestamp [de]serialization on big-endian machines

abr-egn:
RUST-1892

The actual bugfix is the changes in src/bson.rs; while I was debugging, I also noticed that `Document::to_writer/from_reader` had Yet Another implementation of [de]serialization logic, so I ripped that out in favor of the existing stuff.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -1029,19 +1029,21 @@ impl Display for Timestamp {
 }
 
 impl Timestamp {
-    pub(crate) fn to_le_i64(self) -> i64 {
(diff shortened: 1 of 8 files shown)


abr-egn:
Sidebar in case it's ever needed again: I used the instructions [here](https://docs.gitlab.com/omnibus/development/s390x.html) to spin up a local docker container running on an emulated s390x:
```
docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
docker run --rm -it s390x/ubuntu bash
```

PR #471, opened 2024-04-04
RUST-1764 Add `HumanReadable` convenience wrapper

abr-egn:
RUST-1764

This wrapper removes the need for the `human_readable_serialization` option for `Collection` and provides more granular control as well.  There could hypothetically also be a use for a corresponding `NonHumanReadable` wrapper but I'd prefer to wait on adding that until someone asks for it.

### DIFF
--- a/src/de/raw.rs
+++ b/src/de/raw.rs
@@ -295,6 +295,7 @@ impl<'de> Deserializer<'de> {
                         let doc = Bson::JavaScriptCode(code).into_extended_document(false);
                         visitor.visit_map(MapDeserializer::new(
                             doc,
+                            #[allow(deprecated)]
                             DeserializerOptions::builder().human_readable(false).build(),
                         ))
                     }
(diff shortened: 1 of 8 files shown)


src/de/raw.rs
```diff
@@ -454,12 +458,16 @@ impl<'de, 'a> serde::de::Deserializer<'de> for &'a mut Deserializer<'de> {
 
                 self.deserialize_next(visitor, DeserializerHint::RawBson)
             }
+            HUMAN_READABLE_NEWTYPE => {
+                self.human_readable = true;
```
isabelatkinson, line 462: Could this be an issue when a `HumanReadable<T>` is nested in a struct with non-wrapped types following it? e.g.

```
struct Test {
    human_readable: HumanReadable<Document>,
    non_human_readable: Document,
}
```

abr-egn: Good catch!  I've fixed the logic and updated the test to flag that kind of failure.

isabelatkinson (approved):

isabelatkinson:
>IMO, we should also remove the option from [De]SerializerOptions but that'll have to wait for BSON 3.0. If that makes sense to you I'll add deprecation warnings on those and file a ticket for the removal.

SGTM

PR #473, opened 2024-05-14
minor: fix lifetime of RawElement::key

abr-egn:
Without the annotation it gets a lifetime bound to the `RawElement` struct, not the underlying data.


abr-egn:
Clippy failures are unrelated.

isabelatkinson (approved):

PR #480, opened 2024-07-02
RUST-1773 Merge duplicate extjson map parsing between OwnedOrBorrowedRawBsonVisitor and SeededVisitor

abr-egn:
RUST-1773

This puts responsibility for turning a Serde map into bson values squarely on `OwnedOrBorrowedRawBsonVisitor`; `SeededVisitor`'s job is now to efficiently buffer those values.

In my plan for the 🎆Glorious Future🎆, `OwnedOrBorrowedRawBsonVisitor` will also be taking over for the other places that parse extjson (`BsonVisitor`, `TryFrom<serde_json::Value>`, `crate::extjson::models`) and will necessarily be getting more liberal in the forms it accepts, so this means that doesn't need to be mirrored in `SeededVisitor` as well.

With this, I think the separation of logic is clear enough that there's no need to further merge; the reasoning from your original work on why that would involve quite a bit of complication is still very valid.

### DIFF
--- a/src/raw/serde/bson_visitor.rs
+++ b/src/raw/serde/bson_visitor.rs
@@ -44,6 +44,160 @@ use super::{
 /// A visitor used to deserialize types backed by raw BSON.
 pub(crate) struct OwnedOrBorrowedRawBsonVisitor;
 
+pub(super) enum MapParse<'de> {
+    Leaf(OwnedOrBorrowedRawBson<'de>),
+    Aggregate(CowStr<'de>),
+}
+
+impl OwnedOrBorrowedRawBsonVisitor {
+    pub(super) fn parse_map<'de, A>(map: &mut A) -> Result<MapParse<'de>, A::Error>
+    where
+        A: serde::de::MapAccess<'de>,
+    {
+        let first_key = match map.next_key::<CowStr>()? {
+            Some(k) => k,
+            None => {
+                return Ok(MapParse::Leaf(
+                    RawBson::Document(RawDocumentBuf::new()).into(),
+                ))
+            }
+        };
+        Ok(MapParse::Leaf(match first_key.0.as_ref() {
+            "$oid" => {
+                let oid: ObjectId = map.next_value()?;
+                RawBsonRef::ObjectId(oid).into()
+            }
+            "$symbol" => {
+                let s: CowStr = map.next_value()?;
+                match s.0 {
+                    Cow::Borrowed(s) => RawBsonRef::Symbol(s).into(),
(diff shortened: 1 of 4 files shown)


src/raw/serde/bson_visitor.rs
```diff
@@ -44,6 +44,160 @@ use super::{
 /// A visitor used to deserialize types backed by raw BSON.
 pub(crate) struct OwnedOrBorrowedRawBsonVisitor;
 
+pub(super) enum MapParse<'de> {
```
abr-egn, line 47: This is a little awkward; I was hoping to be able to simply delegate via calling `visit_map`, but the need to signal the aggregate case with the first key included means it didn't fit into that signature.

src/raw/serde/seeded_visitor.rs
```diff
@@ -272,122 +264,88 @@ impl<'a, 'de> Visitor<'de> for SeededVisitor<'a, 'de> {
...
+
+                                Ok(ElementType::JavaScriptCodeWithScope)
+                            }
+                            RawBsonRef::DbPointer(dbp) => {
+                                self.append_string(dbp.namespace);
+                                self.buffer.append_bytes(&dbp.id.bytes());
+                                Ok(ElementType::DbPointer)
+                            }
+                            // No other `Leaf` arms are emitted by `parse_map`
+                            _ => unreachable!(),
```
abr-egn, line 343: I considered introducing a new enum so this could be exhaustive, but I believe this is the only place where case-by-case matching of this sort will be needed so it didn't seem worth it.

src/raw/serde/seeded_visitor.rs
```diff
@@ -272,122 +264,95 @@ impl<'a, 'de> Visitor<'de> for SeededVisitor<'a, 'de> {
...
+                                .copy_from_slice(length_index..length_index + 4, &length_bytes);
+
+                            Ok(ElementType::JavaScriptCodeWithScope)
+                        }
+                        RawBsonRef::DbPointer(dbp) => {
+                            self.append_string(dbp.namespace);
+                            self.buffer.append_bytes(&dbp.id.bytes());
+                            Ok(ElementType::DbPointer)
+                        }
+                        RawBsonRef::Double(d) => self.visit_f64(d),
```
abr-egn, line 341: Previously in this PR this was just a single catchall `_ => unreachable()` but I realized I was uncomfortable with that assertion; while it's true that at the moment none of these can be emitted from `parse_map` that could change in the future and would make for a very unpleasant runtime failure.

isabelatkinson (approved):
sorry for the slow review - lgtm!

PR #483, opened 2024-07-18
RUST-1992 Convert raw deserializer to use raw document iteration

abr-egn:
RUST-1992

This removes quite a bit of duplicated binary parsing code and brings us closer to having separation of concerns between binary parsing and serde deserialization.  IMO, the deserializer and helper structs are now easier to understand as well - rather than a mutable "root deserializer" that's threaded downwards, each deserializer or access struct just gets an immutable ref to the thing it's mapping into serde.

### DIFF
--- a/src/de/raw.rs
+++ b/src/de/raw.rs
@@ -2328,7 +2328,7 @@ impl<'de> serde::de::MapAccess<'de> for RegexAccess2<'de> {
             RegexDeserializationStage::Options => "options",
             RegexDeserializationStage::Done => return Ok(None),
         };
-        seed.deserialize(BorrowedStrDeserializer::new(name))
+        seed.deserialize(FieldDeserializer { field_name: name })
             .map(Some)
     }
 
@@ -2419,15 +2419,12 @@ impl<'de> serde::de::MapAccess<'de> for CodeWithScopeAccess2<'de> {
     where
         K: serde::de::DeserializeSeed<'de>,
     {
-        match self.stage {
-            CodeWithScopeDeserializationStage::Code => seed
-                .deserialize(BorrowedStrDeserializer::new("$code"))
-                .map(Some),
-            CodeWithScopeDeserializationStage::Scope => seed
-                .deserialize(BorrowedStrDeserializer::new("$scope"))
-                .map(Some),
-            CodeWithScopeDeserializationStage::Done => Ok(None),
-        }
+        let field_name = match self.stage {
+            CodeWithScopeDeserializationStage::Code => "$code",
+            CodeWithScopeDeserializationStage::Scope => "$scope",
+            CodeWithScopeDeserializationStage::Done => return Ok(None),
+        };
+        seed.deserialize(FieldDeserializer { field_name }).map(Some)
     }
 
     fn next_value_seed<V>(&mut self, seed: V) -> std::result::Result<V::Value, Self::Error>
@@ -2517,7 +2514,7 @@ impl<'de> serde::de::MapAccess<'de> for DbPointerAccess2<'de> {
             DbPointerDeserializationStage::Id => "$id",
             DbPointerDeserializationStage::Done => return Ok(None),
         };
-        seed.deserialize(BorrowedStrDeserializer::new(name))
+        seed.deserialize(FieldDeserializer { field_name: name })
             .map(Some)
     }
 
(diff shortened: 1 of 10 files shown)


src/de/raw.rs
```diff
@@ -393,41 +225,53 @@ impl<'de, 'a> serde::de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
-            ElementType::String => visitor.visit_enum(self.deserialize_str()?.into_deserializer()),
+        match self.element.element_type() {
+            ElementType::String => visitor.visit_enum(self.get_string()?.into_deserializer()),
             ElementType::EmbeddedDocument => {
-                self.access_document(|access| visitor.visit_enum(access))
+                let doc = match self.value()? {
+                    RawBsonRef::Document(doc) => doc,
+                    _ => {
+                        return Err(Error::deserialization(
+                            "internal error: unexpected non-document",
```
abr-egn, line 250: There are a few places where I had to add this kind of error; these should never happen in practice because the `value()` match is always immediately preceded by an `element_type()` check.

src/de/raw.rs
```diff
@@ -1244,118 +1034,80 @@ enum BinaryDeserializationStage {
...
+            CodeWithScopeDeserializationStage::Code => CodeWithScopeDeserializationStage::Scope,
+            CodeWithScopeDeserializationStage::Scope => CodeWithScopeDeserializationStage::Done,
+            CodeWithScopeDeserializationStage::Done => return Err(Error::EndOfStream),
+        };
+        Ok(value)
     }
 }
 
-impl<'de, 'a, 'b> serde::de::Deserializer<'de> for &'b mut CodeWithScopeDeserializer<'de, 'a> {
+impl<'a, 'de> serde::de::Deserializer<'de> for &'a CodeWithScopeAccess<'de> {
```
abr-egn, line 1088: For the various `*Access` types that needed to be updated, I found it simpler to have them implement both the `Access` and `Deserializer` traits rather than splitting across two types.

src/raw/iter.rs
```diff
@@ -256,20 +267,69 @@ impl<'a> RawElement<'a> {
         })
     }
 
+    pub(crate) fn value_utf8_lossy(&self) -> Result<Option<Utf8LossyBson<'a>>> {
```
abr-egn, line 270: Something like this could hypothetically become part of the public API, although we'd want to think about what it should return along the axes of (restricted enum, full enum) x (always owned, some kind of cow).

Personally, because parsing invalid utf8 is such a corner case, what I'd like to do for 3.0 is remove support for that from the Serde deserialization entirely and _just_ have it exposed via a method like this on the raw document iteration.  That would give an escape hatch for those few times it's needed and enable quite a bit of simplification in the deserializer.  If that sounds reasonable, I'll file a bug for that.

src/de/raw.rs
```diff
@@ -1,475 +1,317 @@
...
-    /// [`MapAccess`] for documents or a [`SeqAccess`] for arrays.
-    fn deserialize_document<V>(
-        &mut self,
-        visitor: V,
-        hint: DeserializerHint,
-        document_type: DocumentType,
-    ) -> Result<V::Value>
+    /// Deserialize the element, using the type of the element along with the
+    /// provided hint to determine how to visit the data.
+    fn deserialize_hint<V>(&self, visitor: V, hint: DeserializerHint) -> Result<V::Value>
```
abr-egn, line 68: Github's diffing, unsurprisingly, got very confused with these changes.  This can be directly compared with the previous version's `deserialize_next` method; in particular, the cases in the the main `match` in both are in the same order so it shouldn't be too hard to scan down them side-by-side.

isabelatkinson (commented):
one small suggestion but changes look good otherwise! have you done any benchmarking on this? I can't imagine there would be any major regression but I think it would be worth a sanity check

abr-egn:
> have you done any benchmarking on this? I can't imagine there would be any major regression but I think it would be worth a sanity check

Good call, and yeah, <1% variation in benchmarks 🙂

src/raw/iter.rs
abr-egn: Filed RUST-1998 for this.

isabelatkinson (approved):
> <1% variation in benchmarks

great, thanks for checking!

PR #485, opened 2024-07-23
RUST-1992 Minor parsing cleanup

abr-egn:
RUST-1992

This PR:
* Relocates the remaining raw field parsing from `crate::de` to `crate::raw` to go with the existing other such functions there
* Removes duplicate "read length and contents to buffer" logic from `crate::document` in favor of the function in `crate::de`.
* Cleans up error construction a little bit.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -39,7 +39,7 @@ use crate::{
 };
 
 use ::serde::{
-    de::{DeserializeOwned, Error as _, Unexpected},
+    de::{DeserializeOwned, Error as _},
     Deserialize,
 };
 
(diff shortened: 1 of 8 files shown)


src/de/mod.rs
```diff
@@ -181,8 +155,10 @@ where
     Deserialize::deserialize(de)
 }
 
-fn reader_to_vec<R: Read>(mut reader: R) -> Result<Vec<u8>> {
-    let length = read_i32(&mut reader)?;
+pub(crate) fn reader_to_vec<R: Read>(mut reader: R) -> Result<Vec<u8>> {
```
abr-egn, line 158: I opted to leave this one here because pulling from a reader can generate IO errors and `crate::de::Error` has a case for those where `crate::raw::Error` doesn't.  Ideally IO error, parsing error, and deserialization error would all be distinct cases in a larger hierarchy but that'll have to wait for RUST-1406.

isabelatkinson (approved):

PR #486, opened 2024-07-24
RUST-1992 Factor raw bson encoding out of RawDocumentBuf

abr-egn:
RUST-1992

This takes the actual "write to buffer as bytes" code out of `RawDocumentBuf`; my theory is that `RawSerializer` can be reworked to a wrapper around `RawWriter` the way `RawDeserializer` is for `RawElement`.  Even if I'm wrong, I think this is a little cleaner.

I also pruned some string-writing utilities that had proliferated a bit.

### DIFF
--- a/src/raw/document_buf.rs
+++ b/src/raw/document_buf.rs
@@ -223,7 +223,9 @@ impl RawDocumentBuf {
     ///
     /// If the provided key contains an interior null byte, this method will panic.
     pub fn append_ref<'a>(&mut self, key: impl AsRef<str>, value: impl Into<RawBsonRef<'a>>) {
-        raw_writer::RawWriter::new(&mut self.data).append(key.as_ref(), value.into())
+        raw_writer::RawWriter::new(&mut self.data)
+            .append(key.as_ref(), value.into())
+            .unwrap()
     }
 
     /// Convert this [`RawDocumentBuf`] to a [`Document`], returning an error
--- a/src/raw/document_buf.rs
+++ b/src/raw/document_buf.rs
@@ -225,7 +225,7 @@ impl RawDocumentBuf {
     pub fn append_ref<'a>(&mut self, key: impl AsRef<str>, value: impl Into<RawBsonRef<'a>>) {
         raw_writer::RawWriter::new(&mut self.data)
             .append(key.as_ref(), value.into())
-            .unwrap()
+            .expect("key should not contain interior null byte")
     }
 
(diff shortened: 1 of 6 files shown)


src/ser/mod.rs
```diff
@@ -40,19 +40,18 @@ use crate::{
 };
 use ::serde::{ser::Error as SerdeError, Serialize};
 
-fn write_string<W: Write + ?Sized>(writer: &mut W, s: &str) -> Result<()> {
-    writer.write_all(&(s.len() as i32 + 1).to_le_bytes())?;
-    writer.write_all(s.as_bytes())?;
-    writer.write_all(b"\0")?;
-    Ok(())
+pub(crate) fn write_string(buf: &mut Vec<u8>, s: &str) {
```
abr-egn, line 43: Updating these to work on a `&mut Vec<u8>` instead of a `Write` impl removed most of the need for errors and we're never actually. writing to anything else.

src/raw/document_buf/raw_writer.rs
```diff
@@ -0,0 +1,102 @@
...
+        let new_len = (self.data.len() as i32).to_le_bytes();
+        self.data[0..4].copy_from_slice(&new_len);
+    }
+
+    fn append_string(&mut self, value: &str) {
+        write_string(&mut self.data, value)
+    }
+
+    fn append_cstring(&mut self, value: &str) {
+        write_cstring(&mut self.data, value).unwrap();
```
abr-egn, line 100: The equivalent code path in `DocumentBuf::append_ref` also panic'd.

src/raw/document_buf.rs
```diff
@@ -221,103 +223,9 @@ impl RawDocumentBuf {
...
-        // update element type
-        self.data[original_len - 1] = element_type as u8;
-        // append trailing null byte
-        self.data.push(0);
-        // update length
-        let new_len = (self.data.len() as i32).to_le_bytes();
-        self.data[0..4].copy_from_slice(&new_len);
+        raw_writer::RawWriter::new(&mut self.data)
+            .append(key.as_ref(), value.into())
+            .unwrap()
```
abr-egn, line 228: Previously, the panic if the key had null was part of the inner implementation of `append_cstring`; this way the panic is kept in the function that actually documents that behavior.

isabelatkinson: makes sense, can we make this an `expect` with a "key should not contain interior null byte" message?

isabelatkinson (approved):
lgtm! one small suggestion, no need for re-review either way

PR #492, opened 2024-08-19
RUST-2017 Add method to construct an `ObjectId` from its parts

tyilo:


abr-egn:
Can I ask about the use case for this?  Synthesizing ObjectIds seems likely to run into unexpected collisions or other subtle issues.

tyilo:
> Can I ask about the use case for this?
The use case is that I want to find documents in a specific time interval. The collection doesn't have any indexes, however the auto-created `_id` can be used for this.

> Synthesizing ObjectIds seems likely to run into unexpected collisions or other subtle issues.\
This is already possible using `ObjectId::from_bytes`

abr-egn (approved):
LGTM!

isabelatkinson (commented):
changes look good, can we add a simple test for this in `src/tests/modules/oid.rs`?

tyilo:
> changes look good, can we add a simple test for this in `src/tests/modules/oid.rs`?

Done

isabelatkinson (approved):
lgtm, thanks for your contribution!

PR #493, opened 2024-08-19
update time and ahash to allow building with Rust 1.80

xaverdh:


xaverdh:
Ok, this would take a lot more effort to make it build with newer rust than I thought, closing for the moment (for reference I arrived here as a transitive dependency of `runiq`).

abr-egn:
For what it's worth, I don't think the bson crate needs to be updated for the Rust 1.80 `time` situation; both the most recent crates.io release and the repo at head compile just fine under 1.80.

xaverdh:
ah, you are correct, the version that was pulled in is an old one, sorry for the noise!

Issue #494, opened 2024-09-03
Impl Hash/Eq on BSON


NineLord:
Hi,
I would like to be able to use `BSON` as key in `HashMap`, aka implement `Hash`/`Eq` on `BSON`.
I understand that it isn't a best practice to do that, due to `BSON::Document` and `BSON::Double`, but having this ability (even behind a feature flag) would help a lot.
`serde_json::Value` does support that too.

PR #495, opened 2024-09-03
RUST-2027 Impl Hash/Eq for BSON

NineLord:
#494
Summery of changes:
* Added feature `hashable` to require the user to choose to opt-in using `Bson` as hash key.
 Although, `serde_json` didn't bother with having this behind the feature flag, so If you like I can remove it.
* `Bson::Double(f64)` - implemented the same solution as [`serde_json`](https://docs.rs/serde_json/1.0.127/src/serde_json/number.rs.html#53-70).
* `Bson::Document(Document)` - implemented the same solution as [`serde_json`](https://docs.rs/serde_json/1.0.127/src/serde_json/map.rs.html#395-397).
* `Bson::JavaScriptCodeWithScope(JavaScriptCodeWithScope)` - Able to impl Hash/Eq only if Document is impl Hash/Eq.
* `Bson::{RegularExpression(Regex), Binary(Binary), Decimal128(Decimal128), DbPointer(DbPointer)}` - Added Derive Hash/Eq without feature flag `hashable`, no reason for it not to implement those by default.
* Added unit test to show that using Bson as key works.

I didn't implement it yet, but if the above feature is acceptable,
it might be worth adding another feature like `not(preserve_order)` from `serde_json`,
to change `Bson::Document` to use `BTreeMap` instead of `indexmap` for more efficient hashing (like [`serde_json`](https://docs.rs/serde_json/1.0.127/src/serde_json/map.rs.html#390)).
I don't mind adding this as a separate PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ chrono-0_4 = ["chrono"]
 uuid-1 = []
 # if enabled, include API for interfacing with time 0.3
 time-0_3 = []
-# If enabled, implement Hash/Eq
+# If enabled, implement Hash/Eq for Bson and Document
 hashable = []
 serde_path_to_error = ["dep:serde_path_to_error"]
 # if enabled, include serde_with interop.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ chrono-0_4 = ["chrono"]
 uuid-1 = []
 # if enabled, include API for interfacing with time 0.3
 time-0_3 = []
+# If enabled, implement Hash/Eq
+hashable = []
 serde_path_to_error = ["dep:serde_path_to_error"]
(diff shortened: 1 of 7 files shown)


abr-egn (approved):

abr-egn:
Thank you for the well-thought-out PR!  The changes look good to me and I've started a test run; once the tests pass and this is reviewed by another team member we'll merge it in.

abr-egn:
On the topic of switching `Document` to use `BTreeMap`, I'm a bit more hesitant, but if you have the time to put together a PR and a basic benchmark it's certainly something we'd consider 🙂

isabelatkinson (changes requested):
a few docs suggestions, otherwise looks good. thanks for your contribution!

isabelatkinson (approved):

PR #496, opened 2024-09-03
RUST-2028 Fix Decimal128 panic when parsing strings w/o a char boundary at idx 34

arthurprs:
Fix Decimal128 panic when parsing strings w/o a char boundary at idx 34 (Coefficient::MAX_DIGITS).

### DIFF
--- a/src/decimal128.rs
+++ b/src/decimal128.rs
@@ -333,6 +333,7 @@ pub enum ParseError {
     Overflow,
     Underflow,
     InexactRounding,
+    Unparseable,
 }
 
 impl fmt::Display for ParseError {
@@ -344,6 +345,7 @@ impl fmt::Display for ParseError {
             ParseError::Overflow => write!(f, "overflow"),
             ParseError::Underflow => write!(f, "underflow"),
             ParseError::InexactRounding => write!(f, "inexact rounding"),
+            ParseError::Unparseable => write!(f, "unparseable"),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/decimal128.rs
```diff
@@ -469,6 +469,16 @@ impl std::str::FromStr for ParsedDecimal128 {
 }
 
 fn round_decimal_str(s: &str, precision: usize) -> Result<&str, ParseError> {
+    // TODO: In 1.80+ there's split_at_checked to make sure the split doesn't
+    // panic if the index doesn't falls at a codepoint boundary, until then
+    // we can check it with s.is_char_boundary(precision)
+    if !s.is_char_boundary(precision) {
+        // a non-digit (all single byte utf8) would trigger a ParseIntError::InvalidDigit,
+        // so here we generate a ParseIntError::InvalidDigit kind of error too.
+        return Err(ParseError::InvalidCoefficient(
```
abr-egn, line 478: I think I'd rather have a new enum variant for `ParseError` (something like `ParseError::Unparseable`) than synthesize an error this way; that way it could also carry the source string to make debugging easier.

arthurprs: I'm not saying I disagree, but no other error variants carry more information than this one.

`InvalidDigit` is arguably correct, as that would be the error if the split was character-based (instead of byte-based). The only reason I didn't do that is that it's more and slower code (linear vs. constant time).

abr-egn: If we could directly construct an `InvalidDigit` I'd have less concern.  My worry is that by synthesizing it this way it'll only _probably_ continue to be `InvalidDigit`; the internals of `ParseIntError` are private so there are no guarantees of stability.  If the error changes in the future to include the offending text, for example, it'll be very confusing for downstream users of this crate to see `"❤"` that doesn't actually appear in their data.

Adding a new variant to `ParseError` avoids that issue, and if we're doing that it may as well carry the text.

arthurprs: I'll add a variant. But also, I'd like to point out that InvalidDigit (and any variant of ParseIntError for that matter) do _not_ carry information about the input, such as the offending character.

abr-egn (approved):

PR #500, opened 2024-10-13
test(fuzz): read -> write bson

LeoDog896:
Runs `to_writer` in fuzz tests. Makes the fuzzer slower, but covers any write errors. Also adds a fuzz test for `from_slice_utf8_lossy`.

Note: I did not check for equivalence in `from_reader` and `to_writer`, as, even for accounting for length, there is still some loss in data ordering from the reader, as is to be expected.


abr-egn (approved):

isabelatkinson (approved):
lgtm - thanks for your contribution!

PR #510, opened 2025-01-08
fuzz: remove malformed_length entry

manunio:
Missing malformed_length target is causing build failure at oss-fuzz


abr-egn (approved):
Thanks for catching this!  I'll have to look to see why our normal process didn't pick up on this failure.

PR #513, opened 2025-01-23
RUST-2003 Binary vector subtype support

isabelatkinson:
RUST-2003


src/binary/vector.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// final byte. For example, the vector in the previous example with a padding of 4 would
+    /// represent a 12-bit vector containing the following values:
+    ///
+    /// ```text
+    /// [1, 1, 1, 0, 1, 1, 1, 0, 1, 1, 1, 0]
+    /// ```
+    ///
+    /// Padding must be within 0-7 inclusive. Padding must be 0 or unspecified if the provided
+    /// vector is empty.
+    pub fn new(vector: Vec<u8>, padding: impl Into<Option<u8>>) -> Result<Self> {
```
isabelatkinson, line 93: the spec requires this type to be valid upon construction, hence the opaque type and result-returning `new` method

src/binary/vector.rs
```diff
@@ -0,0 +1,288 @@
...
+/// A vector of numeric values. This type can be converted into a [`Binary`] of subtype
+/// [`BinarySubtype::Vector`].
+///
+/// ```rust
+/// # use bson::binary::{Binary, Vector};
+/// let vector = Vector::Int8(vec![0, 1, 2]);
+/// let binary = Binary::from(vector);
+/// ```
+///
+/// `Vector` serializes to and deserializes from a `Binary`.
```
isabelatkinson, line 24: I don't think there's much value in treating this type as anything other than a convenience wrapper around `Binary` when it comes to serde, but we can revisit if a use case arises

abr-egn (approved):
LGTM!

Issue #502, opened 2024-11-02
base64 crate in dependency is too old


zonyitoo:
The current latest release of `base64` is v0.22.1: https://crates.io/crates/base64

abr-egn:
I've filed RUST-2089 to track this, although unless there's a specific functional issue this is causing it's unlikely to be high priority.  We do happily accept contributions :)

PR #515, opened 2025-02-12
Update base64 crate

barvirm:
Update base64 crate solves: #502


isabelatkinson:
@barvirm Thanks for the PR! I'm going to push a commit to fix a few formatting issues and then this should be good to merge.

isabelatkinson (approved):

PR #519, opened 2025-03-07
RUST-2174 feat(extjson): parse legacy `{"$date": <ms since epoch>}`

xiangjinwu:
According to the specification, there exists a legacy Extended JSON form of Datetime
https://github.com/mongodb/specifications/blob/master/source/extended-json/extended-json.md#datetime
> Mongoexport 2.4 and the Java driver always transform a Datetime object into an Extended JSON string of the form `{"$date": <ms since epoch>}`. This form has the problem of a potential loss of precision or range on the Datetimes that can be represented.

Specially, this legacy form is called `JsonMode.STRICT` in the Java `JsonWriterSettings`, while the canonical form is `JsonMode.EXTENDED`. The former is still used by the Debezium connector for MongoDB.

This PR adds the ability to parse such legacy format, without conflicting with the standard ones.

### DIFF
--- a/src/extjson/models.rs
+++ b/src/extjson/models.rs
@@ -256,6 +256,7 @@ pub(crate) struct DateTime {
 pub(crate) enum DateTimeBody {
     Canonical(Int64),
     Relaxed(String),
+    Legacy(i64),
 }
 
 impl DateTimeBody {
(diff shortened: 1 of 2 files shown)


abr-egn (approved):

abr-egn:
Thank you for the contribution!  The CI failures were unrelated and this was a straightforward addition, so I've merged it in and it'll go out in the next release (should happen in the next week or so).

PR #524, opened 2025-03-25
Allow passing owned `Vector` to `doc!`

isabelatkinson:


src/binary/vector.rs
```diff
@@ -246,10 +246,10 @@ impl TryFrom<Binary> for Vector {
     }
 }
 
-// Convenience impl to allow passing a Vector directly into the doc! macro. From<Vector> is already
+// Convenience impl to allow passing a Vector directly into the doc! macro. From<&Vector> is already
```
isabelatkinson, line 249: this was backwards - the blanket impl is for `&T` when `From<T>` is implemented

PR #525, opened 2025-03-26
Allow serializing BSON to a binary stream

mccullocht:


abr-egn:
I'm pretty hesitant about this approach - it adds a lot of complexity to serialization (which is already overdrawn on the complexity budget) for a relatively narrow use case, and it also seems likely to be a significant performance hit because of the double serialization required for the length calculation.

I'm assuming your current approach is to use `bson::to_vec` and then write that into the destination buffer via `bytes::BufMut::put` - has profiling shown the allocation or copy there to be substantial cost?

PR #528, opened 2025-04-22
add `uncapped_max_size` feature to allow increasing max BSON size

Trevader24135:
For various projects I'm working on, I'm using BSON as an efficient means to transfer potentially quite a bit of data (~200Mb). From what I can read of the BSON standard, it's expected to support document sizes up to 2Gb, but this implementation is currently hard capped to 16Mb to better service MongoDB.

I figured adding an opt-in feature to increase the cap to the maximum would serve to be useful for more people than just me, without affecting the way that it's currently used within MongoDB applications.

I ran all the unit tests and linting tests and didn't encountered any failures with or without the added feature enabled.

Let me know if I'm off-base thinking that this is a useful addition, or if there are any unforseen issues with expanding the max size. I haven't done extensive reading of this library, so I may be missing something else that needs accounted for.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,9 @@ uuid-1 = []
 time-0_3 = []
 # If enabled, implement Hash/Eq for Bson and Document
 hashable = []
+# If enabled, the maximum document size will be increased from the MongoDB cap
+# of 16 MB to the BSON spec maximum size of 2^31 - 1
+uncapped_max_size = []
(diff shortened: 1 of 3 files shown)


abr-egn:
Hi - sorry for the delay getting back to you here.  This sounds very reasonable, we just need to make sure we're not accidentally depending on this limit in the driver before we merge the feature in!

abr-egn (approved):

PR #531, opened 2025-05-08
RUST-1906 Remove human readable options

isabelatkinson:
I removed all of the options types and `with_options` methods, as I think the likelihood that we'll introduce new option is low.


isabelatkinson:
The fuzzer failure is unrelated - the test is expecting a document containing `f64::NAN` to roundtrip, but the equality comparison fails because `f64::NAN` != `f64::NAN`. Filed RUST-2213 to address this.

abr-egn (approved):

src/ser/mod.rs
```diff
@@ -154,37 +140,12 @@ where
 /// The [`Serializer`] used by this function presents itself as human readable, whereas the
 /// one used in [`to_vec`] does not. This means that this function will produce different BSON than
 /// [`to_vec`] for types that change their serialization output depending on whether
-/// the format is human readable or not. To serialize to a [`Document`] with a serializer that
-/// presents itself as not human readable, use [`to_document_with_options`] with
-/// [`SerializerOptions::human_readable`] set to false.
+/// the format is human readable or not.
```
univerz, line 143: the difference in serialization depending on is_human_readable between `insert_one` (raw) & `update_one` (document) is probably not what the user expects from a binary format

PR #548, opened 2025-06-03
RUST-1798 use simd to optimize utf8 validation

abr-egn:
RUST-1798

Our string parsing has changed pretty significantly since #440 was opened, so I opted to create a new PR rather than try to rebase and rework that one.

### DIFF
--- a/src/raw/error.rs
+++ b/src/raw/error.rs
@@ -1,5 +1,3 @@
-use std::str::Utf8Error;
-
 use crate::spec::ElementType;
 
 /// An error that occurs when attempting to parse raw BSON bytes.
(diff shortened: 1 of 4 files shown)


src/raw/error.rs
```diff
@@ -44,7 +42,7 @@ pub enum ErrorKind {
     MalformedValue { message: String },
 
     /// Improper UTF-8 bytes were found when proper UTF-8 was expected.
-    Utf8EncodingError(Utf8Error),
```
abr-egn, line 47: The error value provided by `simdutf8::basic::from_utf8` is zero-sized (because it omits tracking error location in favor of speed) so I didn't see much reason to capture it here; this also means we're not exposing a 0.x crate type in our public API.

PR #550, opened 2025-06-04
RUST-1998 Remove lossy utf8 as a decoder option

abr-egn:
RUST-1998

This removes lossy utf8 decoding as a public option for the bson serde decoder, and as a consequence many of the API entry points.  Dealing with lossy utf8 is now done in one of three ways depending on the specific need:
* The existing `Utf8LossyDeserialization` helper type, if the user is going via `serde` and knows that particular fields might have bad data;
* The newly-exposed `RawElement::value_utf8_lossy` method, which allows fine-grained iteration over potentially-invalid raw data, or
* The new `RawDocument::to_document_utf8_lossy` method, which is a convenience wrapper over the above to bulk-convert an entire `RawDocument` at once.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,12 +50,27 @@ pub enum ErrorKind {
         /// The kind of error that occurred.
         kind: ValueAccessErrorKind,
     },
+
+    /// A wrapped deserialization error.
+    /// TODO RUST-1406: collapse this
+    #[error("Deserialization error")]
+    DeError(crate::de::Error),
 }
 
 impl From<ErrorKind> for Error {
     fn from(kind: ErrorKind) -> Self {
         Self {
             kind,
+            key: None,
+            index: None,
+        }
+    }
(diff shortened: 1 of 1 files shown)


isabelatkinson (commented):
the new API looks good! we are using `from_slice_utf8_lossy` in [insert](https://github.com/mongodb/mongo-rust-driver/blob/e18a7a12a69fc53a74857c04f8997f8511b22770/src/operation/insert.rs#L136) and [update](https://github.com/mongodb/mongo-rust-driver/blob/e18a7a12a69fc53a74857c04f8997f8511b22770/src/operation/update.rs#L173) response parsing in the driver, which is probably a hot path for some users - would the wrapper type be a fine replacement here regarding performance?

isabelatkinson (dismissed):

isabelatkinson (approved):

PR #552, opened 2025-06-05
RUST-1406 Convert raw errors to standard error type

isabelatkinson:


src/de/error.rs
```diff
@@ -76,9 +76,9 @@ impl From<string::FromUtf8Error> for Error {
     }
 }
 
-impl From<crate::raw::Error> for Error {
-    fn from(value: crate::raw::Error) -> Self {
-        Self::deserialization(value)
+impl From<crate::error::Error> for Error {
```
isabelatkinson, line 79: this is a temp change before `de::Error` is removed

PR #554, opened 2025-06-11
RUST-1992 Make serde an optional feature

abr-egn:
RUST-1992

This splits everything `serde`-related into feature-gated optional functionality, with a _whole lot_ of downstream changes:
* Functions got renamed to make it clear which of the encode/decode or serialize/deserialize family they were part of (things like `to_vec` existed in both and were way too easy to confuse otherwise)
* `Document::to_writer` previously round-tripped through `serde` for ... reasons? ... and now just directly encodes (and I added a convenience `encode_to_vec` method because 99% of the time that's the writer target).
* A lot of things that were calling `bson::to_vec(doc)` are now directly encoding
* `RawDocumentBuf::append` and `RawArrayBuf::push` previously would panic if given a value that couldn't be encoded, which made for a slightly more convenient API but is also egregiously unrustacean, so they now return `Result<()>`.
* The `rawdoc!` macro, which uses the above, still has the panic behavior because there's not really anything else it can do.  Everywhere else downstream of those is properly fallible now.
* Core functionality (constants, utility functions) to binary bson handling was moved out of serde modules.

Things under the umbrella of RUST-1992 to come in follow-up PRs:
* user documentation for the distinction between "encode" (concrete bson type to bytestream) and "serialize" (arbitrary type, possibly containing bson, to arbitrary format)
* `serde_json` might also be reasonable to be an optional dependency.  I'm less certain of this, though, it'll need some exploration.
* clean up and consolidation of the core binary-format bson bytestream handling functions, they're still a bit scattered

### DIFF
--- a/src/raw/iter.rs
+++ b/src/raw/iter.rs
@@ -155,7 +155,7 @@ impl<'a> RawElement<'a> {
         })
     }
 
-    pub fn len(&self) -> usize {
+    pub fn size(&self) -> usize {
         self.size
     }
 


src/raw/document_buf.rs
```diff
@@ -112,11 +114,13 @@ impl RawDocumentBuf {
...
     /// ```
-    pub fn from_document(doc: &Document) -> Result<RawDocumentBuf> {
-        let mut data = Vec::new();
-        doc.to_writer(&mut data).map_err(Error::malformed_value)?;
-
-        Ok(Self { data })
+    pub fn from_document(doc: &Document) -> Result<Self> {
+        let mut out = RawDocumentBuf::new();
+        for (k, v) in doc {
+            let val: RawBson = v.clone().try_into()?;
```
abr-egn, line 120: _In theory_, `RawDocumentBuf::from_document` should be able to iterate over a borrowed `Document` and build the new `RawDocumentBuf` without any intermediate copies needed.  Where this breaks down is that `append`ing to a bson byte buffer is defined at its core in terms of `RawBsonRef`s, and there's no way to get one of those from a managed-type `&Bson` without allocating intermediate storage.

We could add logic to translate the managed `&Bson` type directly to bytes, but we've already got too many slightly-different versions of that logic already.  If the `clone()` here turns out to be a profiling hotspot we can revisit.

src/raw/document_buf.rs
```diff
@@ -211,11 +215,14 @@ impl RawDocumentBuf {
...
+        &mut self,
+        key: impl AsRef<str>,
+        value: impl BindRawBsonRef,
+    ) -> crate::error::Result<()> {
         value.bind(|value_ref| {
-            raw_writer::RawWriter::new(&mut self.data)
-                .append(key.as_ref(), value_ref)
-                .expect("key should not contain interior null byte")
+            raw_writer::RawWriter::new(&mut self.data).append(key.as_ref(), value_ref)
+            //.expect("key should not contain interior null byte")
```
isabelatkinson, line 225:
```suggestion

```

src/raw/array_buf.rs
```diff
@@ -181,3 +184,24 @@ impl Default for RawArrayBuf {
...
+        }
+        Ok(tmp)
+    }
+}
+
+impl TryFrom<crate::Array> for RawArrayBuf {
+    type Error = crate::error::Error;
+
+    fn try_from(value: crate::Array) -> Result<Self, Self::Error> {
+        Self::try_from(&value)
```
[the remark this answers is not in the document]
abr-egn: Good call; also updated the similar impl for `RawDocumentBuf`.

src/raw/document_buf.rs
abr-egn: My preference would be to remove `from_document` entirely in favor of the `TryFrom` impls and have a big documentation stanza as part of the work for RUST-2228.

isabelatkinson: sounds good to me

isabelatkinson (dismissed):
LGTM, I can re-approve if you want to do the `from_document` removal as part of this PR

abr-egn:
> LGTM, I can re-approve if you want to do the `from_document` removal as part of this PR

I think this PR's big enough so I'll leave that for later, but I did have to do a rebase :)

PR #555, opened 2025-06-13
RUST-1406 Add type-specific errors to standard error type

isabelatkinson:
This largely maintains the existing structure of the type-specific errors for ease of migration, with a few things collapsed/removed where it made sense.


isabelatkinson:
going to change a few small things - will re-request review when this is ready again!

src/error.rs
```diff
@@ -31,18 +39,56 @@ impl std::fmt::Display for Error {
 }
 
 /// The types of errors that can occur in the `bson` crate.
-#[derive(Debug, Error)]
+#[derive(Clone, Debug, Error)]
 #[non_exhaustive]
 pub enum ErrorKind {
+    /// An error related to the [`Binary`](crate::Binary) type occurred.
```
abr-egn, line 45: This is exposing how arbitrary it is which of our errors carry a string and which carry a subkind enum; that's got a bit of a code smell to it but I can't decide if it's _actually_ a problem worth addressing.  Do you have thoughts here?

isabelatkinson: Yeah, I don't _love_ this design; I agree that it's arbitrary and inconsistent. My concern, however, is that users who do have some kind of programmatic use for the values in these errors will be less inclined to migrate to the new version, so my instinct here is to err on the side of portability for lower-priority changes. Certainly open to other designs here if you had something else in mind though!

abr-egn: Sketch of a proposal:
* Hoist `message` to an optional field of `Error`, with corresponding code in the `Display` impl to append it if it's present.
* All variants of `ErrorKind` are required to be `#[non_exhaustive]` struct variants, even if they're empty (which a lot now will be with `message` stripped out)
* Any sub-`Kind` types follow the same pattern (sub-`Kinds` also don't need to have `message` since they're ultimately attached to an `Error`!)

This makes the error space a lot more uniform, gives us a lot of flexibility for future error evolution, and doesn't introduce substantial migration burden.

src/error.rs
```diff
@@ -98,8 +104,9 @@ pub enum ErrorKind {
     },
 
     /// A [`std::io::Error`] occurred.
-    #[error("An IO error occurred: {0}")]
-    Io(std::io::Error),
+    #[error("An IO error occurred")]
+    #[non_exhaustive]
+    Io {},
```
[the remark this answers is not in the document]
abr-egn: We might want to carry the `std::io::ErrorKind` but that can be added later if it's asked for :)

abr-egn (approved):
LGTM!

PR #566, opened 2025-07-03
RUST-1748 Add serde_conv macro for object_id::AsHexString and object_id::FromHexString

JamieTsai1024:
* Add custom `serde_conv` macro, `my_serde_conv`, to:
  1. Support rustdoc documentation for structs 
  2. Support serializers that may fail by wrapping the return type for `serialize` in a `Result`
     * Original implementation assumes that serialization cannot fail, which is not the case for many of our converters. For example, serializing a String to ObjectId may fail in `object_id::FromHexString`.
* Refactor full struct `(de)serialize_as` implementations to use `my_serde_conv` macro 
  * `object_id::AsHexString`
  * `object_id::FromHexString`

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -428,18 +428,6 @@ macro_rules! rawdoc {
 ///
 /// This macro generates a `SerializeAs`/`DeserializeAs` implementation for a given type,
 /// with optional struct-level attributes like `#[derive(...)]` or `/// doc comments`.
-///
-/// # Example
-/// ```
-/// serde_conv_doc!(
-///     /// Custom serialization for `MyType`.
-///     #[derive(Debug, Clone)]
-///     pub MyTypeSerdeConv,
-///     MyType,
-///     |x| Ok(x.to_string()),
-///     |s: String| Ok(MyType::from_str(&s)?)
-/// );
-/// ```
 macro_rules! serde_conv_doc {
     ($(#[$meta:meta])* $vis:vis $m:ident, $t:ty, $ser:expr, $de:expr) => {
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -421,3 +421,56 @@ macro_rules! rawdoc {
         object
     }};
 }
+
+/// Same as [`serde_with::serde_conv!`], but with support for rustdoc documentation for the struct.
+#[macro_export]
```
abr-egn, line 426: The rules around macros are weird but since this is intended as an internal-only macro I don't think we want `#[macro_export]` - that'll make it part of the public API of the crate.  Instead, you can add `pub(crate) use my_serde_conv;` after the `macro_rules!` definition to bring it into the normal namespace.

src/macros.rs
```diff
@@ -421,3 +421,56 @@ macro_rules! rawdoc {
...
+macro_rules! my_serde_conv {
```
abr-egn, line 427: I'd suggest naming this something like `serde_conv_doc`.

src/macros.rs
```diff
@@ -421,3 +421,56 @@ macro_rules! rawdoc {
...
+    ($(#[$meta:meta])* $m:ident, $t:ty, $ser:expr, $de:expr) => {$crate::serde_conv!(pub(self) $m, $t, $ser, $de);};
```
abr-egn, line 428: I don't think this matcher line is needed (and I think it'd be a compile error if anything actually matched on it, since there's no `$crate::serde_conv!` macro).

abr-egn (approved):
LGTM!

PR #567, opened 2025-07-03
RUST-2240 Move `serde_json` API behind a feature flag

isabelatkinson:
This removes the `extjson` module from the public API, which only exported the now-removed error types, and moves its documentation to `lib.rs`. All of the `serde_json`-related API is now behind a feature flag and lives in `extjson/json.rs`.


abr-egn (approved):
LGTM!

PR #580, opened 2025-07-23
RUST-2228 Normalize conversions between raw types and base crate types

abr-egn:
RUST-2228

This fills in the gaps and removes the redundant explicit conversion methods, making all conversions be via the standard traits.  As part of that, I promoted `Utf8LossyDeserialization` to a more general `Utf8Lossy` wrapper, and tidied up and expanded the documentation in the `raw` module.


src/raw/document.rs
```diff
@@ -633,7 +577,8 @@ impl ToOwned for RawDocument {
     type Owned = RawDocumentBuf;
 
     fn to_owned(&self) -> Self::Owned {
-        self.to_raw_document_buf()
+        // unwrap is ok here because we already verified the bytes in `RawDocumentRef::new`
```
isabelatkinson, line 580: I think this copied comment is outdated - this should reference `RawDocument::decode_from_bytes`

This also got me thinking that `decode_from_bytes` read as a little confusing for these method names; "decode" implies to me that the bytes are being parsed, but we're actually just checking a few invariants and then storing the bytes as-is. I think reverting to `from_bytes` could be more clear. WDYT? (Sorry I missed this in #554!)

isabelatkinson (approved):

Issue #457, opened 2024-02-12
add ObjectId deserialization from sequence to support messagepack


univerz:
de/serializing `ObjectId` works when using `rmp_serde::encode::to_vec_named`, but not with `rmp_serde::encode::to_vec` which stores object as an array without field keys. example how it looks like:

```rust
#[derive(Deserialize, Serialize, Debug)]
struct Test {
    objectid: ObjectId,
    float: f64,
}
let test = Test { objectid: ObjectId::new(), float: 3.14 };

let buf = rmp_serde::encode::to_vec_named(&test)?;
dbg!(rmp_serde::from_slice::<RmpValue>(buf.as_slice())?);
let buf = rmp_serde::encode::to_vec(&test)?;
dbg!(rmp_serde::from_slice::<RmpValue>(buf.as_slice())?);
Ok(())
```

<details><summary>output</summary>
<pre>
rmp_serde::from_slice::<RmpValue>(buf.as_slice())? = Map(
    [
        (
            String(
                Utf8String {
                    s: Ok(
                        "objectid",
                    ),
                },
            ),
            Map(
                [
                    (
                        String(
                            Utf8String {
                                s: Ok(
                                    "$oid",
                                ),
                            },
                        ),
                        String(
                            Utf8String {
                                s: Ok(
                                    "65c9e0a4e3f150942e7f537e",
                                ),
                            },
                        ),
                    ),
                ],
            ),
        ),
        (
            String(
                Utf8String {
                    s: Ok(
                        "float",
                    ),
                },
            ),
            F64(
                3.14,
            ),
        ),
    ],
)
rmp_serde::from_slice::<RmpValue>(buf.as_slice())? = Array(
    [
        Array(
            [
                String(
                    Utf8String {
                        s: Ok(
                            "65c9e0a4e3f150942e7f537e",
                        ),
                    },
                ),
            ],
        ),
        F64(
            3.14,
        ),
    ],
)
</pre>
</details>

because the `ObjectId` behavior is special, i (probably) need something like this added to `ObjectIdVisitor` to make it work (`serialize_object_id_as_hex_string` is not an option):
```rust
#[inline]
fn visit_seq<V>(self, mut visitor: V) -> Result<Self::Value, V::Error>
where
    V: SeqAccess<'de>,
{
    visitor.next_element::<ObjectId>()?.ok_or_else(|| {
        de::Error::custom(format!("failed to deserialize ObjectId from a sequence"))
    })
}
```

[...]

PR #582, opened 2025-07-24
RUST-1886 Provide an opt-in "serialize as bytes" helper for ObjectId

abr-egn:
RUST-1886

The default extjson representation used when serializing can be problematic for some `serde` serializers (see #457).  However, changing it to always directly serialize as a byte array (the "newtype pattern" referenced in that bug) runs the risk of breaking _other_ users who might be, intentionally or not, relying on the current serialization format.

This PR threads the needle by providing a `serde_with` helper so that people who need the byte array serialization behavior can opt into it; care is taken to make sure all the normal serialization paths still work when the helper is enabled, i.e. objectids are still serialized as objectids and not generic arrays.

### DIFF
--- a/src/ser/raw.rs
+++ b/src/ser/raw.rs
@@ -221,7 +221,7 @@ impl<'a> serde::Serializer for &'a mut Serializer {
             }
             #[cfg(feature = "serde_with-3")]
             SerializerHint::ObjectId => {
-                let bytes: [u8; 12] = v.try_into().map_err(|e| Error::serialization(e))?;
+                let bytes: [u8; 12] = v.try_into().map_err(Error::serialization)?;
                 self.serialize_raw(RawBsonRef::ObjectId(ObjectId::from_bytes(bytes)))?;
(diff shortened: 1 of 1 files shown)


src/serde_helpers.rs
```diff
@@ -70,6 +70,40 @@ pub mod object_id {
             Ok(oid.to_hex())
         }
     );
+
+    pub(crate) const OID_BYTES_NEWTYPE: &str = "$__bson_private_oid_bytes";
+    /// When serializing to a non-human-readable destination, will represent the [`ObjectId`] as a
```
isabelatkinson, line 75: If we eliminated the human-readable-specific logic here, could users get the same effect with:

```rust
#[serde_as(as = "IfIsHumanReadable<serde_with::Same, object_id::AsBytes>")]
oid: ObjectId,
```
?
I'd prefer not to be opinionated about human-readable in these helpers if we can avoid it.

PR #590, opened 2025-08-06
Borrow the buffer instead of owning it

LeWimbes:
Similar to the idea in #328, `raw::Serializer` now takes a `&mut Vec<u8>`.
Because the `Vec` is borrowed, the same buffer can be reused for many serializations without reallocating.

* added **`serialize_to_buffer`**: serialize `&T` straight into a caller-supplied `Vec<u8>`.
* **`serialize_to_vec`** still exists. It just allocates a `Vec<u8>` and forwards it to `serialize_to_buffer`.

Breaking change: `raw::Serializer::new()` now expects `&mut Vec<u8>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -474,6 +474,7 @@ pub use self::{
     },
     ser::{
         serialize_to_bson,
+        serialize_to_buffer,
         serialize_to_document,
         serialize_to_raw_document_buf,
         serialize_to_vec,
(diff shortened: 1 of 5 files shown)


isabelatkinson (approved):
Thanks for the contribution! I pushed a fix for a few lint errors and CI is now passing. Going to tag in the other team member for review.

abr-egn (approved):

LeWimbes:
Thanks for fixing the lints and merging. I think #328 can be closed now, since it provides similar functionality.

PR #597, opened 2025-08-19
RUST-2251 document and tidy up serde helpers

abr-egn:
RUST-2251

This adds toplevel documentation to the `serde_helpers` module; I also removed some now-redundant (and, I think, arguably incorrect) `serde_with` trait impls from elsewhere and made it so that the converters can be used without `serde_with`.


abr-egn:
> the updated docs look great! I have a few additional random suggestions from poking around in the generated docs: [isabelatkinson@3c9ebc0](https://github.com/isabelatkinson/bson-rust/commit/3c9ebc0eede4da33d1f86438af69ba00cab3e677)

Applied, thanks :). I can't seem to both fix the merge conflict and preserve the verified status of your patch but I think since it's verified with my key it's fine?

PR #628, opened 2025-12-19
RUST-2316 Expand numeric conversions

### LINKED ISSUE #627: Implement `From<T> for Bson` for more integer types
To make `doc! { "foo": v }` more ergonomic, it would be nice if the following types implemented infallible conversion to `Bson`:

### Smaller integer types
```
i8, i16 -> Bson::Int32
u8, u16 -> Bson::Int32
```

### `NonZero<T>` types where `Bson: From<T>`
This direction is lossless.

### Platform-specific types
```
isize, usize -> Bson::Int64
```

I don't think that this should be added for compatibility reasons, but I wanted to add it as an option.

abr-egn:
Fixes #627.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -26,6 +26,7 @@ use std::{
     convert::TryFrom,
     fmt::{self, Debug, Display, Formatter},
     hash::Hash,
+    num::NonZero,
     ops::Index,
 };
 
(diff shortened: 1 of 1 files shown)


abr-egn:
> Not saying this is a good idea, but I think you can check in the generic array impl that the length is 12 and the type is u8 (using TypeId) and return an object id in that case and an array in other cases.

Ah, true.  I think I'd rather keep the unexpected behavior as "why won't my u8 array conversion compile" instead of "why is this array conversion sometimes returning the wrong bson type at runtime", though.

PR #635, opened 2026-01-08
Implement `Ord` for `BinarySubtype`

tyilo:
I have the following type in my code that I would like to store in a `BTreeMap`:

```rust
enum BsonType {
    Double,
    String,
    Document,
    Array,
    Binary(BinarySubtype),
    Undefined,
    ObjectId,
    Boolean,
    DateTime,
    Null,
    RegularExpression,
    DbPointer,
    JavaScriptCode,
    Symbol,
    JavaScriptCodeWithScope,
    Int32,
    Timestamp,
    Int64,
    Decimal128,
    MinKey,
    MaxKey,
}
```

So I need to implement `Ord` for it. This is easier to do with this change.

### DIFF
--- a/src/spec.rs
+++ b/src/spec.rs
@@ -21,7 +21,7 @@
 
 //! Constants derived from the [BSON Specification Version 1.1](http://bsonspec.org/spec.html).
 
-use std::{convert::From, fmt};
+use std::{convert::From, fmt, hash::Hash};
 
 impl fmt::LowerHex for BinarySubtype {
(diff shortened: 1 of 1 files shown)


tyilo:
> I don't think this works because `UserDefined(123) != Reserved(123)`, but the compare equal with this `Ord` impl.

I fixed changed the `PartialEq` and `Hash` implementation so that `Generic == UserDefined(0x00) == Reserved(0x00)`.

PR #643, opened 2026-02-19
RUST-2354 Add option for max length when parsing cstrings

isabelatkinson:
The maximum length can be configured on an iterator or through new getter methods:

```rust
// document
let key = cstr!("aaaaa");
let doc = rawdoc! { key: "b" };
let iter = doc.iter().max_cstr_parse_len(key.len());
let b = doc.get_with_max_cstr_parse_len(key, key.len());

// array
let array = doc.get_array("array").unwrap();
let iter = array.into_iter().max_cstr_parse_len(5);
let val = array.get_with_max_cstr_parse_len(2, 5);
```

The API additions are exposed behind an `sfp-internal` feature flag. The documentation for them is hidden. This will give us more semver flexibility to make breaking changes if needed. Docs for the new features can be generated locally by running:
```
RUSTDOCFLAGS='-Z unstable-options --document-hidden-items' cargo +nightly doc --all-features --open
```

### DIFF
--- a/src/raw/test.rs
+++ b/src/raw/test.rs
@@ -580,4 +580,39 @@ fn max_cstr_parse_len() {
     let mut iter = array.into_iter().max_cstr_parse_len(key.len() - 1);
     let error = iter.next().unwrap().unwrap_err();
     assert!(matches!(error.kind, ErrorKind::TooLongCStr { .. }));
+
+    let pattern = cstr!("pattern");
+    let options = cstr!("abc");
+    let key = cstr!("r");
+    let regex = RawRegexRef { pattern, options };
+    let doc_with_regex = rawdoc! { key: RawBsonRef::from(regex) };
+
+    let mut iter = doc_with_regex.iter().max_cstr_parse_len(pattern.len());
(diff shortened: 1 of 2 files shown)


src/raw/test.rs
```diff
@@ -508,3 +508,62 @@ proptest! {
         prop_assert_eq!(doc, roundtrip);
     }
 }
+
+#[test]
+#[cfg(feature = "sfp-internal")]
+fn max_cstr_parse_len() {
```
jyemin: Let's add a test to assert that this limit applies to cstring keys within arrays, e.g. an array like

```
["foo", "bar", "baz"] 
```

that is encoded like

```
{"veryveryveryveryveryveryveryveryveryveryveryveryveryveryverylongkeythatissupposedtojustbethearrayindex": "foo", 
 ...
```

isabelatkinson: updated

kevinAlbs (commented):
Looks good with a suggestion to validate regex.

patrickfreed:
Going to give this a quick try in monguard tomorrow to confirm we're not missing any requirements, but the interface LGTM after a quick look.

PR #656, opened 2026-04-09
RUST-2330 Serialize `Facet` types to BSON binary

abr-egn:
RUST-2330

This adds `bson::facet::to_vec`, which serializes arbitrary types that implement `Facet` to BSON bytes, properly handling embedded Rust BSON types.


etc/run-benchmarks.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# make sure we're running in the repo root
+cd $(dirname $0)/..
+
+RUSTFLAGS="--cfg mongodb_internal_bench" cargo +nightly bench tests::bench --all-features
```
abr-egn, line 6: `cargo bench` requires using nightly; I put the benchmarks behind a custom `cfg` flag rather than a feature so they don't get picked up even when using `--all-features`.

src/facet/format.rs
```diff
@@ -0,0 +1,247 @@
...
+    let mut s = Serializer::new();
+    facet_format::serialize_root(&mut s, facet_reflect::Peek::new(value)).map_err(|e| match e {
+        SerializeError::Backend(e) => e,
+        _ => Error::serialization(format!("{e}")),
+    })?;
+    Ok(s.bytes)
+}
+
+#[derive(Debug)]
+struct Serializer {
```
abr-egn, line 41: Because the Facet serialization machinery feeds a stream of events rather than using visitors, the structure here is quite different than the Serde serialization; I find this setup easier to follow but maybe that's just because I wrote it.

.evergreen/config.yml
```diff
@@ -216,9 +216,9 @@ axes:
   - id: "extra-rust-versions"
     values:
       - id: "min"
-        display_name: "1.81 (minimum supported version)"
+        display_name: "1.85 (minimum supported version)"
```
abr-egn, line 219: This along with the edition change allow [let-chains](https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html) which are _really_ handy and this version's well past our declared threshold.

This did come with a bit more churn than I expected (formatted order of imports changed), I can pull that out if you think that'd be better.

src/raw/array.rs
```diff
@@ -239,6 +239,30 @@ impl TryFrom<&RawArray> for Vec<Bson> {
     }
 }
 
+impl TryFrom<&RawArray> for Bson {
```
abr-egn, line 242: These (and similar added elsewhere) made testing more convenient and they seem fine to export.

src/raw/array_buf.rs
```diff
@@ -39,6 +39,7 @@ use super::{document_buf::BindRawBsonRef, RawArrayIter};
 /// Note that accessing elements is an O(N) operation, as it requires iterating through the document
 /// from the beginning to find the requested key.
 #[derive(Clone, PartialEq)]
+#[cfg_attr(feature = "facet-unstable", derive(facet::Facet), facet(opaque))]
```
abr-egn, line 42: All of the module types that derive `Facet` are also tagged with `opaque`; this means that you can't use facet machinery to inspect the internals and unless a [de]serializer special cases them (like `bson::facet::to_vec` does) they can only be [de]serialized via a proxy type (like `bson::facet::ExtJson`).  This seems pretty ideal to me since it means we're not exposing internal field structure.  There's _maybe_ a case for making those structs that have all `pub` fields non-opaque, but that seems more like a "consider it if anyone asks for it" kind of thing.

src/raw/bson.rs
```diff
@@ -18,12 +16,16 @@ use crate::{
...
+    oid::{self, ObjectId},
+    spec::ElementType,
 };
 
 use super::{Error, Result};
 
 /// A BSON value backed by owned raw BSON bytes.
 #[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "facet-unstable", derive(facet::Facet), facet(opaque))]
+#[repr(C)]
```
abr-egn: Facet requires enums to have a specified `repr` since the default unspecified one makes it basically impossible to say anything about in-memory layout.

src/tests/bench.rs
abr-egn, line 1: These show facet to be 3-4x slower than serde, which is a bummer.  I'm not sure how much of that is general facet overhead and how much could be reclaimed from optimization of the facet serializer.  The structure of the latter _should_ mean it's pretty comparable to the serde serializer, e.g. no intermediate buffers, short-circuiting for raw bson bytes, etc.

.evergreen/config.yml
isabelatkinson: I coincidentally discovered yesterday that let chains are finally stabilized - very exciting!

src/tests/bench.rs
isabelatkinson: Yeah that is a bummer. If you don't have any obvious leads, maybe we can just file a ticket to investigate later and pick that up if it seems like there's decent usage of this integration?

src/facet/format.rs
```diff
@@ -0,0 +1,256 @@
...
+    Regex,
+    Timestamp,
+    error::{Error, Result},
+    oid::ObjectId,
+    raw::{CStr, MIN_BSON_DOCUMENT_SIZE},
+    spec::{BinarySubtype, ElementType},
+};
+
+/// Serialize a value to BSON bytes.
+pub fn to_vec<'facet, T: Facet<'facet>>(value: &T) -> Result<Vec<u8>> {
```
isabelatkinson:
```suggestion
pub fn serialize_to_vec<'facet, T: Facet<'facet>>(value: &T) -> Result<Vec<u8>> {
```

For symmetry with the existing `serialize_to_vec` serde method

src/facet/format.rs
isabelatkinson: Yeah this is a lot more straightforward!

src/raw/bson.rs
isabelatkinson: Makes sense, can we put this behind the `facet-unstable` flag? It would be nice to maintain a more efficient layout when this isn't needed

src/facet/format.rs
abr-egn: Done.

src/raw/bson.rs
abr-egn: Good thought, done.

src/tests/bench.rs
abr-egn: 👍 Created RUST-2401 to track this.

isabelatkinson (approved):
lgtm!

PR #659, opened 2026-04-27
RUST-2330 Parse binary bson into Facet types

abr-egn:
RUST-2330

This turned out to be a fair bit more complex than the serialization path, and necessitated some refactoring of the core byte parsing.


src/de/raw.rs
```diff
@@ -54,8 +54,9 @@ impl<'de> RawDeserializer<'de> {
     /// Construct a `RawDeserializer` with the provided bytes. Returns an error if the basic
     /// structure of the bytes is invalid.
     pub fn new(buf: &'de [u8]) -> Result<Self> {
+        RawDocument::from_bytes(buf)?;
         Ok(Self {
-            element: RawElement::toplevel(buf)?,
+            element: RawElement::toplevel(buf),
```
abr-egn, line 59: `toplevel` itself has no reason to be fallible, but at a user API boundary we should validate the bytes, hence the `RawDocument::from_bytes` added above.

src/extjson/models.rs
abr-egn, line 1: These changes just make all of the `From` impls consistent on the receiver (ref vs value).

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+    bytes: &'de [u8],
+    state: ParseState,
+    saved: Option<ParseState>,
+}
+
+#[derive(Debug, Clone)]
+struct ParseState {
+    offset: usize,
+    expects: Expect,
+    outer: Vec<Expect>,
```
abr-egn, line 35: When hitting the terminator for a document/array, the parser needs to know whether that value was contained in a document or array so it can emit / omit element keys as appropriate.

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+            state: ParseState {
+                offset: 0,
+                expects: Expect::DocStart,
+                outer: vec![],
+            },
+            saved: None,
+        }
+    }
+
+    fn peek(&self) -> Result<Option<(ParseEvent<'de>, ParseState)>> {
```
abr-egn, line 61: The core of the parsing: given the current offset and expected next thing, what is the event and the state to advance to?

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+                    offset: self.state.offset + 4, // doc length
+                    expects: Expect::ElemKey,
+                    outer: vec![Expect::Eof],
+                };
+            }
+            Expect::DocStartRaw => {
+                let len = iter.next_document_len(self.state.offset)?;
+                let mut bytes = self.bytes[self.state.offset..self.state.offset + len].to_vec();
+                // type tag for parsing a `Bson`/`RawBson` value
+                bytes.push(ElementType::EmbeddedDocument as u8);
```
abr-egn, line 82: Unfortunately, because the way to handle opaque type parsing in Facet is "hand it a byte array", in order to have this work when parsing into a `Bson` / `RawBson` value that doesn't know the subtype, raw values have to be appended with a type tag (which in turn means this can't be zero-copy).  I think this is a shortcoming of the current Facet API and once I figure out what a good way to resolve it would be I intend to file a ticket with them :)

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+                };
+                event = ParseEvent::new(
+                    ParseEventKind::FieldKey(FieldKey::new(
+                        elt.key().as_str(),
+                        FieldLocationHint::KeyValue,
+                    )),
+                    Span::new(self.state.offset, elt.size()),
+                );
+                next = ParseState {
+                    offset: self.state.offset,
```
abr-egn, line 102: The offset doesn't change to advance after an element key so the kind tag is present for parsing the value.

(It really seems like having the order be "tag, key name, key value" was a mistake in hindsight; so much parsing code would be easier if it were "key name, tag, key value", and not just specific to Facet)

src/raw/bson_ref.rs
```diff
@@ -534,6 +544,46 @@ impl RawBinaryRef<'_> {
     }
 }
 
+impl<'a> RawBinaryRef<'a> {
+    pub(crate) fn parse(bytes: &'a [u8]) -> Result<Self> {
```
abr-egn, line 548: This (and similar below) were moved out of the raw iterator parsing code.  Actual logic is unchanged.

src/raw/iter.rs
```diff
@@ -60,38 +39,55 @@ impl<'a> Iterator for Iter<'a> {
 
 /// An iterator over the document's elements.
 pub struct RawIter<'a> {
-    doc: &'a RawDocument,
+    bytes: &'a [u8],
```
abr-egn, line 42: No part of the iterator parsing requires the bytes to be a complete rawdoc, and dropping that made it easy to use this for advancing in a slice.

src/raw/iter.rs
```diff
@@ -60,38 +39,55 @@ impl<'a> Iterator for Iter<'a> {
...
 
     /// Whether the underlying doc is assumed to be valid or if an error has been encountered.
     /// After an error, all subsequent iterations will return None.
     valid: bool,
+
+    /// When true, the iterator treats a null byte as the document terminator only at the last
+    /// position of `bytes`; a null encountered earlier is an error. When false (used by
+    /// `new_unchecked` for flat parsers that span nested documents), any null byte ends
+    /// iteration.
+    strict: bool,
```
abr-egn, line 53: This is an annoying little flag that's added because the bson corpus tests include ones that require an error on trailing unparsed bytes, whereas using it for a partial parse of a larger slice requires _not_ erroring in that case.

src/raw/iter.rs
```diff
@@ -115,10 +111,7 @@ impl<'a> RawIter<'a> {
 #[derive(Clone)]
 pub struct RawElement<'a> {
     key: &'a CStr,
-    kind: ElementType,
-    doc: &'a RawDocument,
-    start_at: usize,
-    size: usize,
+    value: RawValue<'a>,
```
abr-egn, line 114: Splitting out the new `RawValue` type is a pretty substantial refactor but I think it's useful to have a "type tag + byte buffer" struct in general.

src/raw/iter.rs
```diff
@@ -172,110 +161,20 @@ impl<'a> RawElement<'a> {
...
-                    return Err(self.malformed_error("code with scope length too small"));
-                }
-
-                let slice = self.value_bytes();
-                let code = read_lenencode(&slice[4..])?;
-                let scope_start = 4 + 4 + code.len() + 1;
-                let scope = RawDocument::from_bytes(&slice[scope_start..])?;
+        self.value
+            .parse()
+            .map_err(|e| e.with_key(self.key.as_str()))
```
abr-egn, line 172: Side benefit: we weren't consistent about attaching the key to the error before, now we are :)

src/raw/value.rs
```diff
@@ -0,0 +1,179 @@
...
+    },
+    spec::ElementType,
+};
+
+#[derive(Clone)]
+pub(crate) struct RawValue<'a> {
+    kind: ElementType,
+    bytes: &'a [u8],
+    #[allow(unused)]
+    source_offset: usize,
```
abr-egn, line 38: Tracking this makes constructing spans much easier.

src/facet.rs
```diff
@@ -121,155 +124,38 @@ impl TryFrom<ExtJson> for Bson {
...
-            .and_then(|m| m.parse())
-    }
-}
-
-impl TryFrom<&crate::DateTime> for ExtJson {
-    type Error = ToValueError;
-    fn try_from(value: &crate::DateTime) -> Result<Self, Self::Error> {
-        facet_value::to_value(&models::DateTime::from(*value)).map(ExtJson)
-    }
+macro_rules! ext_json_simple {
```
abr-egn, line 127: I added a tabular macro for this and the opaque adapters because I find it _much_ more readable than a long list of repeated `impl` blocks.

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+impl<'a> RawElement<'a> {
+    fn as_event(&self) -> Result<ParseEvent<'a>> {
+        let value = self.value()?;
+        let span = self.value_raw().span();
+        let ek = match value {
+            RawBsonRef::Double(f) => ParseEventKind::Scalar(ScalarValue::F64(f)),
+            RawBsonRef::String(s) => ParseEventKind::Scalar(ScalarValue::Str(Cow::Borrowed(s))),
+            RawBsonRef::Boolean(b) => ParseEventKind::Scalar(ScalarValue::Bool(b)),
+            RawBsonRef::Null => ParseEventKind::Scalar(ScalarValue::Null),
+            RawBsonRef::Int32(i) => ParseEventKind::Scalar(ScalarValue::I64(i as i64)),
```
isabelatkinson, line 221: Looks like `ScalarValue` doesn't support `i32` - does that mean i32s won't round-trip properly? Not sure there's much we can do about this but I was wondering if it was something you considered

src/facet/parse.rs
```diff
@@ -0,0 +1,347 @@
...
+                    }
+                    ek => {
+                        return Err(Error::deserialization(format!(
+                            "unexpected parse event {ek:?}"
+                        )));
+                    }
+                }
+            }
+            Expect::ElemValueRaw { is_array } => {
+                let Some(elt) = iter.next().transpose()? else {
```
isabelatkinson, line 153: Does this need the same handling for arrays as above on line 109?

src/tests/facet/parse.rs
isabelatkinson, line 1: Suggest adding a few sanity check tests for error cases, e.g. malformed bytes, incorrect element tag, missing field

src/facet/parse.rs
abr-egn: i32s will round-trip because facet does the narrowing conversion back from i64 when parsing; however, it does mean that if someone does a rust type -> bson -> load as document, that field will be a bson int64 in the document.  We could attempt to auto-narrow when serializing via facet but I think having field width depend on value is the worse behavior.

src/facet/parse.rs
abr-egn: Good catch, yup, fixed.

src/tests/facet/parse.rs
abr-egn: Done.

isabelatkinson (approved):
lgtm!

PR #661, opened 2026-05-05
minor: update facet dependency versions

abr-egn:
The normal automatic dependency update failed here (https://github.com/mongodb/bson-rust/pull/660) because just doing a patch version bump ends up pulling in incompatible versions.


BorisDog (approved):
LGTM

PR #663, opened 2026-05-13
RUST-2422 Reduce serde raw deserialization boilerplate

abr-egn:
RUST-2422

This takes some types that had identical deserialization impl patterns and collapses them down to a single helper impl struct (and a secondary glue struct).

### DIFF
--- a/src/de/raw.rs
+++ b/src/de/raw.rs
@@ -108,7 +108,9 @@ impl<'de> RawDeserializer<'de> {
             RawBsonRef::String(s) => visitor.visit_borrowed_str(s),
             RawBsonRef::Boolean(b) => visitor.visit_bool(b),
             RawBsonRef::Null => visitor.visit_unit(),
-            RawBsonRef::ObjectId(oid) => visitor.visit_map(ObjectIdAccess::new(oid, hint)),
+            RawBsonRef::ObjectId(oid) => {
+                visitor.visit_map(SingleFieldAccess::new("$oid", ObjectIdDeserializer { oid, hint }))
+            }
             RawBsonRef::Document(doc) => match hint {
                 DeserializerHint::RawBson => visitor.visit_map(RawDocumentAccess::new(doc)),
                 _ => visitor.visit_map(DocumentAccess::new(doc, self.options.clone())?),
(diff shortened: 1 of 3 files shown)


src/de/raw.rs
```diff
@@ -589,65 +578,141 @@ impl<'de> serde::de::Deserializer<'de> for RawDocumentDeserializer<'de> {
...
+    deserializer: Option<D>,
 }
 
-impl ObjectIdAccess {
-    fn new(oid: ObjectId, hint: DeserializerHint) -> Self {
-        Self {
-            oid,
-            visited: false,
-            hint,
+impl SingleFieldAccess<()> {
```
abr-egn, line 586: The `()` type here is just so calls like `SingleFieldAccess::oid` have a concrete type that doesn't require context  to resolve.  Idiomatically I think
```
impl SingleFieldAccess<ObjectIdDeserializer> { fn oid(..) -> Self }
```
would be more typical but that would mean a separate impl block for each constructor fn which doesn't seem useful.

src/de/raw.rs
```diff
@@ -589,65 +578,141 @@ impl<'de> serde::de::Deserializer<'de> for RawDocumentDeserializer<'de> {
...
+            .ok_or_else(|| Error::custom("value already consumed"))?;
+        seed.deserialize(de)
     }
 }
 
-struct ObjectIdDeserializer {
-    oid: ObjectId,
-    hint: DeserializerHint,
-}
+struct MapAccessDeserializer<A>(A);
```
abr-egn, line 696: This is just glue; it lets `RegexAccess`/`DbPointerAccess` plug into `SingleFieldDeserializer`.

src/de/raw.rs
```diff
@@ -656,51 +721,42 @@ impl<'de> serde::de::Deserializer<'de> for ObjectIdDeserializer {
...
-            visited: false,
-        }
-    }
+struct ObjectIdDeserializer {
+    oid: ObjectId,
+    hint: DeserializerHint,
 }
 
-impl<'de> serde::de::MapAccess<'de> for Decimal128Access {
+impl<'de> serde::de::Deserializer<'de> for ObjectIdDeserializer {
```
abr-egn, line 733: Github's diff parsing is failing here - ObjectIdDeserializer has not changed at all.

isabelatkinson (approved):

PR #664, opened 2026-05-14
RUST-2424 Fix doctest errors and warnings with default features

abr-egn:
RUST-2424

Just a matter of adding the right feature gating.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -54,6 +54,8 @@ use crate::error::{Error, Result};
 ///
 /// e.g.
 /// ```rust
+/// # #[cfg(all(feature = "serde", feature = "chrono-0_4"))]
+/// # {
 /// use serde::{Serialize, Deserialize};
 ///
 /// #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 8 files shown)


src/datetime.rs
```diff
@@ -65,11 +67,10 @@ use crate::error::{Error, Result};
 ///     chrono_datetime: chrono::DateTime<chrono::Utc>,
 /// }
 ///
-/// # fn main() -> bson::error::Result<()> {
```
abr-egn, line 68: Doctests are a little weird; if there's no toplevel `fn main`, one wrapping the body of the doctest will be added automatically.  The added block for the feature-gate means this `main` would no longer be toplevel, making this turn into something like:
```rust
fn main() -> impl std::process::Termination {
  #[cfg(all(feature = "serde", feature = "chrono-0_4"))]
  {
    // ... setup code ...
    fn main() -> bson::error::Result<()> {
      // ... test code ...
    }
  }
}
```
That would mean that the test code would be compiled but not actually executed.  Dropping the inner `main` fixes that.

BorisDog (approved):
LGTM 
Question: Should we have something on EG that surfaces such failures?

abr-egn:
> Question: Should we have something on EG that surfaces such failures?

Good thought; we actually had a workaround in the evergreen test runner script to specifically skip doctests when running on the default feature flags, so I just removed that 🙂

PR #672, opened 2026-06-23
RUST-2401 Fix benchmarks and add deserialization

abr-egn:
RUST-2401

I have these changes in my primary branch for this ticket, but I realized it'd be good to split them out - at the very least, to make it easy to run before and after comparisons.


BorisDog (approved):
LGTM

Issue #684, opened 2026-08-06
Implement `Eq` and `Hash` for `RawBson` and `RawBsonRef`


tyilo:
I started working on this, but then I discovered #683 which should be fixed first.

tyilo:
Fixed by #686

PR #686, opened 2026-08-14
RUST-2471 Fix Eq implementation for Bson; extend Eq/Hash to raw types

### LINKED ISSUE #683: RUST-2471 `Bson` doesn't uphold `Eq`'s reflexivity guarantee
According to the [`Eq` documentation](https://doc.rust-lang.org/stable/std/cmp/trait.Eq.html) any type implementing the trait should ensure that all values are reflexive.

This is not the case for `Bson` even though it implements `Eq`:
```rust
fn main() {
    let v = bson::Bson::Double(f64::NAN);
    assert_eq!(v, v);
}
```

```
thread 'main' (2304097) panicked at src/main.rs:3:5:
assertion `left == right` failed
  left: Double(NaN)
 right: Double(NaN)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Either the `Eq` implementation should be removed or `Bson::eq` should compare the bits of `f64`'s (or something similar).

abr-egn:
RUST-2471

This changes the `Bson` impl for `PartialEq` to compare by byte value rather than the derived comparison, which carried the float non-reflexive behavior and made `Eq` a lie.  To keep things consistent, I updated the raw types to use the same comparison; this also meant making the raw types Eq/Hash was straightforward.

Notably, the raw buffer types (`RawDocument[Buf]`, `RawArray[Buf]`) were _already_ just comparing by byte value, so this change makes everything consistent rather than having an undocumented mismatch.

Fixes #683, #684.

### DIFF
--- a/src/bson.rs
+++ b/src/bson.rs
@@ -44,7 +44,10 @@ use crate::{
 };
 
 /// Possible BSON value types.
-#[derive(Clone, Default, PartialEq)]
+///
+/// Note that `Bson`'s [`PartialEq`] implementation is byte value equality, so `Bson::Double(v)` and
+/// `v` may compare differently.
+#[derive(Clone, Default)]
 #[cfg_attr(
     feature = "facet-unstable",
     repr(C),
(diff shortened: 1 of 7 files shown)


src/raw/bson.rs
```diff
@@ -329,6 +332,66 @@ impl RawBson {
...
+            Self::MinKey => matches!(other, Self::MinKey),
+        }
+    }
+}
+
+impl Eq for RawBson {}
+
+impl std::hash::Hash for RawBson {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        std::mem::discriminant(self).hash(state);
```
abr-egn, line 371: This is unrelated to the rest of the changes, but since we're changing the hash output (or adding it in the case of the raw types), this makes it behave a little better.  Including the discriminant in the hash reduces collisions; previously, all the unit variants would hash equal, as would pairs like `Bson::String("x")` and `Bson::Symbol("x")`.

src/raw/bson_ref.rs
```diff
@@ -300,6 +303,66 @@ impl<'a> RawBsonRef<'a> {
     }
 }
 
+impl<'a> PartialEq for RawBsonRef<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        match self {
+            Self::Double(s) => matches!(other, Self::Double(o) if s.to_bits() == o.to_bits()),
```
abr-egn, line 309: This slightly awkward construction means there doesn't need to be a fallthrough case like there would be for matching directly against `(self, other)`, so we keep the compile-time exhaustiveness check.

src/bson.rs
```diff
@@ -102,34 +105,58 @@ pub type Array = Vec<Bson>;
 
 impl Hash for Bson {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        std::mem::discriminant(self).hash(state);
         match self {
-            Bson::Double(double) => {
-                if *double == 0.0_f64 {
```
abr-egn, line 107: This special casing goes away because both equality and hash are just using the byte values.

isabelatkinson (approved):
very sorry for the delay on this!

PR #689, opened 2026-08-26
RUST-2488 Fix lossy `JavascriptCodeWithScope` decoding

dxbjavid:
When `parse_utf8_lossy` handles a JavaScriptCodeWithScope it works out where the scope document begins from the length of the lossily decoded code string, but that length can be larger than the bytes actually stored whenever the code isn't valid utf-8 (each bad byte decodes to a three-byte replacement character). The offset then overshoots and the conversion either bails out with a misleading "length overrun"/"document too short" error or picks up the wrong scope, which is unfortunate given `Utf8Lossy` is the path meant to tolerate invalid utf-8 in the first place. This computes `scope_start` from the raw code byte length returned by `read_lenencode_bytes` and only decodes the code afterwards. I've added a regression test that round-trips a code-with-scope whose code holds invalid utf-8.

### DIFF
--- a/src/raw/test.rs
+++ b/src/raw/test.rs
@@ -7,8 +7,10 @@ use crate::{
     Bson,
     DateTime,
     Decimal128,
+    Document,
     Regex,
     Timestamp,
+    Utf8Lossy,
     oid::ObjectId,
     spec::BinarySubtype,
 };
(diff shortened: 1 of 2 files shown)


isabelatkinson (commented):
Thanks for the fix! Changes look good. Can you please merge/rebase on main to pull in a lint fix for CI?

abr-egn (approved):

PR #694, opened 2026-09-16
RUST-2472 Add a max nesting limit to decoding/deserialization

abr-egn:
RUST-2472

This adds a limit of 200 (same as C's libbson and the server).  It turned out that the serde deserialization would overflow even with that limit; some straightforward optimization fixed that, and I applied the same to the direct `try_from` path because why not.

Facet gets to dodge this because we parse values as an event stream so there's no direct recursion; the facet-side code driving that stream probably recurses but we don't control that.


src/de/serde.rs
```diff
@@ -546,6 +310,277 @@ impl<'de> Visitor<'de> for BsonVisitor {
...
+    key: &str,
+    visitor: &mut V,
+) -> std::result::Result<Option<Bson>, V::Error>
+where
+    V: MapAccess<'de>,
+{
+    use crate::extjson;
+
+    let bson = match key {
+        "$oid" => {
```
abr-egn, line 341: This is all just relocated unchanged.

src/raw/bson_ref.rs
```diff
@@ -486,7 +519,7 @@ impl<'a> TryFrom<RawBsonRef<'a>> for Bson {
     type Error = Error;
 
     fn try_from(rawbson: RawBsonRef<'a>) -> Result<Bson> {
-        RawBson::from(rawbson).try_into()
+        rawbson.try_into_parsed(0)
```
abr-egn, line 522: This previously delegated to RawBson::try_into, which was convenient for not repeating the match statement, but meant we were actually allocating a copy for any buffer element encountered in traversal and then throwing it away again.

src/raw/bson_ref.rs
```diff
@@ -785,9 +818,12 @@ impl<'a> From<RawRegexRef<'a>> for RawBsonRef<'a> {
 
 impl<'a> From<RawRegexRef<'a>> for Regex {
     fn from(value: RawRegexRef<'a>) -> Self {
+        let mut chars: Vec<_> = value.options.as_str().chars().collect();
```
abr-egn, line 821: Unrelated bug: this wasn't normalizing the options order, so when RawBsonRef::try_into became its own match, the corpus test checking for that started failing.

src/raw/document.rs
```diff
@@ -615,9 +630,10 @@ fn deep_utf8_lossy(src: RawBson) -> RawResult<Bson> {
             let mut tmp = doc! {};
             for elem in scope.iter_elements() {
                 let elem = elem?;
+                // the scope is a nested document, so it counts as an extra level of nesting
                 tmp.insert(
                     elem.key().as_str(),
-                    deep_utf8_lossy(elem.value_utf8_lossy()?)?,
+                    deep_utf8_lossy(elem.value_utf8_lossy()?, depth.saturating_add(2))?,
```
abr-egn, line 636: Code with scope is weird: it has to pay a stack cost for both the element itself and the enclosing document, so it accumulates frames twice as fast.  Counting it as two frames here seems the most reasonable approach.  The `try_from` path ends up with an equivalent count because it increments at slightly different points.

src/raw/test.rs
```diff
@@ -506,6 +479,82 @@ fn fuzz_oom() {
...
+        cstr,
+    };
+
+    #[derive(Debug, Clone, Copy)]
+    enum Nesting {
+        Document,
+        Array,
+        CodeWithScope,
+    }
+    fn nested_bson(depth: usize, nesting: &[Nesting]) -> RawDocumentBuf {
```
abr-egn, line 500: This is the example from the original user report, generalized a bit and using raw types to avoid having to do byte buffer manipulation directly.

src/tests/modules/bson.rs
```diff
@@ -173,7 +173,7 @@ fn from_impls() {
         Bson::from(vec![1, 2, 3]),
         Bson::Array(vec![Bson::Int32(1), Bson::Int32(2), Bson::Int32(3)])
     );
-    #[cfg(feature = "serde")]
+    #[cfg(all(feature = "serde", feature = "serde_json-1"))]
```
abr-egn, line 176: Not directly related, but we had a bunch of tests that would fail to compile if serde was enabled but serde-adjacent things weren't, which was a hassle while testing this change so I fixed it.

Cargo.toml
```diff
@@ -51,6 +51,7 @@ serde_with-3 = ["dep:serde_with", "dep:serde"]
 serde = ["dep:serde"]
 serde_json-1 = ["dep:serde_json"]
 facet-unstable = ["dep:facet", "dep:facet-value", "dep:facet-format", "dep:facet-reflect", "dep:paste"]
+unbounded-depth = []
```
abr-egn, line 54: Making the limit opt-out as this does could hypothetically break an existing process, if they happened to be parsing values above the new limit but below actual stack overflow depth.  Given that this is protecting against a process crash I think this is reasonable, we should just make sure to call it out in the release notes.

.evergreen/run-tests.sh
```diff
@@ -6,3 +6,4 @@ set -o errexit
 
 RUST_BACKTRACE=1 cargo test
 RUST_BACKTRACE=1 cargo test --all-features
+RUST_BACKTRACE=1 cargo test --features serde nesting_limit
```
abr-egn, line 9: Without this, the serde path wouldn't be tested since it's not in default features and --all-features would disable the limit.

isabelatkinson (approved):

